PR #23, opened 2017-11-23
Pull json schemas from json schema store for well known yaml files

JPinkney:
This PR gets json schemas from the json schema store api [catalog endpoint](http://schemastore.org/api/json/catalog.json).

The question is whether we want to have this intergrated as a feature, especially if we can't verify the quality of the yaml files on the json schema store.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -223,12 +223,8 @@ function getSchemaStoreMatchingSchemas(){
 		for(let schemaIndex in schemas.schemas){
 			
 			let schema = schemas.schemas[schemaIndex];
-			if(schema && schema.name){
-				
-				if(schema.name.indexOf('.yml') !== -1 || schema.name.indexOf('.yaml') !== -1){
-					languageSettings.schemas.push({ uri: schema.url, fileMatch: schema.fileMatch });
-				}
-			
+			if(schema && schema.name && schema.name.indexOf('.yml') !== -1 || schema.name.indexOf('.yaml') !== -1){
+				languageSettings.schemas.push({ uri: schema.url, fileMatch: schema.fileMatch });
 			}
 
 		}
--- a/src/server.ts
+++ b/src/server.ts
@@ -239,7 +239,7 @@ function getSchemaStoreMatchingSchemas(){
 		return languageSettings;
 
 	}, (error: XHRResponse) => {
-		return Promise.reject(error.responseText || getErrorStatusDescription(error.status) || error.toString());
+		throw error;
 	});
 
 }
--- a/src/server.ts
+++ b/src/server.ts
@@ -162,6 +162,8 @@ let formatterRegistration: Thenable<Disposable> = null;
 let specificValidatorPaths = [];
 let schemaConfigurationSettings = [];
 let yamlShouldValidate = true;
+let hasSchemaStoreBeenAdded = false;
+let schemaStoreSettings = [];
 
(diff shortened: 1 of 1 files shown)


fbricon:
maybe add a preference to enable/disable automatic schemas support from schemastore.org

src/server.ts
```diff
@@ -197,9 +201,50 @@ connection.onDidChangeConfiguration((change) => {
...
+			schemas: []
+		};
+
+		let schemas = JSON.parse(response.responseText);
+		for(let schemaIndex in schemas.schemas){
+			
+			let schema = schemas.schemas[schemaIndex];
+			if(schema && schema.name){
+				
+				if(schema.name.indexOf('.yml') !== -1 || schema.name.indexOf('.yaml') !== -1){
```
fbricon: both `if` statements can be combined

src/server.ts
```diff
@@ -223,12 +223,8 @@ function getSchemaStoreMatchingSchemas(){
 		for(let schemaIndex in schemas.schemas){
 			
 			let schema = schemas.schemas[schemaIndex];
-			if(schema && schema.name){
-				
-				if(schema.name.indexOf('.yml') !== -1 || schema.name.indexOf('.yaml') !== -1){
-					languageSettings.schemas.push({ uri: schema.url, fileMatch: schema.fileMatch });
-				}
-			
+			if(schema && schema.name && schema.name.indexOf('.yml') !== -1 || schema.name.indexOf('.yaml') !== -1){
```
fbricon: `schema && schema.name && (schema.name.indexOf('.yml') !== -1 || schema.name.indexOf('.yaml') !== -1))` 
or else could be in trouble

src/server.ts
```diff
@@ -162,6 +162,8 @@ let formatterRegistration: Thenable<Disposable> = null;
 let specificValidatorPaths = [];
 let schemaConfigurationSettings = [];
 let yamlShouldValidate = true;
+let hasSchemaStoreBeenAdded = false;
```
gorkem: Do you need this boolean? Can you just check `schemaStoreSettings` ?

src/server.ts
```diff
@@ -197,9 +201,46 @@ connection.onDidChangeConfiguration((change) => {
...
+		
+		let languageSettings= {
+			schemas: []
+		};
+
+		let schemas = JSON.parse(response.responseText);
+		for(let schemaIndex in schemas.schemas){
+			
+			let schema = schemas.schemas[schemaIndex];
+			if(schema && schema.name && schema.name.indexOf('.yml') !== -1 || schema.name.indexOf('.yaml') !== -1){
```
felixfbecker: The name has no semantic meaning, you should look only look at the `fileMatch` property

src/server.ts
```diff
@@ -197,9 +200,54 @@ connection.onDidChangeConfiguration((change) => {
...
+				}
+				
+			}
+
+		}
+
+		return languageSettings;
+
+	}, (error: XHRResponse) => {
+		return Promise.reject(error.responseText || getErrorStatusDescription(error.status) || error.toString());
```
felixfbecker: inside a Promise callback you can use `throw` instead of `return Promise.reject()`
It's a good idea to always throw/reject `Error` instances instead of strings because strings don't have a stack trace and cannot have additional metadata.

PR #28, opened 2017-12-06
Added multi-root setting support

JPinkney:


src/server.ts
```diff
@@ -97,6 +104,22 @@ let workspaceContext = {
...
 
 let schemaRequestService = (uri: string): Thenable<string> => {
+	//For the case when we are multi root and specify a workspace location
+	if(hasWorkspaceFolderCapability){
+		for(let folder in workspaceFolders){
+			let currFolder = workspaceFolders[folder];
+			let currFolderUri = currFolder["uri"];
+			let currFolderName = currFolder["name"];
+
+			var isUriRegex = new RegExp('^(?:[a-z]+:)?//', 'i');
```
gorkem: use `let` instead of `var`

src/server.ts
```diff
@@ -97,6 +104,22 @@ let workspaceContext = {
...
+			if(uri.indexOf(currFolderName) !== -1 && !uri.match(isUriRegex)){
+				let beforeFolderName = currFolderUri.substring(0, currFolderUri.indexOf(currFolderName));
+				let afterFolderName = uri.substring(uri.indexOf(currFolderName), uri.length);
+				uri = beforeFolderName + "/" + afterFolderName;
```
gorkem: Using `String.split()` will probably make this simpler.

src/languageService/services/configuration.ts
```diff
@@ -1,257 +0,0 @@
-/*---------------------------------------------------------------------------------------------
```
gorkem: Is this removal related to multi-root?

gorkem:
@JPinkney I think we need to update travis.yml to match the version update for typescript in this one.

PR #31, opened 2018-01-11
Mocha should be a dev dependency and not a regular dependency

rcjsuen:
As Mocha is only used for testing and not for the actual execution of the language server, it should be listed in the `devDependencies` section of the package.json file instead of the `dependencies` section.


gorkem:
Thanks Remy for the PR.

PR #38, opened 2018-01-18
Add custom schema provider feature.

andxu:


JPinkney:
Hi!, I just need some quick clarification on this PR. Is the plan to somehow setup the client to send different schemas to the server based on the context of the kind field in the yaml file? In this case, we could prefer this to stay fully on the server side if possible because we use the language server in Eclipse Che, Atom, and VSCode.

Right now we are looking into adapting our kubernetes schema to use https://github.com/garethr/openshift-json-schema instead of the current one.

PR #41, opened 2018-01-18
WIP - New Kubernetes Schema + Updated support

JPinkney:
The original kubernetes schema required a lot of patches in order to have it perform well. Moving to https://github.com/garethr/openshift-json-schema/ allows us to have better user experience with kubernetes, however we are losing some of the auto completion on values because the schema does not provide them. Additionally, autocompletion on keys will perform much better and there will be much better support for other kubernetes features.

Work Required
- [X] Removal of kubernetes from tests, server.ts
- [X]  Removal of anyOfMatching from jsonParser
- [X]  Removal of kubernetesTransformer
- [X]  Addition of new schema
- [x]  Fix tests to reflect the loss of features
- [x]  Adopt kubernetes to use the kind field to perform better autocompletion results


andxu:
I encounter "Matches multiple schemas when only one must validate." when using https://raw.githubusercontent.com/garethr/openshift-json-schema/master/v3.6.0-standalone-strict/all.json

andxu:
For the last work item, I think it is inconvenient for users configure their yaml through vscode settings for each yaml file precisely, so we can not come into the step of determining which yaml schema[Pod? Service?...] without the step of determining it is a kubernetes yaml file, as you might see, there are various of kubernetes yaml configurations like helm/chart/kedge, and it is not suitable for a single yaml-language-server to determine all kinds of yaml files, so on my PR, I open an extension point for all plugins based on this yaml server to have the ability of first determining which kind of the yaml first(not only  to the yaml content but also the project structure and even the schema choosing result from user input), would you please consider it to make an framework to determine file type and then to provide custom schema and keep this project neutral for common yaml files plus schema validation/completion?

andxu:
About this PR, it only solves the validation, how about the schema for completion/hover?

JPinkney:
This PR is WIP. Hover should work fine afaik though we can't get any results out of "x-kubernetes-group-version-kind" anyways. Completion should be added in the next few days.

PR #85, opened 2018-08-29
intellisense doesn't work for array item

928PJY:
if there is no space between `-` and `holder`, the yaml structure will be wrong, so the array intellisense doesn't work


928PJY:
Some comments? @JPinkney

928PJY:
This PR has been inactive for 7 days, can you @JPinkney give some update?

JPinkney:
I have been on vacation.

928PJY:
oh~ sorry for that

---
In reply to: [418946400](https://github.com/redhat-developer/yaml-language-server/pull/85#issuecomment-418946400) [](ancestors = 418946400)

PR #87, opened 2018-09-06
support generate snippets for object and array according to schema

928PJY:


928PJY:
can you give me schema that I can repro this? it's not an expected behavior

---
In reply to: [419091296](https://github.com/redhat-developer/yaml-language-server/pull/87#issuecomment-419091296) [](ancestors = 419091296)

JPinkney:
I was testing on the kubernetes schema (https://gist.githubusercontent.com/JPinkney/ccaf3909ef811e5657ca2e2e1fa05d76/raw/f85e51bfb67fdb99ab7653c2953b60087cc871ea/openshift_schema_all.json) or you can set it in the settings by doing: 
```
"yaml.schemas": {
    "kubernetes": "*.yaml"
}
```

928PJY:
got it, this happened in the case that there is no required property or property with default value

---
In reply to: [419298080](https://github.com/redhat-developer/yaml-language-server/pull/87#issuecomment-419298080) [](ancestors = 419298080,419091296)

JPinkney:
> can we add test to test completion detail but not just completion item count

Yeah, thats something thats always been needed. I'm going to create an issue and will start converting the tests when I get a change.

Other then that it LGTM!

928PJY:
Cool~
when would you publish a new version Npm package and extension?

JPinkney:
Should be published in a few hours, I just have to do a few other things first

928PJY:
@JPinkney Cool! Thanks~

PR #96, opened 2018-10-14
Enable to set bracketSpacing option to false

iquiw:
Also moved "setting.yaml" existence check in one if block.

This fixes a problem that `yaml.format.bracketSpacing` is unable to be set to false.


JPinkney:
Sorry about the delay and thank you for your contribution 😃!

PR #101, opened 2018-11-05
Fix for redhat-developer/vscode-yaml#116

remcohaszing:
This change was made over a month ago, but it was never released.

See redhat-developer/vscode-yaml#116


JPinkney:
Thanks for bringing this to my attention, I must have forgotten about that branch

Issue #58, opened 2018-04-01
Kubernetes apiVersion of extensions is not supoported


tdabasinskas:
It seems that the following YAML file for a cert-manager configuration is not supported (getting `Value is not accepted. Valid values: "v1"`):

```yaml
apiVersion: certmanager.k8s.io/v1alpha1
kind: Certificate
...
```

It also seems to be the case with other extension versions, e.g. [`certificates./v1beta1`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#certificatesigningrequest-v1beta1-certificates) and [`extensions/v1beta1`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#ingress-v1beta1-extensions).

thearrow:
Running into the same thing on Kubernetes Ingress resources:

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
[...]
```

> Value is not accepted. Valid values: "v1beta1".

mattayes:
Running into the same issues with Ingress as well.

litaxc:
Same kind of issue here:
```yml
apiVersion: apps/v1
kind: Deployment
```
Is there any way to fix this? I am willing to contribute.

JPinkney:
I'm writing a python script that should fix the generation of that portion of the schema. I didn't write the original schema and the creators of the schema used a weird formatting that can't be correctly parsed. Hopefully, I can get it working for the next release.

PR #108, opened 2018-12-07
Update kube schema url to v1.14.0

jburianek:
This seems to fix https://github.com/redhat-developer/yaml-language-server/issues/58 for me.

Is the fix that straightforward, or is there something I'm missing? Thanks!

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -169,7 +169,7 @@ let schemaRequestService = (uri: string): Thenable<string> => {
 	});
 };
 
-export let KUBERNETES_SCHEMA_URL = "https://gist.githubusercontent.com/JPinkney/ccaf3909ef811e5657ca2e2e1fa05d76/raw/f85e51bfb67fdb99ab7653c2953b60087cc871ea/openshift_schema_all.json";
+export let KUBERNETES_SCHEMA_URL = "https://raw.githubusercontent.com/garethr/kubernetes-json-schema/master/v1.14.0-standalone-strict/all.json";
 export let KEDGE_SCHEMA_URL = "https://raw.githubusercontent.com/kedgeproject/json-schema/master/master/kedge-json-schema.json";
 export let customLanguageService = getCustomLanguageService(schemaRequestService, workspaceContext, []);
 


JPinkney:
Does autocompletion still work on kind/apiVersion also validation? The problem is that the official schema has this "x-kubernetes-group-version-kind" field in that holds different apiVersion, kind, and group they can be associated with e.g. https://github.com/garethr/kubernetes-json-schema/blob/master/v1.9.9-local/_definitions.json#L24. The problem with this is that it isn't supported by [json schema spec](https://json-schema.org/specification.html) and when I was trying to get it to work in the parser I ran into some issue (I cannot remember what the issue was) that made it so you had to transform the schema. IIRC I have a python script somewhere that basically took the "x-kubernetes-group-version-kind" and put it inside of the properties field in a different format that made the schema work correctly. Thats why there is the patched schema in there.

[Here](https://gist.github.com/JPinkney/847045c3b6691ba4b85c52fd77783867#file-openshift_schema-json-L420) is an example of what had to be done. apiVersion became its own field inside of properties with type string and enum v1 (basically the transformed version of x-kubernetes-group-version-kind below).

jburianek:
Ah ok, that makes sense. You're right, that does break autocompletion for apiVersion/kind. I knew I had to be missing something 😃

jburianek:
I'd be interested in that script if you can find it. Might be worthwhile to commit it to source somewhere too.

JPinkney:
I'll try to look for it but if I can't I'll have to wait until after my exams and I'll create a repo or a gist of the code

pengx17:
> IIRC I have a python script somewhere that basically took the "x-kubernetes-group-version-kind" and put it inside of the properties field in a different format that made the schema work correctly. Thats why there is the patched schema in there.

@JPinkney I am also interested in the script you mentioned here. And good luck to your exams :D

JPinkney:
@pengx17 I couldn't find the script and was going to write a new one but it looks like @jburianek already has a PR open on the schema side for all the fixes (https://github.com/garethr/kubernetes-json-schema/pull/16).

Awesome work @jburianek !

jburianek:
Things have been fixed upstream, so I've updated this PR to point to the JSON schema that should resolve this.

jburianek:
@JPinkney Now that the schema side has been fixed, any chance we could merge this to pick up the latest schema?

JPinkney:
@jburianek Have you played around with it to make sure everything is okay? (I haven't checked it out yet so I was just wondering if there were any other issues found). If we can't find anything I'll try and cut a release near the end of this week with it in it!

jburianek:
@JPinkney Thanks, yeah I've played around with it and things seem to work as expected.

Before:
Screen Shot 2019-05-21 at 11 55 23 AM
Screen Shot 2019-05-21 at 12 41 57 PM

After:
Screen Shot 2019-05-21 at 11 57 49 AM
Screen Shot 2019-05-21 at 11 53 10 AM

PR #109, opened 2018-12-13
fix: syntax error should start from col 0 and end at mark position

pengx17:
YAML syntax error position was not correctly reported like this:

Since we can only get error's starting position from parsing from yaml-ast-parser, I guess the solution here is to mark the error from line start and end at the parsing position.

After the fix:

### DIFF
--- a/src/languageservice/parser/yamlParser.ts
+++ b/src/languageservice/parser/yamlParser.ts
@@ -200,7 +200,7 @@ function recursivelyBuildAst(parent: ASTNode, node: Yaml.YAMLNode): ASTNode {
 }
 
 function convertError(e: Yaml.YAMLException) {
-	return { message: `${e.reason}`, location: { start: e.mark.position, end: e.mark.position + e.mark.column, code: ErrorCode.Undefined } }
+	return { message: `${e.reason}`, location: { start: e.mark.position - e.mark.column, end: e.mark.position, code: ErrorCode.Undefined } }
 }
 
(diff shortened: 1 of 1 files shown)


JPinkney:
Sorry for the late reply but I've finally had the time to look at this. It looks like its having an issue with [this test](https://github.com/redhat-developer/yaml-language-server/blob/master/test/schemaValidation.test.ts#L172). From my quick investigation it looks like the issue is because [this part of the parser](https://github.com/redhat-developer/yaml-language-server/blob/master/src/languageservice/parser/yamlParser.ts#L218) checks to see if the merge key is in the message and if it is it disables the duplicate error message. With the new change the duplicate error message is thrown before the merge keys and thus the merge key can't be captured in the text. I think a simple fix would be to use the old convert error location i.e. start: e.mark.position, end: e.mark.position + e.mark.column inside of that piece of code and then your new code for the rest.

I have found one other (related) issue though. When you have a set of duplicate keys that are indented the error does not properly highlight the keys. Instead it puts the error message right before. I'm not sure if this is a unique case or whether indentation might cause the fix to not work in some cases. I'm still testing it out.
screen shot 2018-12-21 at 11 12 25 pm

joshuawilson:
Can we close this? Or do we plan to merge it?

ssbarnea:
@JPinkney @joshuawilson I think this is safe to close because vscode does use index-from 1 for line and column numbers, zero is not a valid value for either of them. This is in sync with all editors/linters I know so far.  First character of is `filename:1:1`. I recently had to address a bug related to this in ansible-lint.

PR #119, opened 2019-01-08
Only set CompletionItem.textEdit if it encompasses a single line

segevfiner:
This is a limitation imposed by the language server protocol. Doing otherwise will make some language protocol clients, such as VS Code, reject the completions. VS Code prints the following in the developer tools console for if you try and give it a multi-line `CompletionItem.textEdit`: `INVALID text edit -> must be single line and on the same line`, and it ignores all such completions, leading to it offering the default word completions instead.

See [LSP Spec - textDocument/completion](https://microsoft.github.io/language-server-protocol/specification#textDocument_completion):
```ts
	/**
	 * An edit which is applied to a document when selecting this completion. When an edit is provided the value of
	 * `insertText` is ignored.
	 *
	 * *Note:* The range of the edit must be a single line range and it must contain the position at which completion
	 * has been requested.
	 */
	textEdit?: TextEdit;
```

Fixes redhat-developer/vscode-yaml#139

cc @fbricon

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -107,7 +107,7 @@ export class YAMLCompletion {
 				let existing = proposed[suggestion.label];
 				if (!existing) {
 					proposed[suggestion.label] = suggestion;
-					if (overwriteRange) {
+					if (overwriteRange && overwriteRange.isSingleLine) {
 						suggestion.textEdit = TextEdit.replace(overwriteRange, suggestion.insertText);
 					}
 					result.items.push(suggestion);
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -107,7 +107,7 @@ export class YAMLCompletion {
 				let existing = proposed[suggestion.label];
 				if (!existing) {
 					proposed[suggestion.label] = suggestion;
-					if (overwriteRange && overwriteRange.isSingleLine) {
+					if (overwriteRange && overwriteRange.start.line === overwriteRange.end.line) {
 						suggestion.textEdit = TextEdit.replace(overwriteRange, suggestion.insertText);
 					}
 					result.items.push(suggestion);
--- a/test/autoCompletion3.test.ts
+++ b/test/autoCompletion3.test.ts
@@ -65,6 +65,16 @@ suite("Auto Completion Tests", () => {
 					assert.notEqual(result.items.length, 0);
 				}).then(done, done);
 			});
+
+			it('Array of enum autocomplete with multiline text', (done) => {
+				let content = "optionalUnityReferences:\n  - T\n    e\n";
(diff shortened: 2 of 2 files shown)


fbricon:
I believe splitting multiline textedits into one single line textedit and multiple additionalTextEdits would be more accurate than just ignoring the completion item.

segevfiner:
It's not ignoring it. Just not setting a `textEdit`. So it just completes normally using `insertText`.

`additionalTextEdits` documentation says:
```
	 * Additional text edits should be used to change text unrelated to the current cursor position
	 * (for example adding an import statement at the top of the file if the completion item will
	 * insert an unqualified type).
```
So I'm not sure if it's intended for this use case and whether it will work well... There might be a reason why the main `textEdit` is limited to one line, but I don't really know...

Also, if we do try to do this, than `additionalTextEdits` can probably be multiline, so we can probably just add text edits that delete the relevant text in the lines before and after.

segevfiner:
https://github.com/Microsoft/language-server-protocol/issues/92

One issue raised there is that the proposed hack to workaround the one line limitation would cause the cursor to end up in an awkwad position after completion.

segevfiner:
ping @fbricon

fbricon:
relying on insertText doesn't seem like a good idea since this field is deprecated. While I understand textEdit should cover a single line range, as per the spec, vscode certainly supports multiline ranges, as I can see yaml (or java) multiline edits being properly inserted. It's possible the issue you're seeing is the textedit's range not starting on the line being edited.

redborian:
@segevfiner Thanks for the fix.
With this bug, the experience in VSCode is pretty bad. Maybe we can try to fix this the correct way after merging this fix?

gorkem:
@segevfinder Can you possibly add a test that covers this case?

segevfiner:
Actually... This had a bug. `vscode-languageserver-types` doesn't implement `Range.isSingleLine` :man_shrugging: 

Also added a test for this.

gorkem:
I agree with @fbricon that falling back to `insertText` is not the right way on the long term nor  depending on the tolerant implementation of vscode around TextEdit. 
@JPinkney can you also review and test for any regressions due to insertText usage.

segevfiner:
Not sure when they changed it but `CompletionItem.textEdit` is marked as deprecated in the VS Code API docs: https://code.visualstudio.com/api/references/vscode-api#CompletionItem.textEdit. They suggest you **do** use `insertText`, and use `range` to mark what to replace, which *still has* a single-line limitation. It looks to me that `additionalTextEdits` is meant for edits in the document that are side effects of the completion, like adding imports, and not for deleting lines around it, the VS Code docs state:

> When selecting a completion item in the editor its defined or synthesized text edit will be applied to all cursors/selections whereas additionalTextEdits will be applied as provided.

And no idea where the editing cursor ends up if you use it to delete lines around the completion.

I guess someone will need to ask someone in the VS Code team who is familiar with this on what we should do for this case.

PR #126, opened 2019-02-20
Fix formatter

JPinkney:
This PR fixes the formatter so that it respects the yaml.format.enable setting.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -199,7 +199,8 @@ let yamlFormatterSettings = {
 	singleQuote: false,
 	bracketSpacing: true,
 	proseWrap: "preserve",
-	printWidth: 80
+	printWidth: 80,
+	enable: true
 } as CustomFormatterOptions;
 let yamlShouldHover = true;
 let yamlShouldCompletion = true;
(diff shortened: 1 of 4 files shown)


src/server.ts
```diff
@@ -230,6 +231,9 @@ connection.onDidChangeConfiguration((change) => {
 			if (settings.yaml.format.bracketSpacing === false) {
 				yamlFormatterSettings.bracketSpacing = false;
 			}
+			if (settings.yaml.format.enable) {
+				yamlFormatterSettings.enable = settings.yaml.format.enable || true;
```
walles: Maybe I'm misunderstanding something here, but will this ever set `yamlFormatterSettings.enable` to `false`?

If `settings.yaml.format.enable` is `false`, won't this code block just be ignored?

walles: And won't the `|| true` ensure that this would always be set to `true` anyway?

But I'm bad at Javascript so maybe this actually works.

walles: And at Typescript :)

JPinkney: Yep! Good catch! I must not have had my coffee yet when I made the PR

Issue #142, opened 2019-03-27
Realign features of YAML Language Server with JSON Language Server


JPinkney:
Currently, the YAML language server is quite a bit behind the JSON language server in terms of features and will take quite a bit of work to get everything back up to date. This issue will keep track of everything that needs to be done.

- [x] Extend the JSON parser from vscode-json-languageservice rather then use the current version in yaml-language-server
- [x] Replace yamlParser code that recursively builds the AST with new types
- [x] Replace documentSymbols with vscode-json-languageservice's implementation of document symbols and change it to work with YAML's multiple documents
- [x] Replace yamlHover with vscode-json-languageservice's implementation for json, should work the same since theres nothing specific to YAML in this.
- [x] Replace yamlValidation with vscode-json-languageservice's implementation for json, should work the same since theres nothing specific to YAML in this.
- [x] See if there is a way to extend the jsonSchemaService
- [x] See what other features we can use from vscode-json-languageservice that would be relevant

By doing this we solve:
https://github.com/redhat-developer/vscode-yaml/issues/154
https://github.com/redhat-developer/vscode-yaml/issues/151
https://github.com/redhat-developer/yaml-language-server/issues/133
https://github.com/redhat-developer/yaml-language-server/issues/105

marosg42:
@JPinkney I only wanted to bump this issue. I just started to use yaml extension in vscode and realized jsonschema draft-7 enhancement if-then-else is  not available here. It would be really cool to have this available.

PR #144, opened 2019-05-16
WIP - Better test strategy

JPinkney:
This PR is a WIP that reduces the setup of tests, tests the content of the element (previously it just checked the return length) and uses local schemas instead of schemas living on servers.


JPinkney:
Closed because its going to come in with https://github.com/redhat-developer/yaml-language-server/issues/142

PR #148, opened 2019-05-28
Deprecate kedge

JPinkney:
Since kedge is not very activate anymore we can deprecate

see: https://github.com/redhat-developer/vscode-yaml/issues/185


fbricon:
That's not how deprecating works ;-)

PR #155, opened 2019-06-12
Improve code clarity

Vidminas:
No code behaviour changes:
* Documented server.ts with comments for code clarity
* Used linter to annotate more types and fix string quote inconsistencies
* Added tslint whitespace rules and fixed whitespace inconsistencies

Should be functionally identical:
* Simplified onInitialize and schemaRequestService functions

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -9,7 +9,7 @@
 import {
     createConnection, IConnection,
     TextDocuments, TextDocument, InitializeParams, InitializeResult, NotificationType, RequestType,
-    Disposable, Position, ProposedFeatures, CompletionList, DocumentRangeFormattingRequest
+    Disposable, Position, ProposedFeatures, CompletionList, DocumentRangeFormattingRequest, ClientCapabilities, WorkspaceFolder
 } from 'vscode-languageserver';
 
 import { xhr, XHRResponse, configure as configureHttpRequests, getErrorStatusDescription } from 'request-light';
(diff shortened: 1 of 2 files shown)


src/server.ts
```diff
@@ -553,60 +605,69 @@ function completionHelper(document: TextDocument, textDocumentPosition: Position
...
+    if (!document) {
         return Promise.resolve(void 0);
     }
 
     const jsonDocument = parseYAML2(document.getText());
     return customLanguageService.doHover(jsonLanguageService, document, textDocumentPositionParams.position, jsonDocument);
 });
 
+/**
+ * Called when the code outline in VS Code needs to be populated
```
JPinkney: I don't think we can explicitly say VS Code here because there are multiple clients available (Eclipse Che, Theia, Atom, etc) that all use this language server

src/server.ts
```diff
@@ -553,60 +605,69 @@ function completionHelper(document: TextDocument, textDocumentPosition: Position
...
 }
 
+/**
+ * Like onCompletion, but called only for currently selected completion item
+ * Provides additional information about the item, not just the keyword
+ */
 connection.onCompletionResolve(completionItem => customLanguageService.doResolve(completionItem));
 
+/**
+ * Called when the user hovers with their mouse over a keyword in VS Code
```
JPinkney: I think you can remove VS Code in this instance, and just keep at as user hovers over a keyword

src/server.ts
```diff
@@ -485,40 +524,49 @@ function validateTextDocument(textDocument: TextDocument): void {
...
     });
+
     if (hasChanges) {
         documents.all().forEach(validateTextDocument);
     }
 });
 
-connection.onCompletion(textDocumentPosition =>  {
+/**
+ * Called when auto-complete is triggered in VS Code
```
JPinkney: Remove VS Code here as well

src/server.ts
```diff
@@ -485,40 +524,49 @@ function validateTextDocument(textDocument: TextDocument): void {
...
             diagnostics.push(diagnosticResults[diagnosticItem]);
         }
 
         connection.sendDiagnostics({ uri: textDocument.uri, diagnostics: removeDuplicatesObj(diagnostics) });
-    }, function (error){});
+    }, function (error) { });
 }
 
+/**
+ * Called when a monitored file is changed in VS Code
```
JPinkney: Here as well

PR #156, opened 2019-06-13
Fix relative paths and schema changes

### LINKED ISSUE #154: Relative path schema loading
With a custom schema configuration in the workspace, such as

```
"yaml.schemas" : {
      "./json_schema/my_schema.json" : "/*.yaml"
    }
```

The schema fails to load with the output message:

> [Error - 2:17:09 PM] Notification handler 'workspace/didChangeConfiguration' failed with message: [UriError]: Scheme is missing: {scheme: "", authority: "", path: "./json_schema/my_schema.json", query: "", fragment: ""}

This does not happen with the npm v0.4.1 yaml-language-server package, only with the latest GitHub build

Vidminas:
Fixes #154 as described in the issue


JPinkney:
I won't be able to review it for a few hours, but does it work when you're doing something like ../../my_schema.json?

We may need to make some more schema tests to cover resolution for relative and absolute cases

Vidminas:
Seems to work when trying manually, although the server does not react to file system changes (e.g. if the schema file is deleted). Touching the config file is enough for it to pick up on the change. Probably expected behaviour, given how the change is external to the editor?

Will see about adding some tests to cover schema paths 👍

Vidminas:
Turns out the tests were passing on my machine but not on Travis, because the node path module behaves differently on Windows and Linux - on Linux it does not consider \ to be valid separators and thus does not normalise paths with them 🤔 I've rewritten tests to dynamically pick just one for a platform, but this might not be the best solution.

I could also not think of a good way to write tests for schema reloading - that a schema should be updated in memory if changed on disk.

JPinkney:
Its probably safer to use something like path.join() for the tests e.g.: https://github.com/redhat-developer/vscode-yaml/blob/master/src/extension.ts#L29. That way regardless of the platform it should automatically detect which slashes should be used

PR #157, opened 2019-06-13
Implemented API for modifying and deleting parts of schemas in memory

JPinkney:
This API introduces the ability to modify and delete certain parts of schema in memory:

#### If you are using the yaml language server as a library itself then you can use these methods that are available on the yaml language service:

```typescript
addSchema (schemaID: string, schema: JSONSchema);
deleteSchema(schemaID: string);
modifySchemaContent(schemaAdditions: SchemaAdditions);
deleteSchemaContent(schemaDeletions: SchemaDeletions);
```

#### If you are calling from a client (such as VSCode):

Modify can be used as:
```typescript
client.sendRequest(SchemaModificationNotification.type, {
			action: MODIFICATION_ACTIONS.add,
			path: 'oneOf/0/properties/kind',
			key: 'enum',
			content: [
				"hello_world",
				"i have overridden some enums"
			],
			schema: 'kubernetes'
		});
```

where action is the type (in this case add), because we are adding in json into the schema. Path is the path to the object you want to modify. Key is the key of the item that you want to modify, content is the new value, and the schema is which schema you want to modify.

Delete can be used as:
```typescript
client.sendRequest(SchemaModificationNotification.type, {
			action: MODIFICATION_ACTIONS.delete,
			path: 'oneOf/1/properties/kind',
			schema: 'kubernetes',
			key: 'enum'
		});
```

where action is the type (in this case delete), because we are deleting json from the schema. Path is the path to the object you want to delete. Key is the key of the item that you want to delete, and the schema is which schema you want to modify.

In order to set the path you will have to know a bit about the schema otherwise errors will be thrown. Also, when you are selecting a path that leads to a JSON array you must select an item in the array by following the JSON array with a number i.e. `path: 'oneOf/1'`. Likewise, if it leads to a JSON object, select a key inside of the object.


src/languageservice/services/jsonSchemaService.ts
```diff
@@ -385,11 +386,31 @@ export class JSONSchemaService implements IJSONSchemaService {
...
+        this.schemasById[id] = new SchemaHandle(this, schemaId, schemaContent);
+        return Promise.resolve(undefined);
+    }
+
+    /**
+     * Delete a schema with schema ID.
+     */
+    public async deleteSchema(schemaId: string): Promise<void> {
+        const id = this.normalizeId(schemaId);
+        delete this.schemasById[id];
```
gorkem: Check if `this.schemasById[id]` is not null before deleting?

src/languageservice/apis/schemaModification.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+export interface SchemaDeletions {
+    schema: string,
+    action: MODIFICATION_ACTIONS.delete,
+    path: string,
+    key: string
+}
+
+export class SchemaModification {
```
gorkem: I am not sure why we have a new `api` folder for this file. `SchemaModification` is not really an API for the 3rd parties. If I am reading this correctly `YAMLSchemaService` together with `SchemaAdditions` and `SchemaDeletions` are the public 3rd party APIs for the functionality.

src/languageservice/utils/kubernetesResolver.ts
```diff
@@ -0,0 +1,12 @@
+export const KUBERNETES_SCHEMA_URL = 'https://raw.githubusercontent.com/garethr/kubernetes-json-schema/master/v1.14.0-standalone-strict/all.json';
```
gorkem: What is the reason to introduce this file with this PR? Feels like it could be its own PR.

PR #167, opened 2019-07-23
Make prettier an optional dependency since it doesn't work with webpack

JPinkney:
This PR makes prettier an optional dependency because prettier does not play well with webpack.


JPinkney:
I'm going to go ahead and merge this and release 0.5.3.

PR #170, opened 2019-07-26
Add WorkspaceFilesRequest

Vidminas:
This add the WorkspaceFilesRequest type to `requestTypes.ts`, to match the behaviour added to `vscode-yaml` [(PR#194)](https://github.com/redhat-developer/vscode-yaml/pull/194) that allows to send a WorkspaceFilesRequest and receive a list of absolute paths to all YAML files in the workspace.

Usage:
```
import { WorkspaceFilesRequest } from './requestTypes';

connection.sendRequest(WorkspaceFilesRequest.type, undefined)
.then((yamlFiles: string[]) => {
    // do something with yamlFiles
});
```

### DIFF
--- a/src/requestTypes.ts
+++ b/src/requestTypes.ts
@@ -23,3 +23,7 @@ export namespace CustomSchemaRequest {
 export namespace ColorSymbolRequest {
     export const type: RequestType<{ }, { }, { }, { }> = new RequestType('json/colorSymbols');
 }
+
+export namespace WorkspaceFilesRequest {
+    export const type: RequestType<{ }, { }, { }, { }> = new RequestType('custom/workspace/yamlfiles');
+}


gorkem:
TBH I am not comfortable with adding a new custom API without an explanation of use cases and real consumers. Can you point me to the information how this is going to be used?

Vidminas:
That's fair enough.

This request type is useful in enhancing autocomplete to suggest the names of other YAML files in the workspace. I'm using it to help write comments more easily (for example, in a comment, I can write "link:" and it shows filenames to autocomplete. Next, I'd also like to implement right-click navigation to those files.

It's not something everyone might want to use, so I can suggest these options:
* the whole feature and request type could live just on my fork
* the request type could be in this language server repo, but without a particular implementation
* a setting could be introduced to toggle it (for example, for this particular use case, it could be a setting that sets a custom prefix for referring to files and if none is set, the feature is off completely)

What do you think?

gorkem:
I do not think this is the right place to host this feature. Another option would be to offer this enhancement to the https://github.com/microsoft/language-server-protocol and see if it can be considered for a future release of the protocol. 

I am closing this one. 
@JPinkney Can you also revert the corresponding PR on the server repo?

PR #189, opened 2019-10-04
Add support for loading schemas stored as YAML files

### LINKED ISSUE #186: YAML Schema files not correctly parsed
We store our JSON Schemas as YAML files.
I have configured my workspace's `settings.json` as follows
```
{
    "yaml.schemas": {
        "./schema.d/myschema.yaml": "/myfile.yaml"
    }
}
```
However, when I load `myfile.yaml` I get the following error: `Unable to parse content from '/path/to/schema.d/myschema.yaml'`

I would assume the issue lies in https://github.com/redhat-developer/yaml-language-server/blob/master/src/languageservice/services/jsonSchemaService.ts#L394-L413

Would a pull request to add support for YAML representations of schemas be considered? If so I am happy to attempt to add this.

hreeder:
I've added rudimentary support for loading schemas stored as YAML.
The logic here is to attempt to load as JSON first. If loading as JSON fails, attempt loading as YAML. If YAML loading succeeds, clear the list of errors. If YAML loading fails then add the YAML error to the errors array.

Fixes #186
Fixes redhat-developer/vscode-yaml#182

### DIFF
--- a/src/languageservice/services/jsonSchemaService.ts
+++ b/src/languageservice/services/jsonSchemaService.ts
@@ -6,6 +6,7 @@
 'use strict';
 
 import * as Json from 'jsonc-parser';
+import * as yaml from 'js-yaml';
 import {JSONSchema, JSONSchemaMap} from '../jsonSchema04';
 import { URI } from 'vscode-uri';
 import * as Strings from '../utils/strings';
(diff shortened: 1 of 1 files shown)


hreeder:
I've just seen #187 which deprecates this PR against `yaml-language-server`. I'll attempt to bring this functionality in upstream in `vscode-json-languageservice`

JPinkney:
We can still probably get this in I think if vscode-json-languageservice doesn't want this implemented on their end. That PR just extends the json schema service from vscode-json-languageservice so we don't run into https://github.com/redhat-developer/vscode-yaml/issues/218.

But what we could do is basically override the load schema thats coming from the vscode-json-languageservice and provide the implementation you had instead

hreeder:
Yes, that's looking like what I'll bring in @JPinkney - I'll have a play around and get something together over the next couple of days.

Would you recommend just overwriting the entire `loadSchema` function, or would you still delegate the attempt to load JSON to the JSONSchemaService?

JPinkney:
You can try delegating the json loading to the JSONSchemaService and then if that doesn't work just try overwriting the entire `loadSchema` function

PR #191, opened 2019-10-08
Add support for loading YAML schema files

hreeder:
This is the re-write of #189 to better make use of the changes introduced whereby JSON schema are loaded as part of the `vscode-json-languageservice` package rather than duplicate that functionality here.

A tradeoff has been made in that the YAML file will be queried twice before data is read however, as the logic relies on the full JSON parsing function to complete before then attempting the same flow but parsing as YAML instead.


JPinkney:
When you get a chance can you write a few tests for this

gregnr:
@hreeder Have you had any progress with the tests for this? We require the same YAML schema parsing functionality with [`app-config`](https://github.com/servall/app-config).

evidolob:
Changes of this PR is merged in https://github.com/redhat-developer/yaml-language-server/pull/318

Issue #150, opened 2019-06-10
Make auto completion use JSON Schema 7 parser


JPinkney:
Currently auto completion uses the JSON Schema 4 parser instead of the JSON Schema 7 parser.

There are a few things to do:
- [x] Make necessary modifications to JSON Schema 7 parser so auto completion will work with it
- [x] Make auto completion use the JSON Schema 7 parser
- [x] Remove JSON Schema 4 parser

ToonvanStrijp:
@JPinkney For my own library I've implemented JSON schema 7 parser

I just changed all the code to compile and changed the parser to 7 etc. Feel free to use this code and improve it. I didn't test all features yet so that's something that still needs to be done.

[yamlCompletion.ts](https://gist.github.com/ToonvanStrijp/44b8f96f934ee18dce6a96b887beca64)

JPinkney:
That's awesome! Do you mind submitting as a PR and I can work off your changes

ToonvanStrijp:
@JPinkney sure no problem. I'll create one in a few hours or so.

PR #198, opened 2019-11-06
yaml completion switch to jsonParser07

ToonvanStrijp:
This PR is a implementation to support json 7 schemas.

This PR will solve #150 eventually :)

Some features are probably missing and we'll have to implement those.

(I didn't fully test this code yet, so It probably needs some modifications to work properly)


gorkem:
@JPinkney I think we can close this one

PR #209, opened 2019-12-09
First draft of server side snippets

JPinkney:
This is the first draft of server side snippet support.

The basic idea is that we want the snippets to produce valid yaml without or with a schema involved. In the first case, without a schema, it's pretty simple. You can just ask the user who is adding/creating the snippet which context that the snippet should appear in:
```
export enum SnippetContext {
  any = 'any',
  scalar = 'scalar',
  object = 'object',
  array = 'array'
}
```
and then when you are performing auto completion you can determine the context of your current cursor position and grab a list of snippets associated with that context.

Showing snippets when you have a schema associated with the yaml file is a lot more complicated. This is because you have to determine whether the snippet matches the context of the schema and the context of the autocompletion.

E.g. if you want a snippet for "{hello: $1\ngoodbye: $2}" we need to determine what we are inserting is an object and then that also matches up with an object in the schema. You would then need to check if additionalProperties is true or false in the schema and if the resulting 'hello' and 'goodbye' are in the properties of the object.

When a schema is associated with a yaml file you need to make sure whatever snippet you insert doesn't lead the yaml file to have an error

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -6,16 +6,18 @@
 'use strict';
 
 import * as Parser from '../parser/jsonParser04';
-import { parse as parseYAML } from '../parser/yamlParser04';
 import * as Json from 'jsonc-parser';
 import { YAMLSchemaService } from './yamlSchemaService';
 import { JSONSchema } from '../jsonSchema04';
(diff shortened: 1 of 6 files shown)


gorkem:
Should we conclude #157 before completing this one?

JPinkney:
@gorkem Yeah, #157 will be finished first

JPinkney:
This can be closed as defaultSnippets has been implemented, allowing you to specify snippets that via json schema!

Issue #180, opened 2019-09-08
Mergekeys not allowed when JSON schema's are used


keesschollaart81:
When validating a YAML file with a JSON schema like this:

```JSON
{
    "type": "object",
    "properties": {
        "prop1": {
            "type": "string"
        },
        "prop2": {
            "type": "string"
        }
    },
    "additionalProperties": false,
    "required": [
        "prop1"
    ],
    "$schema": "http://json-schema.org/draft-07/schema#"
}
```

It is not possible to use merge keys, it says that '<<' is not a valid property:

![image](https://user-images.githubusercontent.com/6755359/64485682-9363f280-d223-11e9-837f-7e1e718e5c37.png)

Would it be possible to ignore '<<' properties when validating schema's?

JohannLai:
I had the same problem.  😭 
![image](https://user-images.githubusercontent.com/10769405/64498868-18fda600-d2e9-11e9-96bf-56c0ebbc0441.png)

nfriend:
Related: https://github.com/redhat-developer/vscode-yaml/issues/233

mapleeit:
Hi, @JPinkney I noticed that the 0.6.0 already been released. It seems that this version didn't fix this bug.

ssbarnea:
That is a critical bug as use of anchors is almost a guarantee today. It should not be a big deal to fix it because it involves pre-processing the file. The only downside is that reported errors from the schema may not have a line number to reffer to.

JPinkney:
I think I have a fix for this that should make it into the next release, when I was switching everything from using a JSON 4 parser to a JSON 7 parser I missed a part of code that processes merge keys correctly

ssbarnea:
Please do not close bugs without having at least a pre-release. Are there any instructions on how to test master? I may be able to give faster feedback on incoming changes.

JPinkney:
Github auto closed it when that PR was merged. I can reopen it until a version is released.

I'll have to release a new version of vscode-yaml and yaml-language-server before everyone can benefit but if you want to try out the changes before release you can follow: https://github.com/redhat-developer/vscode-yaml#developing-the-client-and-server-together. Depending on how much time I have I might be able to create a pre-release of vscode-yaml in the next few days

JPinkney:
The new version has been released so I'm closing.

PR #222, opened 2020-01-21
Fixed merge key error with JSON schema

JPinkney:
This new code is a port from the json schema 4 parser to the new json schema 7 parser. It basically replaces the merge key with the actual values of what the node value points to.

Fixes: https://github.com/redhat-developer/yaml-language-server/issues/180 and https://github.com/redhat-developer/vscode-yaml/issues/233


gorkem:
Can we cover this case with a test?

PR #227, opened 2020-01-24
Made yamlSchemaService accept multiple uris from requestCustomSchema

JPinkney:
This PR allows the customSchemaProvider to receive multiple uri's, load them, and combine them into one giant schema. The giant schema is a oneOf schema. This technically means that multiple contributors can match a specific file and send a schema back to the language server but only the schema that's matching the current file contents will be used!

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -14,7 +14,7 @@ import * as nls from 'vscode-nls';
 import { convertSimple2RegExpPattern } from '../utils/strings';
 const localize = nls.loadMessageBundle();
 
-export declare type CustomSchemaProvider = (uri: string) => Thenable<string>;
+export declare type CustomSchemaProvider = (uri: string) => Thenable<string | string[]>;
 
 export enum MODIFICATION_ACTIONS {
     'delete',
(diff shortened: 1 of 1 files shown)


src/languageservice/services/yamlSchemaService.ts
```diff
@@ -14,7 +14,7 @@ import * as nls from 'vscode-nls';
 import { convertSimple2RegExpPattern } from '../utils/strings';
 const localize = nls.loadMessageBundle();
 
-export declare type CustomSchemaProvider = (uri: string) => Thenable<string>;
+export declare type CustomSchemaProvider = (uri: string) => Thenable<string[]>;
```
gorkem: Do we need to consider compatibility with the current clients? How about making this return string or string array to keep the compatibility??
```
export declare type CustomSchemaProvider = (uri: string) => Thenable<string | string[]>;
```

src/languageservice/services/yamlSchemaService.ts
```diff
@@ -241,25 +241,40 @@ export class YAMLSchemaService extends JSONSchemaService {
...
-                        })
-                       .then(schema => schema, err => resolveSchema());
+                            };
+                        }, err => resolveSchema());
         } else {
             return resolveSchema();
         }
     }
 
+    private resolveCustomSchema(schemaUri, doc) {
```
gorkem: convert to async function for readability.

PR #241, opened 2020-02-25
Issue 118 null literals

timo-klarshift:
This PR parses Null literals correctly and stated by the YAML spec: https://yaml.org/type/null.html

### DIFF
--- a/src/languageservice/parser/yamlParser07.ts
+++ b/src/languageservice/parser/yamlParser07.ts
@@ -115,6 +115,13 @@ function recursivelyBuildAst(parent: ASTNode, node: Yaml.YAMLNode): ASTNode {
                 return new BooleanASTNodeImpl(parent, parseYamlBoolean(value), node.startPosition, node.endPosition - node.startPosition);
             }
 
+            //This is patch to support YAML null values: https://yaml.org/type/null.html
+            // TODO: fix null values // TODO: also add <empty>
+            const possibleNullValues = ['~', 'NULL', 'Null', 'null'];
+            if (instance.plainScalar && possibleNullValues.indexOf(value.toString()) !== -1) {
+                return new NullASTNodeImpl(parent, node.startPosition, node.endPosition - node.startPosition);
+            }
+
             switch (type) {
                 case Yaml.ScalarType.null: {
                     return new StringASTNodeImpl(parent, instance.startPosition, instance.endPosition - instance.startPosition);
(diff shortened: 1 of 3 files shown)


test/schemaValidation.test.ts
```diff
@@ -96,7 +96,9 @@ suite('Validation Tests', () => {
                 }).then(done, done);
             });
 
-            it('Null scalar value should be treated as string', done => {
+            // TODO: remove this test, because it is just plain wrong: https://yaml.org/type/null.html
+            // See: https://github.com/redhat-developer/yaml-language-server/issues/118
+            it.skip('Null scalar value should be treated as string', done => {
```
JPinkney: I think instead of skipping we should probably just removed it at this point since that's technically a violation of the spec

src/languageservice/parser/yamlParser07.ts
```diff
@@ -115,6 +115,12 @@ function recursivelyBuildAst(parent: ASTNode, node: Yaml.YAMLNode): ASTNode {
                 return new BooleanASTNodeImpl(parent, parseYamlBoolean(value), node.startPosition, node.endPosition - node.startPosition);
             }
 
+            //This is patch to support YAML null values: https://yaml.org/type/null.html
+            const possibleNullValues = ['~', 'NULL', 'Null', 'null'];
+            if (instance.plainScalar && possibleNullValues.indexOf(value.toString()) !== -1) {
```
JPinkney: I wonder if it makes sense to just to change:
```typescript
case Yaml.ScalarType.null: {
      return new StringASTNodeImpl(parent, instance.startPosition, instance.endPosition - instance.startPosition);
}
```

to 
```typescript
case Yaml.ScalarType.null: {
     return new NullASTNodeImpl(parent, node.startPosition, node.endPosition - node.startPosition)
}
```

JPinkney:
@itowlson Just wanted to cc you in on this because I think the kubernetes schema would need to be updated. In cases in the kubernetes schema where a value can be a null or string you'll have to update the type.

Currently, `null` is parsed as a string but that violates the yaml spec so its going to go back to parsing `null` as a NullAstNode and now `null` will match json schemas `"type": "null"`

test/schemaValidation.test.ts
timo-klarshift: Yes! I just left it there to not be too destructive, wanted someone to double check. I will remove it.

src/languageservice/parser/yamlParser07.ts
timo-klarshift: I will check if the tests succeed by doing so.

timo-klarshift:
Just out of interest, when will this be released and upstreamed to vscode-yaml?

JPinkney:
@timo-klarshift Probably not for a couple of weeks

PR #245, opened 2020-03-21
yaml completion: add a space between ':' and value

fbaligand:
Fix the missing space between ':' and value in YAML completion, especially for default values, but not only.
Fix issue https://github.com/redhat-developer/vscode-yaml/issues/281

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -673,7 +673,7 @@ export class YAMLCompletion extends JSONCompletion {
                 if (propertySchema.defaultSnippets.length === 1) {
                     const body = propertySchema.defaultSnippets[0].body;
                     if (isDefined(body)) {
-                        value = ' ' + this.getInsertTextForSnippetValue(body, '', {
+                        value = this.getInsertTextForSnippetValue(body, '', {
                             newLineFirst: true,
                             indentFirstObject: false,
                             shouldIndentWithTab: false
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -673,7 +673,7 @@ export class YAMLCompletion extends JSONCompletion {
                 if (propertySchema.defaultSnippets.length === 1) {
                     const body = propertySchema.defaultSnippets[0].body;
                     if (isDefined(body)) {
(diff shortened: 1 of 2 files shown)


fbaligand:
@JPinkney 
Concerning the failed checks:
- CI build: first job just didn't run because an error occurred while generating the build script.
- But CI build second job run successfully!
- I'm surprised that coverage has decreased, given that there is no added code line, just trailing spaces.

JPinkney:
@fbaligand The PR looks good to me! Do you mind writing a couple small tests so that we can make sure this issue doesn't happen again

fbaligand:
Ok to add a test for default value. 
What is best existing .ts test file where it would be the more relevant?

JPinkney:
Any of autocompletion ones should work. The only difference between them is the schemas that they use. If none of them can be used (if they don't provide a schema that allows you to call the correct sections of the code) you can do what autoCompletion5 did and manually create a schema in the fixtures folder and just create autoCompletion6.ts. The whole testing structure of those files needs to be revamped so I'd be ok to just add another file if needed

fbaligand:
Hi,

Well, I'm quite bothered and need help for my unit test...
Actually, I added a test, trying to do exactly as other unit tests. 
My test is inspired by `autoCompletion.test.ts` and `autoCompletion5.test.ts`.
But despite that, the result is not at all what I expect, and I trully don't understand why...

I expect to have only one result item, that is 'directory' entry with `directory: bower_components` completion.
But for real, I get all of the 19 possible yaml entries, starting with `analytics`.

So can you tell me what to do so that my test works?
Or directly fix it in my branch?

Thanks in advance. I can assure you that I spent some time to try to make this test work.
I ask for your help, really because I don't understand and don't see a solution.

JPinkney:
I think what you can do for now is:
```
assert.equal(result.items[2].insertText, 'directory: ${1:bower_components}');
```
and the tests will start working. It looks like during the completion request everything gets sent back to the client and its not until the completion resolve request that it narrows down to directory

fbaligand:
Hi @JPinkney,

Thanks a lot for your help!
I applied your tip and it works great!
Now all tests are OK.

Tell me if there is something else to add/change.

fbaligand:
Great!
Thanks @JPinkney for the merge!

Issue #256, opened 2020-04-30
Defaults for didChangeConfiguration Messages


Trajamsmith:
Hey — we're trying to extend the functionality of [JupyterLab](https://github.com/jupyterlab/jupyterlab) to include all the features that language servers provide. At present, we're taking a look at server configuration messages. Right now, our `onDidChangeConfiguration` messages to a few server implementations are working well, but we're having small issues with the `yaml-language-server`.

We want our settings objects in the JupyterLab editor look something like this:
```json
{
    "language_servers": {
        "pyls": {
            "serverSettings": {
                "pyls.plugins.pydocstyle.enabled": true,
                "pyls.plugins.pyflakes.enabled": false,
                "pyls.plugins.flake8.enabled": true
            }
        },
        "r-languageserver": {
            "serverSettings": {
                "r.lsp.debug": true,
                "r.lsp.diagnostics": false
            }
        },
        "yaml-language-server": {
            "serverSettings": {
                "yaml.format.enable": true,
                "yaml.format.singleQuote": true
            }
        }
    }
}
```

We'd like to be able to send partial configurations like the above, but I _believe_ the issue we're having is that the `onDidChangeConfiguration` listener for the YAML server doesn't like partial configurations. 

My best guess is that this is because there are no defaults [here:](https://github.com/redhat-developer/yaml-language-server/blob/f6b89fdb9fedc1de38d6f7da42f0d09e92a41731/src/server.ts#L399)
```js
    if (settings.yaml) {
        yamlConfigurationSettings = settings.yaml.schemas;
        yamlShouldValidate = settings.yaml.validate;
        yamlShouldHover = settings.yaml.hover;
        yamlShouldCompletion = settings.yaml.completion;
        customTags = settings.yaml.customTags ? settings.yaml.customTags : [];

        if (settings.yaml.schemaStore) {
            schemaStoreEnabled = settings.yaml.schemaStore.enable;
        }
...
```

So if we send the configuration settings like I've written them up top, then the server turns off validation, even though my configuration object didn't mention validation.

Is that hunch correct? It's very possible that we're just looking at this the wrong way.

JPinkney:
Yeah it looks like that's correct, partial configurations don't seem to be working correctly

Trajamsmith:
May I submit a PR for this? This fixes the issue on my local:
```js
    // change this
    if (settings.yaml) {
        yamlConfigurationSettings = settings.yaml.schemas;
[...]

PR #258, opened 2020-05-01
Fix Partial Configuration Bug

Trajamsmith:
Small patch to fix [#256](https://github.com/redhat-developer/yaml-language-server/issues/256).

When submitting partial settings objects to the language server using the `onDidChangeConfiguration` listener, the server was overwriting its default values with null values for every unspecified parameter.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -404,10 +404,18 @@ connection.onDidChangeConfiguration(change => {
 
     specificValidatorPaths = [];
     if (settings.yaml) {
-        yamlConfigurationSettings = settings.yaml.schemas || yamlConfigurationSettings;
-        yamlShouldValidate = settings.yaml.validate || yamlShouldValidate;
-        yamlShouldHover = settings.yaml.hover || yamlShouldHover;
-        yamlShouldCompletion = settings.yaml.completion || yamlShouldCompletion;
+        if (settings.yaml.hasOwnProperty('schemas')) {
+            yamlConfigurationSettings = settings.yaml.schemas;
+        }
+        if (settings.yaml.hasOwnProperty('validate')) {
+            yamlShouldValidate = settings.yaml.validate;
+        }
+        if (settings.yaml.hasOwnProperty('hover')) {
+            yamlShouldHover = settings.yaml.hover;
+        }
+        if (settings.yaml.hasOwnProperty('completion')) {
+            yamlShouldCompletion = settings.yaml.completion;
+        }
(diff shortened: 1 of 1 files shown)


src/server.ts
```diff
@@ -404,10 +404,10 @@ connection.onDidChangeConfiguration(change => {
 
     specificValidatorPaths = [];
     if (settings.yaml) {
-        yamlConfigurationSettings = settings.yaml.schemas;
-        yamlShouldValidate = settings.yaml.validate;
-        yamlShouldHover = settings.yaml.hover;
-        yamlShouldCompletion = settings.yaml.completion;
```
Trajamsmith, line 410: These were overwritten if, e.g., `yaml.validate.enabled` was not specified in the configuration message.

JPinkney:
Hmm after looking at it again I think there might be a small bug here:
ex for yamlShouldValidate 

if settings.yaml.validate is undefined it will take the value of yamlShouldValidate :heavy_check_mark: 
if settings.yaml.validate is true it will take that value :heavy_check_mark: 

but
I think if settings.yaml.validate is false it will take the value of yamlShouldValidate if that's true, leading it to not change

I think we might have to do something like `settings.yaml.validate ? settings.yaml.validate : yamlShouldValidate`

That way of settings.yaml.validate is undefined it uses yamlShouldValidate
If settings.yaml.validate is defined and true it will use it
If settings.yaml.validate is defined and false it will use it

Trajamsmith:
You're absolutely right, I got the configuration object confused with that of another language server 😅 I think those ternary operators would still cause a problem though — if `settings.yaml.validate` is false (we _want_ to set it to false) and `yamlShouldValidate` is true, then it will still be true.

What about just checking for the property:
```js
if (settings.yaml.hasOwnProperty('validate') {
  yamlShouldValidate = settings.yaml.validate
}
```

JPinkney:
> if settings.yaml.validate is false (we want to set it to false) and yamlShouldValidate is true, then it will still be true.

Oh right :sweat_smile: 

I think just checking the property should suffice then!

In that case, if its undefined then nothing changes and if it is defined then it changes to whatever it was defined as

Trajamsmith:
Alright, it's a little more verbose but I feel like that should work 👍

Trajamsmith:
Do you all have an idea of when you might next publish an NPM release? Would be great to have these configs available sooner than later, but it's not a big deal either way. Thanks for letting me help!

JPinkney:
Hopefully in the next week or so, there's a decent amount of stuff that needs to be released now

Issue #260, opened 2020-05-23
Unexpected Diagnostic Positions


trajamsmith:
Hey again!

We're running into some strange unexpected behavior surrounding the diagnostic positions since upgrading our project's dependency from `0.4.0 -> 0.8.0` (I know it's a pretty big jump).

These are the expected diagnostic positions when using `0.4.0`:
<img width="413" alt="Screen Shot 2020-05-23 at 11 32 56 AM" src="https://user-images.githubusercontent.com/11686766/82734559-a9e6a700-9ce9-11ea-8902-bb085c43dafa.png">

Upgrading to `0.8.0`, with no other changes at all, the diagnostic positions are no longer correct:
<img width="413" alt="Screen Shot 2020-05-23 at 11 35 48 AM" src="https://user-images.githubusercontent.com/11686766/82734571-c256c180-9ce9-11ea-8169-07d12f751260.png">

I'm poking around the server locally trying to figure out why this might be the case. 

For context, the content of the textDocument is: `greetings:↵    hello:↵    hello:↵`

JPinkney:
It looks like duplicate key for some reason is 1 line off.  I _think_ something in here:
https://github.com/redhat-developer/yaml-language-server/blob/master/src/languageservice/parser/yamlParser07.ts#L190-L205 is causing it

trajamsmith:
I think you're right, I think it's specifically something here with the `convertError` function:
https://github.com/redhat-developer/yaml-language-server/blob/a9169daf6ac7c91273a53c72e322df0707c5e40b/src/languageservice/parser/yamlParser07.ts#L202

I'll do some more poking around and open a PR if I find a fix — if that's alright with you @JPinkney. Thanks for the prompt response!

JPinkney:
@trajamsmith Yeah that works for me!

trajamsmith:
Should be fixed with the #261 merge.

PR #147, opened 2019-05-28
Realign features with JSON Language Server

JPinkney:
This is a massive PR but it aims to align the feature set with the JSON Language Server as much as possible.

Currently the only feature that cannot rely on the JSON Language Server is auto completion. The auto completion is currently too customised to be aligned right now.

Fixes:
https://github.com/redhat-developer/yaml-language-server/issues/142
https://github.com/redhat-developer/yaml-language-server/issues/140
https://github.com/redhat-developer/yaml-language-server/issues/139
https://github.com/redhat-developer/yaml-language-server/issues/137
https://github.com/redhat-developer/yaml-language-server/issues/136
https://github.com/redhat-developer/yaml-language-server/issues/133
https://github.com/redhat-developer/yaml-language-server/issues/105
https://github.com/redhat-developer/vscode-yaml/issues/153
https://github.com/redhat-developer/vscode-yaml/issues/154
https://github.com/redhat-developer/vscode-yaml/issues/130
https://github.com/redhat-developer/vscode-yaml/issues/129
https://github.com/redhat-developer/vscode-yaml/issues/106

PR #261, opened 2020-05-24
Fix Diagnostic Positions

trajamsmith:
See issue #260


trajamsmith:
Moving things around a little in `yamlParser07.ts` to try and make the logic easier to follow. Adding more typing where possible. 

@JPinkney Please let me know if you'd rather I didn't move things around (e.g. extracting helpers into their own util file). I've been running `yarn test` after every change, but if there's more I need to do to prevent regressions, I'm obviously happy to do it!

trajamsmith:
Looking at the pipeline, clearly I should be compiling after every step too — on it!

JPinkney:
Moving around things works for me if it makes the code cleaner :+1: Let me know when this PR is ready to review and i'll take a look!

trajamsmith:
So, as I dig deeper on this, the issue is that the exceptions from `yaml-ast-parser` (or the fork we're using here) are _not very useful_ for constructing the actual language server diagnostics. 

While the exceptions give accurate starting positions (including line and column numbers), there is — as far as I can tell — no good way to reconstruct the exceptions' *ending* positions. I've tried doing so with the `getSnippet()` method from the library, but it does some weird, unhelpful formatting that I'm hesitant to manually undo.

I'm not really sure what the next step is. I'm now getting accurate diagnostics for a few error types, and only a couple of tests are failing right now, but I'm really not confident that trying to hack around the AST parsing library won't just create more problems not reflected in the tests.

I could, of course, be completely overlooking something obvious, but it seems like we'd need to build out a different approach for the diagnostic generation. I'll investigate this further, but diagnostic positions way back in `0.4.0` seemed to be doing fine. I think that was before switching to the `yaml-ast-parser` (could be wrong).

JPinkney:
We've had the yaml ast parser since the beginning and yaml-ast-parser-custom-tags (the one we use) should just be master of yaml-ast-parser with one patch ontop. If I were to guess, the problems arose with this massive PR: https://github.com/redhat-developer/yaml-language-server/pull/147/files. I'll have to take a deeper look.

One noticeable thing I've found is that convert error in 0.4.1 seems to use a much simpler convertError function: https://github.com/redhat-developer/yaml-language-server/blob/0.4.1/src/languageservice/parser/yamlParser.ts#L203. I can't recall why the convertError function was changed and I can't find any associated issue with that change either. So maybe the solution is to just change the convertError function back

JPinkney:
It looks like doing something like:
```javascript
const line = e.mark.line < 0 ? 0 : e.mark.line;
const character = e.mark.position + e.mark.column < 0 ? 0 : e.mark.position + e.mark.column;
```

in convert error might work. The _only_ issue with that code is it seems to be showing the error on the value instead of the key.

```javascript
const line = e.mark.line < 0 ? 0 : e.mark.line;
const character = e.mark.column + 1;
```
might work as well. It seems to show the error on the correct location in my limited testing

trajamsmith:
Sorry for being a little slow with this PR — @JPinkney if you want to take just a quick peek at this, I'm pretty sure that I successfully reverted the diagnostic positions to how they were handled in `0.4.1`. Let me know if anything looks awry.

test/schemaValidation.test.ts
```diff
@@ -203,14 +203,15 @@ suite('Validation Tests', () => {
                 const content = 'cwd: hello\nan';
                 const validator = parseSetup(content);
                 validator.then(function (result) {
+                    console.log(`RESULT: ${JSON.stringify(result)}`)
                     assert.equal(result.length, 2);
                     assert.deepEqual(
                         result[0],
-                        createExpectedError(BlockMappingEntryError, 1, 13, 1, 13)
+                        createExpectedError(BlockMappingEntryError, 1, 2, 1, 2)
```
trajamsmith, line 210: I feel pretty confident that the old test was wrong — the error should be on character two, the line doesn't extend to a thirteenth character.

This is in line with the logging I did on the testing suite in `0.4.1`.

trajamsmith:
_And just to document this here:_
@JPinkney You were totally right that the diagnostics were using the `yaml-ast-parser` without any additional shenanigans.

I didn't realize that the language server spec includes a function that will determine the line and character position from the absolute position in the string, `TextDocument.atPosition()`. Somehow in that big update after `0.4.x`, the line and character positions were no longer being calculated using that method.

test/schemaValidation.test.ts
trajamsmith: image

trajamsmith: ^output from `0.4.1`

test/schemaValidation.test.ts
```diff
@@ -203,14 +203,15 @@ suite('Validation Tests', () => {
                 const content = 'cwd: hello\nan';
                 const validator = parseSetup(content);
                 validator.then(function (result) {
+                    console.log(`RESULT: ${JSON.stringify(result)}`)
```
JPinkney, line 206: Remove

package-lock.json
```diff
@@ -0,0 +1,2046 @@
+{
```
JPinkney, line 1: We no longer use npm, instead we use yarn so I don't think we need this file

JPinkney:
Also it will need to be rebased

test/schemaValidation.test.ts
trajamsmith: Whoops!

trajamsmith:
Implemented the requested changes — let me know if we need anything else for a merge!

I think the positions are still a little weird with some kinds of errors, but reverting to the old working implementation I think will fix a lot of them.

PR #268, opened 2020-07-03
Adding test for autocompletion for Camel K schema

apupier:
created on request https://github.com/redhat-developer/vscode-yaml/issues/310#issuecomment-653164457

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "@types/prettier": "^1.18.0",
     "@typescript-eslint/eslint-plugin": "^3.2.0",
     "@typescript-eslint/parser": "^3.2.0",
+    "chai": "^4.2.0",
     "coveralls": "^3.0.5",
     "eslint": "^7.2.0",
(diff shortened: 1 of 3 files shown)


apupier:
> Is there any reason why we need to bring in chai? I guess its technically easier but adding one dependency for 2 tests seems like a lot

it is technically easier. Chai is a relatively common library used for test assertion (3M+ downloads https://www.npmjs.com/package/chai ). I'm not familiar with TypeScript language, so using this kind of utilities saves me a lot of time.
The dependency is only for test, not at runtime.

apupier:
By the way I opened this PR on request to emphasize that there is a difference between the yaml laanguage server and the VS Code yaml behavior.

i'm not sure that it is worthy to merge it as-is

joshuawilson:
Can we close this? or update it with the intent to merge.

apupier:
this would need to be reworked and no time for that currently. Closing

Issue #269, opened 2020-07-03
Avoid providing completion when "maxProperties" is reached


apupier:
it is possible to specify "maxProperties". So for instance I have a value set to 2, and I can have tag property in [tag1, tag2, tag3] if I have:
```
- tag1:
  tag2:
  
```
And that I call completion on last line, there should be no completion.

nota: Issue detected while working on Apache Camel K schema https://gist.github.com/lburgazzoli/70027ee39196c18008e1a455ef02566c

apupier:
provided a PR with tests to show expected behavior https://github.com/redhat-developer/yaml-language-server/pull/270

apupier:
Pr updated with an implementation respecting the maxProperties value

PR #270, opened 2020-07-03
Avoid providing completion when "maxProperties" is reached #269

apupier:
PR to show which cases would be nice to be supported

### DIFF
--- a/ploparraymax.yaml
+++ b/ploparraymax.yaml
@@ -1,3 +1,0 @@
-- prop1: ded
-- prop2: ded
-- prop3: plop
--- a/ploparraymax.yaml
+++ b/ploparraymax.yaml
@@ -1,0 +1,3 @@
+- prop1: ded
+- prop2: ded
+- prop3: plop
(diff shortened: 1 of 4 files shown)


apupier:
updated PR providing an implementation for maxProperties respect.

please note that I'm not familiar neither with Yaml, Json schema and this code base, so please review carefully.

ploparraymax.yaml
```diff
@@ -0,0 +1,3 @@
+- prop1: ded
+- prop2: ded
+- prop3: plop
```
JPinkney, line 3: Is thie file needed anywhere or was it just used to test?

test/fixtures/testArrayMaxProperties.json
```diff
@@ -0,0 +1,17 @@
...
+                "type": "string"
+            },"prop2": {
+                "type": "string"
+            },"prop3": {
+                "type": "string"
+            }
+        },
+        "maxProperties": 2
+    }
+  }
```
JPinkney, line 17: newline

ploparraymax.yaml
apupier: just for test

apupier: I shouldn't have pushed it!!! my bad. Thanks for pointing it

PR #272, opened 2020-07-08
Fix required attributes insertion with first array item

apupier:
it seems that a `\t` is used instead of 4 spaces. I thought that tabs shouldn't be used in yaml files. To investigate more. @JPinkney does it ring a bell to you?

```
  1) yamlCompletion with required properties
       doComplete
         Insert required attributes at correct level:

      AssertionError [ERR_ASSERTION]: 'top:\n\tprop1: $1' == 'top\n    prop1: $1'
      + expected - actual

      -top:
      -	prop1: $1
      +top
      +    prop1: $1
```

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -250,7 +250,7 @@ export class YAMLCompletion extends JSONCompletion {
                                     // indent
                                     const sourceText = document.getText();
                                     const indexOfSlash = sourceText.lastIndexOf('-', node.offset - 1);
-                                    if (indexOfSlash > 0) {
+                                    if (indexOfSlash >= 0) {
                                         // add one space to compensate the '-'
                                         identCompensation = ' ' + sourceText.slice(indexOfSlash + 1, node.offset);
                                     }
--- a/test/autoCompletion.requiredproperties.test.ts
+++ b/test/autoCompletion.requiredproperties.test.ts
@@ -1,0 +1,58 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Red Hat. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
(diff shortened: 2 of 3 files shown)


JPinkney:
I can't recall \t being used anywhere but I can a look

apupier:
there is 12 places where the \t is used in yamlCompletion.ts.
for instance https://github.com/redhat-developer/yaml-language-server/blob/292c956a3412c73bce5eba0ffbed2968dd26f61f/src/languageservice/services/yamlCompletion.ts#L261

JPinkney:
Yep, just saw those. Those will need to be switched to use whatever the user has set has their number of spaces in the document

apupier:
in this case, isn't the problem that it should be 2 tabs instead of a single one? I think the "array" level is not counted in this specific case.

Based on specification, each level can have a different number of spaces https://yaml.org/spec/1.2/spec.html#id2777534 ( :'-( )  will be funny to search for number of spaces on other elements on the same indentation level, when there is one. If not, which defaults? 2 spaces par indentation level?

apupier:
the problem occurs only when it is the first element in the document. I updated the PR to provide a a fix and test the 2 cases.

@JPinkney I only changed a `>` to a `>=`. Tests are passing. Do you remind if using a strict equality was useful in a specific case?

JPinkney:
>  will be funny to search for number of spaces on other elements on the same indentation level, when there is one 

I was thinking more that if we could find a way to detect what number of spaces the editor was set to use then we could use that, otherwise default to 2. I know that in the lsp tabSize is available for formatting but i'm not sure we can grab that somewhere else and then use that to insert instead of \t but yeah I think this can be done in another PR

> Do you remind if using a strict equality was useful in a specific case?

I don't remember off the top of my head, that code was added a long time ago in this PR:
https://github.com/redhat-developer/yaml-language-server/commit/ad072447741967067652978c11e2f9e418111f86. Since all of the autocompletion5 tests are passing it seems like there isn't a regression anywhere

Issue #273, opened 2020-07-09
Anchor on property which uses alias fails validation


dleavitt:
Schema: 
```json
{
  "title": "Field definitions schema!",
  "$schema": "http://json-schema.org/draft-04/schema#",
  "definitions": {
    "field": {
      "type": "object",
      "additionalProperties": false,
      "required": ["label"],
      "properties": {
        "label": { "type": "string" }
      }
    },
    "model": {
      "type": "object",
      "additionalProperties": false,
      "properties": {
        "props": {
          "type": "object",
          "additionalProperties": { "$ref": "#/definitions/field" }
        }
      }
    }
  },
  "type": "object",
  "additionalProperties": { "$ref": "#/definitions/model" }
}

```

Document:
```yaml
Named:
  props: &NamedProps
    name:
      label: Name

Timestamped:
  props: &TimestampedProps
    <<: *NamedProps
    ts:
      label: Timestamp

Record:
  props:
    <<: *TimestampedProps # commenting out this line removes the errors
    id:
      label: ID
```
Looks like it's incorrectly adding a layer of nesting or something?

![Screenshot 2020-07-08 20 25 57](https://user-images.githubusercontent.com/71686/86993686-7ce73a00-c159-11ea-82bf-163f48465a52.png) ![Screenshot 2020-07-08 20 25 51](https://user-images.githubusercontent.com/71686/86993687-7e186700-c159-11ea-8026-e665c9ec6305.png)

JPinkney:
From what I remember the way the merge key works in the code is it basically replaces ```<<``` with the value so 
```
Timestamped:
  props: &TimestampedProps
    <<: *NamedProps
    ts:
      label: Timestamp
```
becomes

```
Timestamped:
  props: &TimestampedProps
    name:
      label: Name
    ts:
      label: Timestamp
```

and then 
```
Record:
  props:
    <<: *TimestampedProps
    id:
      label: ID
```

_should_ become 

```
Record:
  props:
    name:
      label: Name
    ts:
      label: Timestamp
    id:
      label: ID
```

but my guess is somewhere along the lines something is going wrong but then again anchors are the piece of the spec I understand the least

dleavitt:
Yep, and that's what it does become when parsed. It's just an issue with the schema validation, from what I can tell.

JPinkney:
I think I've finally figured out what the issue is here. I _think_ only the first merge key is being transformed and not the second one so the representation to the parser is:
```
Named:
[...]

PR #276, opened 2020-07-13
failing test for anchor that contains an alias

dleavitt:
failing test for #273


evidolob:
We already has test for anchor and aliases

PR #280, opened 2020-07-15
204 support schema associations in file

apupier:
Choose to use modeline with pattern `# yaml-language-server: $schema=<urlToTheSchema>`

Why?
- needs to make a choice :-)
- using directive likes using `#SCHEMA` for instance is creating an invalid file with current specifications state, so need to move to a comment-like `#`
- use `$schema` because it is what is used by VS Code for JSON schema

completionUsingInlinedSchemaDeclaration

big advantages of allowing defining the schema in the file is that the filename can be whatever is needed. It avoids the requirement of specific file name pattern. It ensures users can use a schema in all IDEs, no need for a specific IDE integration (apart from relying on Yaml Language Server of course).

For instance in OpenShift dev console, there is no way for user to choose the filename. it would allow to add have completion/validation there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,10 @@ and that will associate the composer schema with myYamlFile.yaml.
 
 ## More examples of schema association:
 
+### Using yaml.schemas settings
+
+#### Single root schema association:
+
 When associating a schema it should follow the format below
 ```
 yaml.schemas: {
@@ -118,7 +122,7 @@ yaml.schemas: {
 }
 ```
 
-## Multi root schema association:
+#### Multi root schema association:
 You can also use relative paths when working with multi root workspaces.
 
 Suppose you have a multi root workspace that is laid out like:
@@ -140,6 +144,14 @@ yaml.schemas: {
 ```
 
 `yaml.schemas` allows you to specify json schemas that you want to validate against the yaml that you write. Kubernetes is an optional field. It does not require a url as the language server will provide that. You just need the keyword kubernetes and a glob pattern.
+
+### Using inlined schema
+
+It is possible to specify a yaml schema using a modeline.
+
+```
+# yaml-language-server: $schema=<urlToTheSchema>
+```
 
 ## Clients
 This repository only contains the server implementation. Here are some known clients consuming this server:
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -17,7 +17,8 @@ import { convertSimple2RegExpPattern } from '../utils/strings';
 import { TextDocument } from 'vscode-languageserver';
 import { SingleYAMLDocument } from '../parser/yamlParser07';
 import { stringifyObject } from '../utils/json';
-import { getNodeValue } from '../parser/jsonParser07';
+import { getNodeValue, JSONDocument } from '../parser/jsonParser07';
+import { Parser } from 'prettier';
 const localize = nls.loadMessageBundle();
 
 export declare type CustomSchemaProvider = (uri: string) => Thenable<string | string[]>;
@@ -218,7 +219,7 @@ export class YAMLSchemaService extends JSONSchemaService {
     }
     //tslint:enable
 
-    public getSchemaForResource (resource: string, doc = undefined): Thenable<ResolvedSchema> {
+    public getSchemaForResource (resource: string, doc : JSONDocument): Thenable<ResolvedSchema> {
         const resolveSchema = () => {
             const seen: { [schemaId: string]: boolean } = Object.create(null);
             const schemas: string[] = [];
@@ -253,8 +254,8 @@ export class YAMLSchemaService extends JSONSchemaService {
 
             if (schemas.length > 0) {
                 return super.createCombinedSchema(resource, schemas).getResolvedSchema().then(schema => {
-                    if (schema.schema && schema.schema.schemaSequence && schema.schema.schemaSequence[doc.currentDocIndex]) {
-                        return new ResolvedSchema(schema.schema.schemaSequence[doc.currentDocIndex]);
+                    if (schema.schema && schema.schema.schemaSequence && schema.schema.schemaSequence[(<SingleYAMLDocument>doc).currentDocIndex]) {
+                        return new ResolvedSchema(schema.schema.schemaSequence[(<SingleYAMLDocument>doc).currentDocIndex]);
                     }
                     return schema;
                 });
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -305,7 +305,7 @@ export class YAMLSchemaService extends JSONSchemaService {
             if (yamlLanguageServerModeline != undefined) {
                 const schemaKey = '$schema=';
                 const indexOfJsonSchemaParameter = yamlLanguageServerModeline.indexOf(schemaKey);
-                if (yamlLanguageServerModeline.indexOf(schemaKey) != -1) {
+                if (indexOfJsonSchemaParameter !== -1) {
                     const startIndex = indexOfJsonSchemaParameter + schemaKey.length;
                     const indexOfNextSpace = yamlLanguageServerModeline.indexOf(' ', startIndex);
                     const endIndex = indexOfNextSpace !== -1 ? indexOfNextSpace : yamlLanguageServerModeline.length;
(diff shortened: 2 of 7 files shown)


apupier:
@JPinkney @gorkem Can you review it and tell what you think please about this approach please?

test/autoCompletion.withschemainfile.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+            it('Provide completion from schema declared in file', done => {
+                const content = `# yaml-language-server: $schema=${uri}\n- `;
+                const completion = parseSetup(content, content.length);
+                completion.then(function (result) {
+                    assert.equal(result.items.length, 3);
+                }).then(done, done);
+            });
+            
+            it('Provide completion from schema declared in file with several attributes', done => {
```
JPinkney, line 42: Is there any way you can rebase on master and integrate these tests into the main autoCompletion.test.ts file? And also check the actual values of result.items. Also can you add another test for completion with yaml-language-server modelline in multiple documents. I've been trying to harden the tests lately

README.md
```diff
@@ -141,6 +145,14 @@ yaml.schemas: {
 
 `yaml.schemas` allows you to specify json schemas that you want to validate against the yaml that you write. Kubernetes is an optional field. It does not require a url as the language server will provide that. You just need the keyword kubernetes and a glob pattern.
 
+### Using inlined schema
```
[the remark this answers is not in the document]
apupier: no, it is to be at same level to `Using yaml.schemas settings`

test/autoCompletion.withschemainfile.test.ts
apupier: I added a test with multiple documents. it seems there is no "inserttext" which is added:
(node:25674) UnhandledPromiseRejectionWarning: AssertionError [ERR_ASSERTION]: Expected values to be loosely deep-equal:

```
{
  documentation: '',
  insertText: 'prop1: $1',
  insertTextFormat: 2,
  kind: 10,
  label: 'prop1'
}

should equal

{
  documentation: '',
  insertText: 'prop1: $1',
  insertTextFormat: 2,
  kind: 10,
  label: 'prop1',
  textEdit: {
    newText: 'prop1: $1',
    range: {
      end: {
        character: 2,
        line: 0
      },
      start: {
        character: 2,
        line: 0
      }
    }
  }
}
```

I don't get why it woudl be specific to what I changed. I will have a look tomorrow.

src/languageservice/services/yamlSchemaService.ts
```diff
@@ -216,9 +220,15 @@ export class YAMLSchemaService extends JSONSchemaService {
 
     public getSchemaForResource (resource: string, doc = undefined): Thenable<ResolvedSchema> {
```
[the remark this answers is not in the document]
apupier: it is Parser;JSONDocument https://github.com/microsoft/vscode-json-languageservice/blob/908ede3fd3c3b1e1c2f5f9b96ad9c4056e8fb1d0/src/services/jsonSchemaService.ts#L531

gorkem:
@apupier Is there a reason why we can not use a pattern such as `# $schema=<urlToTheSchema>`

src/languageservice/services/yamlSchemaService.ts
```diff
@@ -282,6 +293,28 @@ export class YAMLSchemaService extends JSONSchemaService {
...
         }
     }
+    
+    /**
+     * Retrieve schema if declared as modeline
+     * @param doc 
+     */
+    private getSchemaFromModeline(doc: any) : string{
+        if (doc instanceof SingleYAMLDocument) {
+            const modelineDeclaration = '# yaml-language-server:';
```
gorkem, line 303: What if I put 2 spaces between `#` and `yaml-language-server:`? Do we want it to be this strict? 

I wonder if tokenizing the line with something like  `string.split("/\b\s+(?!$)/");` would result on a more flexible implementation.

src/languageservice/services/yamlSchemaService.ts
```diff
@@ -282,6 +293,28 @@ export class YAMLSchemaService extends JSONSchemaService {
...
+            const yamlLanguageServerModeline = doc.lineComments.find(lineComment => lineComment.startsWith(modelineDeclaration));
+            if (yamlLanguageServerModeline != undefined) {
+                const schemaKey = '$schema=';
+                const indexOfJsonSchemaParameter = yamlLanguageServerModeline.indexOf(schemaKey);
+                if (yamlLanguageServerModeline.indexOf(schemaKey) != -1) {
```
gorkem, line 308: You can eliminate the second call to index if with `if (indexOfJsonSchemaParameter != -1)`

src/languageservice/services/yamlSchemaService.ts
apupier: I went to the easiest. it is also easier to read code. (for instance, I don't see the regexp that you provided to fit what is expected but i'm not sure if it is just an example or not)
That's possible to use regular expression. It allows to be more flexible for the users.
Not sure it is really needed.
I propose to report an issue to handle it in a separate PR. because there is the case of the double-space between `#` and `yaml-language-server` but there is also tabs. and if it is several spaces between the modeline attributes. and surely a bunch of other cases.

apupier:
> Is there a reason why we can not use a pattern such as `# $schema=<urlToTheSchema>`

it is to better match [_modeline_](https://vim.fandom.com/wiki/Modeline_magic) like way of writing it. It allows to separate concerns, to clarify in which context it is expected to be used and also it will be easier to add other possible parameters.

That said, technically it is absolutely possible to use directly the notation `# $schema=<urlToTheSchema>`.

From a functional point of view, `# $schema=<urlToTheSchema>` is shorter.

src/languageservice/services/yamlSchemaService.ts
apupier: good catch (still not acustomed to this language where there is no good refactoring tooling compared to Java)

apupier:
> can you set the modelline for each specific yaml doc or does setting it once set it globally for the entire file?

you can set one per yaml doc.

and it is joining some limitations I thought during the night:
- it doesn't work if there is only the comment for completion because there is no document created
- I think that if the comment is in a directive section, it should apply to all documents. unless there is another one defined in document itself.

test/autoCompletion.withschemainfile.test.ts
apupier: this is an existing bug on master https://github.com/redhat-developer/yaml-language-server/issues/288

I think we can go without checking for the specific completion in this PR.

src/languageservice/services/yamlSchemaService.ts
gorkem: `string.match(/\S+/g)` would be the way to get non-whitespace  and it is a loop from there. 

We should build this flexibility because at the current implementation if somebody tries to use the feature and misses a single space we fail silently. Or we can build a way to warn the users of such "syntax" errors when we detect the intention but that is essentially the same implementation.

apupier: as always not so trivial with regexp. `\S` is matching also new line and page break Which we don't want here.

that said, given that we are providingline by line, I think we won't have the false-positives thanks to that.

src/languageservice/services/yamlSchemaService.ts
```diff
@@ -282,6 +293,30 @@ export class YAMLSchemaService extends JSONSchemaService {
...
+        if (doc instanceof SingleYAMLDocument) {
+            const yamlLanguageServerModeline = doc.lineComments.find(lineComment => {
+                const matchModeline = lineComment.match(/^#\s+yaml-language-server\s*:/g);
+                return matchModeline !== null && matchModeline.length === 1;
+            });
+            if (yamlLanguageServerModeline != undefined) {
+                const schemaMatchs = yamlLanguageServerModeline.match(/\$schema=\S+/g);
+                if(schemaMatchs !== null && schemaMatchs.length >= 1) {
+                    if (schemaMatchs.length >= 2) {
+                        console.log('Several $schema attributes has been found on the yaml-language-server modeline. The first one will be picked.');
```
apupier, line 312: How can I send a notification to the client?

src/languageservice/services/yamlSchemaService.ts
apupier: updated to cover more cases with extra spaces

src/languageservice/services/yamlSchemaService.ts
gorkem: This can be fine for now. server.ts binds `console.log` and `console.error` to the connection.

fbricon: ...have been found...

apupier: https://github.com/redhat-developer/yaml-language-server/pull/308

shadycuz:
@apupier Does this support local schema files? I'm trying this: `# yaml-language-server: $schema= "./stack_schema.json"` but its not working.

PR #291, opened 2020-08-03
Provide default object as completion snippet

evidolob:
Fix: https://github.com/redhat-developer/vscode-yaml/issues/200


JPinkney:
I've just tried testing this and it looks like I'm getting unexpected results.

Attempting to autocomplete:
```
env: 
```
in VSCode gives
```
env: 	KEY: VALUE
```
which is invalid YAML. I think it needs to do:
```
env:
  KEY: VALUE
```

evidolob:
@JPinkney Yes, you are right

evidolob:
@JPinkney I done:
ezgif com-video-to-gif (28)

PR #292, opened 2020-08-04
Add validation of date and time formats

apgrucza:
These changes are copied from https://github.com/microsoft/vscode-json-languageservice/blob/master/src/parser/jsonParser.ts, upon which this file appears to be based.


apgrucza:
Sure. There was no test coverage for `jsonParser07.ts` so I've taken all the tests from https://github.com/microsoft/vscode-json-languageservice/blob/master/src/test/parser.test.ts and put them here.

PR #293, opened 2020-08-04
Fix document symbols computation if yaml has complex mappings

evidolob:
Fix: https://github.com/redhat-developer/vscode-yaml/issues/285

### DIFF
--- a/test/documentSymbols.test.ts
+++ b/test/documentSymbols.test.ts
@@ -299,7 +299,8 @@ suite('Document Symbols Tests', () => {
         });
 
         it('Document Symbols with complex mapping and aliases', () => {
-            const content = `version: 0.0.1
+            const content = `
+            version: 0.0.1
             structure:
               ? &root root
               :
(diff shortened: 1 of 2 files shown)


test/documentSymbols.test.ts
```diff
@@ -298,6 +298,28 @@ suite('Document Symbols Tests', () => {
...
+              :
+                element: div
+            conditions:
+              ? *root
+              :
+                style:
+                  height: 41
+            `;
+
+            try {
```
JPinkney, line 314: Do we need the try catch here? wont `const symbols = parseHierarchicalSetup(content);` just error the test which is probably more what we want

test/documentSymbols.test.ts
```diff
@@ -298,6 +298,28 @@ suite('Document Symbols Tests', () => {
...
+                const symbols = parseHierarchicalSetup(content);
+                assert.equal(symbols.length, 2);
```
JPinkney, line 316: Can you check the actual symbols themselves so that we know we are getting the symbols back with the correct start and end etc, similiar to the other tests

Issue #278, opened 2020-07-14
Include eslint checks in build cycle


evidolob:
Now we have 

```
✖ 84 problems (21 errors, 63 warnings)
  21 errors and 0 warnings potentially fixable with the `--fix` option.
```

eslint errors/warnings. 

Need to fail build if PR doesn't not follow codestyle.

JPinkney:
Yeah, we recently switched from tslint to eslint thats why theres so many. I agree, we should add this in. I'm currently re-doing the tests to massively improve them so I can fix up any eslint errors in that folder

evidolob:
@JPinkney Is any reason to use custom rules for eslint? We can consider to use `@typescript-eslint/eslint-recommended` and `@typescript-eslint/recommended` with some modifications. They are pretty good and commonly used for TS projects.

JPinkney:
I'm fine with using @typescript-eslint/recommended if other projects are using it too

evidolob:
Yes, that recommended by [typescript-eslint](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/README.md#configuration).

evidolob:
@JPinkney Another question is two or four space indentation? 
It is more common for JS/TS projects to use two, but most people from Java/C# world prefers four.
A few links:

- https://javascript.info/coding-style
- https://google.github.io/styleguide/jsguide.html#formatting-block-indentation
- https://standardjs.com/rules.html
- https://www.reddit.com/r/javascript/comments/5rjrcy/why_is_twospace_tab_indent_becoming_the_standard/

WDYT?

JPinkney:
I think we should probably go with what the JS/TS community uses, so probably 2

PR #297, opened 2020-08-11
Enable eslint and prettier check, reformat all code

evidolob:
This PR changes rules for eslint to use:
```json
'eslint:recommended',
'plugin:@typescript-eslint/eslint-recommended',
'plugin:@typescript-eslint/recommended',
```
Also it add prettier check with:
```json
  "arrowParens": "always",
  "printWidth": 100,
  "singleQuote": true,
  "trailingComma": "es5",
```

> Now we use two space indentation

Fixes all eslint errors, and reformat code according to prettier settings.
> Note: We still have `✖ 402 problems (0 errors, 402 warnings)` they subject to fix in future
> I include `test` directory in to lint check

Also this PR provide new yarn scripts:

- `yarn clean` - to delete `out/server` folder
- `yarn prettier-check` - to run prettier check
- `yarn prettier-fix` - to reformat all codebase with prettier
- `yarn build` - to clean, lint, prettier-check and compile - new recommended way to build, used by CI

Fix: #278


package.json
```diff
@@ -31,6 +31,7 @@
   "dependencies": {
     "js-yaml": "^3.13.1",
     "jsonc-parser": "^2.2.1",
+    "prettier": "2.0.5",
```
JPinkney, line 34: IIRC this has to be an optional dependency, otherwise projects like Openshift Console cannot use the YAML Language Server because it generates a lot of webpack errors

joshuawilson:
Shouldn't Prettier be under `devDependencies`? I see that you moved it from `dependencies` to `optionalDependencies`. 
That way we have it for build but not required in the release.

@JPinkney can you confirm that will work with OpenShift Console?

evidolob:
@joshuawilson No we use Prettier to format yaml in https://github.com/redhat-developer/yaml-language-server/pull/297/files#diff-671d5a7f3c6a25f68221006e173feb8bR27

joshuawilson:
I thought these changes were related to the code itself.

evidolob:
@apisim 
Prettier is optional dependency, at some environment(when yaml-ls bundled with webpack) it not available.
It seems there are no better way to import module which may not exist, that why we use `require` and try/catch block to handle that. If you know  how to handle that better, without `require`, please provide sample/PR.

.prettierrc.json
```diff
@@ -0,0 +1,14 @@
+{
+  "arrowParens": "always",
+  "printWidth": 100,
```
joshuawilson, line 3: For me, I find this too short. On my laptop alone I can comfortably get to 150. With external monitors it is much higher.
The reason I point this out is that it will unnecessarily change a line and make it harder to read.
I think 120-130 is reasonable. 
But this is just my opinion and I will not hold up the PR for this.

.eslintrc.js
```diff
@@ -1,82 +1,26 @@
...
-            'error',
-            {
-                'anonymous': 'always'
-            }
-        ]
-    }
+  root: true,
+  parser: '@typescript-eslint/parser',
+  env: {
+    node: true,
```
joshuawilson, line 5: Why did you remove `'es6': true,`?

evidolob: I replace all config, it seems just mist that

apisim:
Hi @evidolob 

Here we [go](https://github.com/apimastery/yaml-language-server/blob/master/src/languageservice/services/yamlFormatter.ts). The code actually has more changes...like to remedy the [tight coupling](https://github.com/redhat-developer/yaml-language-server/blob/master/src/server.ts#L590) server.ts has with prettier and allow for passsing in standard `FormattingOptions`...

That fork was created to make it possible to bundle yaml-language-server as [ESM modules](https://github.com/apimastery/yaml-language-server/tree/master/lib/esm) so it can be used by monaco editor in the browser (packed by webpack, indeed). `require` doesn't play well with that, IIRC.

HTH

evidolob:
I that, it seems that you include prettier on your browser code https://github.com/apimastery/monaco-yaml/blob/master/scripts/bundle-esm.js#L47 and do not use webpack.
But as @JPinkney mentioned on https://github.com/redhat-developer/yaml-language-server/pull/297#discussion_r468531115
If I make prettier not optional, and use `import` instead of `require` I may brake downstream projects. 

@JPinkney Can you share more information, why we have `require` in  https://github.com/redhat-developer/yaml-language-server/pull/297/files#diff-671d5a7f3c6a25f68221006e173feb8bR27 originally?

JPinkney:
I can't recall, I just remember https://github.com/redhat-developer/yaml-language-server/pull/167/files was done to make it work for Openshift Console and I only had ~1 week to make those changes

apisim:
> I that, it seems that you include prettier on your browser code https://github.com/apimastery/monaco-yaml/blob/master/scripts/bundle-esm.js#L47 and do not use webpack.

The final application (just prototype as of now) does use webpack. webpack bundles the application (not in a public repo) with this [monaco-yaml](https://github.com/apimastery/monaco-yaml) fork, which currently makes (an unconventional) use of the [yaml-language-server](https://github.com/apimastery/monaco-yaml/blob/master/package.json#L38) fork for the [ESM modules](https://github.com/apimastery/yaml-language-server/tree/master/lib/esm).

I understand the concern with making `prettier` a required dependency. Due to changes in priorities, after the protoype was successful I never got to submitting a PR for bundling yaml-language-server as ESM (and UMD) modules. I'm planning to do that in the next few days. At that time I'll see what can be done about `prettier`.

Cheers!

joshuawilson:
@apisim Do you have a suggestion or request? Or are you just making plans for your PR? What needs to be done to get your Approval?

apisim:
> @apisim Do you have a suggestion or request? Or are you just making plans for your PR? What needs to be done to get your Approval?

I'm OK with the changes. Thanks for checking, @joshuawilson! 

The formatter [already uses](https://github.com/redhat-developer/yaml-language-server/blob/master/src/languageservice/services/yamlFormatter.ts#L28) `require` so it isn't something introduce by this PR.

PR #299, opened 2020-08-12
Add GH Actions workflow to build PR

evidolob:
It is simple Action to build each PR across all OS.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -43,8 +43,10 @@ jobs:
       - name: Run Test
         uses: GabrielBB/xvfb-action@v1.0
         with:
-          run: yarn test --silent
+          run: yarn coverage --silent
       
         # Run Coveralls
-      - name: Run Coveralls
-        run : yarn coveralls
+      - name: Coveralls
+        uses: coverallsapp/github-action@master
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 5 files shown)


joshuawilson:
What is Travis used for it you switch to github actions?

.github/workflows/CI.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, windows-latest, ubuntu-18.04]
```
joshuawilson, line 21: Why is the version set for ubuntu, instead of `ubuntu-latest`?

evidolob:
Now it is using for publishing npm package, as it is dome from @gorkem account.
I think for publishing we need to have separate workflow

.prettierrc.json
```diff
@@ -1,6 +1,8 @@
 {
+  "$schema": "http://json.schemastore.org/prettierrc",
   "arrowParens": "always",
   "printWidth": 130,
   "singleQuote": true,
-  "trailingComma": "es5"
+  "trailingComma": "es5",
+  "endOfLine": "lf"
```
joshuawilson, line 7: `"endOfLine": "auto"` works for me across the different OS'es

PR #305, opened 2020-08-17
Build as ESM and UMD libs

apisim:
The build now _also_ produces libraries in ESM and UMD format. That makes it possible, for example, to use yaml-language-server in [Monaco Editor](https://microsoft.github.io/monaco-editor/) with the YAML language server running as a worker in the browser.

Summary of the Changes
==

**New Files**

* tsconfig.esm.json - configuration to compile the code as an ES module.
* tsconfig.umd.json - configuration to compile the code in UMD format.

The output for both is a new directory `/lib`.

**Modified Files**

* package.json - _added_ to the existing build the compiling of the code to ESM and UMD libraries/modules; bumped `@types/prettier` to the latest version.
*  .gitignore - added `lib/` so Git will ignore the untracked files in it.
* src/languageservice/services/schemaRequestHandler.ts - replaced `require` with `import`.
* src/languageservice/services/yamlFormatter.ts - replaced `require` with `import`. This makes it possible to compile with webpack and have the formatting work when running in the browser. There is another change described below.
* yarn.lock - created by a "fresh" `yarn install`.

**yamlFormatter.ts**
`yamlFormatter.ts` now supports configuring indentation using the standard `FormattingOptions` from 'vscode-languageserver-types'. The existing code in [server.ts](https://github.com/redhat-developer/yaml-language-server/blob/master/src/server.ts#L636) can be seen as tightly coupled with `prettier` - `tabWidth` is [specific to `prettier`](https://prettier.io/docs/en/options.html#tab-width).
Moreover, to configure the tab size when creating a Monaco Editor instance one has to use the standard [`tabSize`](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.istandaloneeditorconstructionoptions.html#tabsize); using `tabWidth` doesn't work.
Notice that this change shouldn't break any existing code that makes use of `tabWidth` (e.g. when running the YAML language server in something like VS Code).

Others
==
Notice that according to the [npm-publish docs](https://docs.npmjs.com/cli/publish), "...All files in the package directory are included if no local .gitignore or .npmignore file exists. If both files exist and a file is ignored by .gitignore but not by .npmignore then it will be included". Therefore, my assumption is that the `lib/` directory and its content will be published along with `out/server`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ out/server
 .DS_Store
 yarn-error.log
 coverage
+lib/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 - Make yaml-language-server available as ESM and UMD modules in the `/lib` directory [#305](https://github.com/redhat-developer/yaml-language-server/pull/305)
 
+#### 0.10.1
+
+- Fix for cannot read property 'lineComments' of undefined Code: -32603 [redhat-developer/vscode-yaml#312](https://github.com/redhat-developer/vscode-yaml/issues/358)
+
 #### 0.10.0
 
 - Allows to declare a schema inside the yaml file through modeline `# yaml-language-server: $schema=<urlOfTheSchema>`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+#### 0.11.0
+
+- Make yaml-language-server available as ESM and UMD modules in the `/lib` directory [#305](https://github.com/redhat-developer/yaml-language-server/pull/305)
+
 #### 0.10.0
 
 - Allows to declare a schema inside the yaml file through modeline `# yaml-language-server: $schema=<urlOfTheSchema>`
--- a/README.md
+++ b/README.md
@@ -199,3 +199,7 @@ This repository only contains the server implementation. Here are some known cli
 ### Connecting to the language server via stdio
 
 We have included the option to connect to the language server via [stdio](https://github.com/redhat-developer/yaml-language-server/blob/681985b5a059c2cb55c8171235b07e1651b6c546/src/server.ts#L46-L51) to help with intergrating the language server into different clients.
+
+### ESM and UMD Modules
+
+Building the YAML Language Server produces [CommonJS](http://www.commonjs.org/) modules in the `/out/server/src` directory. In addition, a build also produces [UMD](https://github.com/umdjs/umd) (Universal Module Definition) modules and [ES Modules](https://tc39.es/ecma262/#sec-modules) (ESM) in the `/lib` directory. That gives you choices in using the YAML Language Server with different module loaders on the server side and in the browser with bundlers like webpack.
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "typescript": "^3.8.3"
   },
   "scripts": {
-    "clean": "rimraf out/server",
+    "clean": "rimraf out/server && rimraf lib",
     "compile": "tsc -p .",
     "watch": "tsc --watch -p .",
     "test": "mocha --require ts-node/register --ui tdd ./test/*.test.ts",
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
   "devDependencies": {
     "@types/mocha": "2.2.48",
     "@types/node": "^12.11.7",
-    "@types/prettier": "^1.18.0",
+    "@types/prettier": "2.0.2",
     "@typescript-eslint/eslint-plugin": "^3.2.0",
     "@typescript-eslint/parser": "^3.2.0",
     "coveralls": "^3.0.5",
@@ -69,7 +69,10 @@
     "coveralls": "nyc --reporter=lcov --reporter=text mocha --require ts-node/register --require source-map-support/register --recursive --ui tdd ./test/*.test.ts",
     "lint": "eslint -c .eslintrc.js --ext .ts src test",
     "prettier-fix": "yarn prettier --write .",
-    "build": "yarn clean && yarn lint && yarn compile"
+    "build": "yarn clean && yarn lint && yarn compile && yarn build:libs",
+    "build:libs": "yarn compile:umd && yarn compile:esm",
+    "compile:umd": "tsc -p ./tsconfig.umd.json",
+    "compile:esm": "tsc -p ./tsconfig.esm.json"
   },
   "nyc": {
     "extension": [
@@ -80,6 +83,7 @@
       "**/*.d.ts",
       "test/",
       "out",
+      "lib",
       "coverage/",
       ".eslintrc.js"
     ],
--- a/src/languageservice/services/schemaRequestHandler.ts
+++ b/src/languageservice/services/schemaRequestHandler.ts
@@ -1,7 +1,7 @@
 import { URI } from 'vscode-uri';
 import { IConnection } from 'vscode-languageserver';
 import { xhr, XHRResponse, getErrorStatusDescription } from 'request-light';
-import fs = require('fs');
+import * as fs from 'fs';
 
 import { VSCodeContentRequest, CustomSchemaContentRequest } from '../../requestTypes';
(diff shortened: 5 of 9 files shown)


joshuawilson:
@apisim Thanks for the PR. I think that updating all the dependencies may have caused the build error. I would like to see a massive version update like that done in a separate PR or at least commit as it has a high chance of breaking things.

apisim:
@joshuawilson As far as I can tell, the Travis CI build was successful.  From the two that failed, one is a 503 Gateway error, which has nothing to do with the code,  and the one for the windows build has no info in the raw log. Did something just time out? My builds on Windows run fine and the tests all run successfully.

I don't see a way to manually re-run the GitHub workflow actions.  From what I've been reading,  some [workflow_dispatch configuration](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) is needed?

I guess I'll have to push some trivial changes to trigger the workflow (when I get back to my laptop). 

Thanks!

joshuawilson:
I was able to manually re-trigger them. We just got github actions setup, so I didn't know how to do it either.

apisim:
Thanks, @joshuawilson! 
I knew it wasn't problems with the code because the builds from the GitHub actions were all successful in my playground.

evidolob:
> @apisim Thanks for the PR. I think that updating all the dependencies may have caused the build error. I would like to see a massive version update like that done in a separate PR or at least commit as it has a high chance of breaking things.

+1 to @joshuawilson 
It is better to make so huge dependency update in separate PR.

apisim:
> @apisim Could you update README.md?
> For `tsconfig*` you can use [extends](https://www.typescriptlang.org/tsconfig#extends), to put all common settings in one `tsconfig.base.json` and put different settings in proper `tsconfig.esm.json` and `tsconfig.umd.json`. And, I think, we do not need of old `tsconfig.json`

@evidolob Thank you for the review and the comments.
The build fails if I remove `tsconfig.json`. Here's a partial output:
```
$ yarn build
...
$ tsc -p .
error TS5057: Cannot find a tsconfig.json file at the specified directory: '.'.
```

Sure - I can extract common tsconfig settings into a `tsconfig.base.json` and have `tsconfig.esm.json` and `tsconfig.umd.json` inherit from it. Should I also make `tsconfig.json` extend it? `tsconfig.json` doesn't seem to have that many common configuration options with `tsconfig.esm.json` and `tsconfig.umd.json`...but let me know what you think.

Regarding updating README.md - what information in particular should be included or changed?

apisim:
> > @apisim Thanks for the PR. I think that updating all the dependencies may have caused the build error. I would like to see a massive version update like that done in a separate PR or at least commit as it has a high chance of breaking things.
> 
> +1 to @joshuawilson
> It is better to make so huge dependency update in separate PR.

OK - pushed a commit to back out the updates to the dependencies in `yarn.lock`. The only update I left is for `@types/prettier` - from "1.18.0" to the latest "2.0.2".
@joshuawilson @evidolob

evidolob:
> @evidolob Thank you for the review and the comments.
> The build fails if I remove `tsconfig.json`. Here's a partial output:
> 
> ```
> $ yarn build
> ...
> $ tsc -p .
> error TS5057: Cannot find a tsconfig.json file at the specified directory: '.'.
> ```
> 
> Sure - I can extract common tsconfig settings into a `tsconfig.base.json` and have `tsconfig.esm.json` and `tsconfig.umd.json` inherit from it. Should I also make `tsconfig.json` extend it? `tsconfig.json` doesn't seem to have that many common configuration options with `tsconfig.esm.json` and `tsconfig.umd.json`...but let me know what you think.

I think we may have "default" `package.json`, which builds umd js files, and have additional `tsconfig.esm.json` which extends default and just brings changes which it need.

> Regarding updating README.md - what information in particular should be included or changed?

At least we need to update [getting-started](https://github.com/redhat-developer/yaml-language-server#getting-started) section. And it would be nice to have some section which describes that now we have to type of modules, and how and when use them.

apisim:
@evidolob Are you thinking of not producing and publishing `/out/server/**` anymore but use `/lib/umd/**` (or `/lib/esm/**`) instead?

evidolob:
I was thinking that it was your intention.
That why I ask to update README. In case if you want just to add more targets, we can keep everything as is.
Just extend `tsconfig.json` from `tsconfig.esm.json` and `tsconfig.umd.json`. And add to readme section about libraries.

JPinkney:
Btw a lot of clients currently use `out/server/**` so if we got rid of that it would be a breaking change

apisim:
I'm sorry for the confusion - no, my intention wasn't to replace `out/server/**`. I "...added to the existing build the compiling of the code to ESM and UMD libraries/modules".

Could you clarify, @evidolob, "...Just extend `tsconfig.json` from `tsconfig.esm.json` and `tsconfig.umd.json`"? Which will be the base, the "parent" and which will be the child[ren]?


I guess this "missed the train" for [Release 0.10.0](https://github.com/redhat-developer/yaml-language-server/pull/306)?

evidolob:
`tsconfig.json` will be the 'base'(parent) ideally without modification, and  `tsconfig.esm.json` and `tsconfig.umd.json` child.

As for release, I think you have time, till tomorrow, I do not have time to do it today.

joshuawilson:
With these changes to the `tsconfig` name, we may need to set it in `.eslintrc.js` in the `parserOptions` section with something like `"project": "./tsconfig.json"` in there. I don't remember if we need it for each different tsconfig. I just want to make sure that the linter still has access to the tsconfig settings.

apisim:
I'm not sure I understand, @joshuawilson - the original `tsconfig.json` didn't change and has the same (default) name.

On another note, @evidolob - it is a bit more complicated to update README.md with info on how to use the new ESM and UMD libraries. I'm currently using them in a [monaco-yaml fork](https://github.com/apimastery/monaco-yaml) with the goal of merging all the changes in the fork back into the [original monaco-yaml](https://github.com/pengx17/monaco-yaml) repo. 
By using these libraries I am able to have `redhat-developer/yaml-language-server` running with [Monaco Editor](https://github.com/microsoft/monaco-editor) as a worker in the browser - no remote calls to a server are required like, for example, in case of [Theia](https://github.com/eclipse-theia/theia) (if I properly understood Theia's architecture the last time I looked at it).

Given the transient nature of the monaco-yaml fork and the unknowns of when the changes will be merged into the original monaco-yaml repo (supposedly after the ESM and UMD libraries are published with yaml-language-server), I can't currently offer "how to use" instructions.

I don't want to be holding up Release 0.10.0. I think it might be best to close this PR. I will still probably submit a separate PR for the changes to `schemaRequestHandler.ts` and `yamlFormatter.ts` for future consideration.

Sincerely - thank you very much for considering the PR and for your time. Thanks!!

evidolob:
@apisim Sorry, maybe I was not clear on update REDME request, I just want to inform users that we produce several different JS modules, and they may use this project not only like node modules.

As for release, we quite flexible, we may release `0.10.1` with this or any other your PR.

We, with @joshuawilson , wants to make release schedule more transparent, usually we make releases every 3 week, I will create some document which describe release process.

If you need to have some bug fix or feature released, don't hesitate to contact us, we will glad to help you.

apisim:
If that's OK with you, I'd like to move forward with this pull request @joshuawilson @evidolob @JPinkney 

I believe the comments so far were all addressed. Let me know, please, if there's some tweaking you think is needed or of anything else.

Thanks!

JPinkney (approved):
Looks good from my side, I think both @evidolob and @joshuawilson are on PTO so I'm going to wait for them to get back before I merge

evidolob (approved):
LGTM
A few last things:
- I think it would be better if we delete `lib` folder with `clean` script
- It would be great if you add record about new libraries in to `CHANGELOG.md`

apisim:
OK. I added a step to the `clean` script to clean up the `/lib` directory before a build; tested it when the `/lib` directory exists already or not. Also, I added a one-liner to CHANGELOG.md assuming the release would be `0.11.0`.

Let me know if there's anything else.

PR #309, opened 2020-08-20
Fix for redhat-developer/vscode-yaml#353, error when editing yaml wit…

JPinkney:
…h comments

This PR should fix https://github.com/redhat-developer/vscode-yaml/issues/353.

This is occurring because https://github.com/redhat-developer/yaml-language-server/blob/master/src/languageservice/parser/yamlParser07.ts#L125 is incrementing yamlDocCount by 1 when `---` is the first thing present in the YAML file. Then when https://github.com/redhat-developer/yaml-language-server/blob/master/src/languageservice/parser/yamlParser07.ts#L129 is hit yamlDocs[1] will be undefined when there is only a single YAML document, thus creating the error.

The solution is to only increment yamlDocCount when you see `---` and it's not the first line of the file. That way these two instances (with and without `---` at the start get parsed correctly)
```
# some comment
test: test
...
```
and
```
---
# some comment
test: test
...
```

### DIFF
--- a/src/languageservice/parser/yamlParser07.ts
+++ b/src/languageservice/parser/yamlParser07.ts
@@ -17,7 +17,6 @@ import { getLineStartPositions } from '../utils/documentPositionCalculator';
 import { ASTNode } from '../jsonASTTypes';
 import { ErrorCode } from 'vscode-json-languageservice';
 
-const YAML_DIRECTIVE_PREFIX = '%';
 const YAML_COMMENT_PREFIX = '#';
 const YAML_DATA_INSTANCE_SEPARATOR = '---';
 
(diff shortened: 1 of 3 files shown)


src/languageservice/parser/yamlParser07.ts
```diff
@@ -118,11 +118,11 @@ export function parse(text: string, customTags = []): YAMLDocument {
 function parseLineComments(text: string, yamlDocs: SingleYAMLDocument[]): void {
   const lines = text.split(/[\r\n]+/g);
   let yamlDocCount = 0;
-  lines.forEach((line) => {
+  lines.forEach((line, index) => {
     if (line.startsWith(YAML_DIRECTIVE_PREFIX) && yamlDocCount === 0) {
       yamlDocCount--;
     }
-    if (line === YAML_DATA_INSTANCE_SEPARATOR) {
+    if (line === YAML_DATA_INSTANCE_SEPARATOR && index !== 0) {
```
gorkem, line 125: index > 0 is probably safer

JPinkney:
Moving this out of WIP, @apupier can you take a look again

charlesguo404:
LGTM

src/languageservice/parser/yamlParser07.ts
```diff
@@ -118,14 +118,14 @@ export function parse(text: string, customTags = []): YAMLDocument {
 function parseLineComments(text: string, yamlDocs: SingleYAMLDocument[]): void {
   const lines = text.split(/[\r\n]+/g);
   let yamlDocCount = 0;
+  let firstSeperatorFound = false;
```
apupier, line 121: typo seperator -> separator

apupier (changes requested):
minor stuff that would be nice to change before merging

- constant 'YAML_DIRECTIVE_PREFIX' is no more used and thus an be removed

apupier (approved):

PR #313, opened 2020-09-09
Fix scheme load on windows with full path

evidolob:
This PR fix scheme loading on Windows, in case if URI has window path, ie `c:\some\path\to\scheme.json`.

Fix: https://github.com/redhat-developer/vscode-yaml/issues/347

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 #### 0.11.0
 
 - Make yaml-language-server available as ESM and UMD modules in the `/lib` directory [#305](https://github.com/redhat-developer/yaml-language-server/pull/305)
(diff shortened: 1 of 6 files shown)


JPinkney (approved):
I think we should probably update the readme on the yaml language server and vscode-yaml saying that this now works

evidolob:
Hm, we never mention that this not work, I found that bug during launching tests on windows, I should add changelog record for this, but I do not think that this worth mentioning in readme.

JPinkney:
In the past, people have asked whether this works, whether absolute paths vs relative paths work, etc. From my experience, some people have to have examples added in the readme thats identical to how they'd use it because they aren't going to try it out themselves until the readme says it

JPinkney (approved):

Issue #283, opened 2020-07-17
Provide completion text to insert without tabs, with only space


apupier:
tabs are not supported in yaml files.
Some IDEs are converting automatically such as VS Code Yaml but not all, like Eclipse client https://github.com/eclipse/wildwebdeveloper/issues/478

apupier:
see https://github.com/redhat-developer/yaml-language-server/pull/272#issuecomment-655581898 for previous discussion on the subject

evidolob:
We may use [workspace_configuration](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_configuration) `workspace/configuration` request to get preference value for `editor.tabSize`. At least it will work for VSCode. For other IDE it will depends on their implementation of `workspace_configuration`.
Or as alternative we may calculate `space` or `tab` for each line or yaml file.
And then use that value for snippet/completion generation.

mickaelistria:
Since the YAML specification explicitly banishes tabs for indent ( https://yaml.org/spec/1.2/spec.html#id2777534 ), it can be seen as a that the LS can produce incorrect yaml by inserting tabs in any case.
The IDE shouldn't have to post-process the LS responses to make the file correct. They may post-process to make it more consistent with X or Y that's specific to the tool, but not to make the file conform to its spec/grammar.
So IMO, it's important to change that in the LS

My suggestion:
If no workspace_configuration is set, the LS should look at the file to infer the number of spaces for indent (looking at the number of spaces before the 1st char of the 2nd line can be one sufficient heuristic). In case the current file is empty, the LS should look in other known yaml files in the folder, in case nothing is available, it should pick the favourite convention of its developers.

PR #315, opened 2020-09-11
Use space for completion text instead of tab

evidolob:
This PR replace hardcoded `\t` character in completion snippets.
It try to take space preference, if it not defined, it try to detect indentation by file content, if it is not possible it use double space for completion text indentation.

Fix: #283

### DIFF
--- a/src/languageservice/utils/indentationGuesser.ts
+++ b/src/languageservice/utils/indentationGuesser.ts
@@ -3,6 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+//copied from https://github.com/Microsoft/vscode/blob/015c3afe96966df50c15a7d66be2ab0ef1dc5f49/src/vs/editor/common/model/indentationGuesser.ts
+
 import { CharCode } from './charCode';
 import { TextBuffer } from './textBuffer';
 
--- a/src/languageservice/utils/indentationGuesser.ts
+++ b/src/languageservice/utils/indentationGuesser.ts
@@ -1,0 +1,224 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { CharCode } from './charCode';
+import { TextBuffer } from './textBuffer';
+
+class SpacesDiffResult {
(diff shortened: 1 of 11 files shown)


src/languageservice/utils/indentationGuesser.ts
```diff
@@ -0,0 +1,224 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
```
apupier, line 2: Microsoft? copy-pasted from elsewhere? Maybe worth adding a link to where it comes from to ease maintenance

src/languageservice/utils/textBuffer.ts
```diff
@@ -0,0 +1,45 @@
+import { SSL_OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION } from 'constants';
```
apupier, line 1: unused import

src/languageservice/utils/textBuffer.ts
```diff
@@ -0,0 +1,45 @@
...
+export class TextBuffer {
+  constructor(private doc: TextDocument) {}
+
+  getLineCount(): number {
+    return this.doc.lineCount;
+  }
+
+  getLineLength(lineNumber: number): number {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const lineOffsets: number[] = (this.doc as any).getLineOffsets();
```
apupier, line 19: out of curiosity because I'm not familiar with TypeScript. Why do you need to disable eslint rule? Is there another way to write it to avoid it?

evidolob: I cast to `any` to get access to 'internal' method.
Hm, it can be done without `any`

JPinkney (commented):
Not sure if I'm doing something wrong but when I use https://json.schemastore.org/composer as the schema, set the editor tabSize to 8 and then autocomplete authors in the root I get:
```
# yaml-language-server: $schema=https://json.schemastore.org/composer
authors:
  - name: 
```
whereas I _think_ it should get:
```
# yaml-language-server: $schema=https://json.schemastore.org/composer
authors:
        - name: 
```
if i'm not mistaken?

evidolob:
@JPinkney You are right, let me check

evidolob:
@JPinkney I fix the problem. And I completely forgot to create proper PR for vscode-yaml, to test this PR you need to use changes from https://github.com/redhat-developer/vscode-yaml/pull/362

JPinkney (approved):

PR #316, opened 2020-09-12
Set true for documentFormattingProvider

mattn:
I'm not sure why yaml-language-server return documentFormattingProvider: false, but I make sure formatting is working with my small modification.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -411,7 +411,7 @@ connection.onInitialize(
         completionProvider: { resolveProvider: true },
         hoverProvider: true,
         documentSymbolProvider: true,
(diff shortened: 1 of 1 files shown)


JPinkney:
From what I remember someone from the VSCode team told us that it has to be registered dynamically so this is done instead: https://github.com/redhat-developer/yaml-language-server/commit/ca40f01ffd5d8082af1ff54b51317d65307b58f4

mattn:
It should use fixed value false if client support dynamicRegistration.

ssbarnea:
IMHO, safe to close.

gorkem:
Closing.

PR #318, opened 2020-09-16
Add support of Yaml Schema

evidolob:
This PR based on #191, with resolving merge conflicts and tests.

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -22,6 +22,7 @@ import * as nls from 'vscode-nls';
 import { convertSimple2RegExpPattern } from '../utils/strings';
 import { SingleYAMLDocument } from '../parser/yamlParser07';
 import { JSONDocument } from '../parser/jsonParser07';
+import * as yaml from 'js-yaml';
 
 const localize = nls.loadMessageBundle();
 
@@ -83,6 +84,8 @@ export class YAMLSchemaService extends JSONSchemaService {
   private customSchemaProvider: CustomSchemaProvider | undefined;
   private filePatternAssociations: JSONSchemaService.FilePatternAssociation[];
   private contextService: WorkspaceContextService;
+  private customSchemaProvider: CustomSchemaProvider | undefined;
+  private requestService: SchemaRequestService;
 
   constructor(
     requestService: SchemaRequestService,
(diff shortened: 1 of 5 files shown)


test/yamlScheme.test.ts
```diff
@@ -0,0 +1,64 @@
...
+      items: { type: 'string' },
+      type: 'array',
+    });
+  });
+
+  test('Error while loading yaml', async () => {
+    requestServiceStub.rejects();
+    const service = new SchemaService.YAMLSchemaService(requestServiceStub, workspaceContext);
+    const result = await service.loadSchema('fooScheme.yaml');
+    expect(result.errors).length(1);
```
JPinkney, line 53: Should we check what error this is? Or is it just a throwaway error?

JPinkney (approved):

src/languageservice/services/yamlSchemaService.ts
```diff
@@ -488,13 +492,59 @@ export class YAMLSchemaService extends JSONSchemaService {
...
+              const errorMessage = localize(
+                'json.schema.nocontent',
+                "Unable to load schema from '{0}': No content.",
+                toDisplayString(schemaUri)
+              );
+              return new UnresolvedSchema(<JSONSchema>{}, [errorMessage]);
+            }
+
+            try {
+              const schemaContent: JSONSchema = yaml.safeLoad(content);
```
joshuawilson, line 521: If you are doing a simple conversion then I'm concerned that it might not work with k8s yaml schemas. The structure of which has the schema held within it.

evidolob: @joshuawilson I'm not sure that understand your point, can you add sample of such k8s yaml scheme?

joshuawilson: Here is a link to the Knative Eventing Subscription schema. Notice that is has structure wrapping the actual schema. 
https://github.com/knative/eventing/blob/master/config/core/resources/subscription.yaml

evidolob: No, because that file itself is not YAML Scheme, it is k8's CRD. To get actual scheme you need to extract that [yaml](https://github.com/knative/eventing/blob/master/config/core/resources/subscription.yaml#L31-L277) in to separate yaml file

joshuawilson: This is a fine, I was just hoping to be able to point to the URL for a k8s schema and use it. But I do understand that has become a more generic tool.

joshuawilson (approved):

PR #322, opened 2020-09-24
Fix 'tabWidth' option setup for formatter

evidolob:
Fix: https://github.com/redhat-developer/vscode-yaml/issues/366

### DIFF
--- a/src/languageservice/services/yamlFormatter.ts
+++ b/src/languageservice/services/yamlFormatter.ts
@@ -8,7 +8,6 @@
 import { TextDocument, Range, Position, TextEdit, FormattingOptions } from 'vscode-languageserver-types';
 import { CustomFormatterOptions, LanguageSettings } from '../yamlLanguageService';
 import * as prettier from 'prettier';
-import { Options } from 'prettier';
 import * as parser from 'prettier/parser-yaml';
 
 export class YAMLFormatter {
@@ -28,7 +27,7 @@ export class YAMLFormatter {
     try {
       const text = document.getText();
 
-      const prettierOptions: Options = {
+      const prettierOptions: prettier.Options = {
         parser: 'yaml',
         plugins: [parser],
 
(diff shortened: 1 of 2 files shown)


JPinkney (approved):

joshuawilson:
Is there a way we can test this?

evidolob:
@joshuawilson I thinking about that, it may take time, as I need to mock `IConnection` somehow.

joshuawilson:
I don't know how to verify that your change will fix the issue. That is why I'm asking for the test. That will prove it and make sure it doesn't break again.

JPinkney:
We can also extract everything out of server.ts so that the server.ts file basically has the hookups to the language features and nothing else. That way all of the other features would technically be abstracted enough that you can test them directly

apisim:
Yeah, probably #305 is the cause for the problem, I'm sorry.

For what it's worth...I think the changes proposed by @evidolob will fix the issue. Not only that but they will remove the tight coupling `server.ts` has with `prettier` - `tabWidth` is specific to `prettier` whereas `tabSize` is the "standard" [FormattingOptions](https://github.com/codedownio/vscode-languageserver-types/blob/e9bdff0a2370afcfcd453330ed09fe7810d06e5d/src/main.ts#L2186) option.

Here is another, I think, [possible solution](https://github.com/redhat-developer/yaml-language-server/compare/master...apimastery:pr-322) that aims to preserve the old behavior (and it keeps the tight coupling). Notice the unit tests. In hindsight, I should have probably coded it that way to begin with:
```
tabWidth: (options.tabWidth as number) || options.tabSize,
```
In my local testing, the formatting still works with this in monaco-yaml in the browser.

evidolob:
@JPinkney Yes, it seems that refactoring `server.ts` is the only way to write tests for it.

@apisim I like your solution more, can you create PR instead of this?

apisim:
Sure, @evidolob. Please see #324.

evidolob:
Closing, as #324 is better.

Issue #310, opened 2020-08-21
Add schema info to diagnostics


gorkem:
when validation against a schema fails also add the schema information to the published diagnostics. This way users can identify the schemas that is affecting their YAML files.

evidolob:
@gorkem @JPinkney 
I have prototype of this, I add scheme URL in `source` field of `Diagnostic`
Example URL:
<img width="1085" alt="Screenshot 2020-09-24 at 14 52 36" src="https://user-images.githubusercontent.com/929743/94141844-0ca5a600-fe76-11ea-8552-31403b20b69b.png">
Example local file:
<img width="1085" alt="Screenshot 2020-09-24 at 14 53 02" src="https://user-images.githubusercontent.com/929743/94141922-2646ed80-fe76-11ea-86be-7de7a44daa21.png">

WDYT?

JPinkney:
That looks good to me

gorkem:
Are you sending it as part of the message filed on `Diagnostic`?

evidolob:
I not sure that understand the question, I add URL/path to [Diagnostic#source](https://code.visualstudio.com/api/references/vscode-api#Diagnostic) field

gorkem:
That is exactly what I was asking. 
Can you also provide a screenshot the problem view when you have one of these issues.

evidolob:
<img width="1140" alt="Screenshot 2020-09-28 at 10 28 25" src="https://user-images.githubusercontent.com/929743/94402928-5abbf180-0175-11eb-94e0-e9606aafc1ce.png">

gorkem:
OK this is half bad :) Can we have a format such as `yaml-schema: ${schema}` and for generic yaml issues just `yaml`

evidolob:
We can, but what is the value of `${schema}` ? URI or just name?
We can use schema `title` if present, if not, than URI.

gorkem:
> We can use schema `title` if present, if not, than URI.
+1 Sounds like a plan

rivy:
Thanks, nice addition!

ssbarnea:
I looked for the entire morning on how to expose this information and I still do not know how to enable the diagnosticst. Was this fixed only in an unreleased version? -- I am currently using the latest YAML 0.15.0 extension, which was released two weeks ago, so I expect to be able to see the diagnostics.  Is there a hidden option to enable these?

evidolob:
@ssbarnea this works since `0.12.0`

[...]

PR #326, opened 2020-09-30
Add schema info to diagnostics

evidolob:
Fix: #310

Also fix: https://github.com/redhat-developer/vscode-yaml/issues/373 partially, as I add `YAML`  for YAML syntax errors

Demo:
Screenshot 2020-09-30 at 14 46 57

Schema with `title`:
Screenshot 2020-09-30 at 14 47 30

Local schema:
Screenshot 2020-09-30 at 14 48 38

Yaml error:
Screenshot 2020-09-30 at 14 49 38

Problems:
Screenshot 2020-09-30 at 14 52 02

Kubernetos schema:
Screenshot 2020-10-13 at 12 20 06

### DIFF
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -1051,7 +1051,7 @@ function validate(
               validationResult.propertiesValueMatches++;
             }
           } else {
-            propertySchema.url = schema.url;
+            propertySchema.url = schema.url ?? originalSchema.url;
             const propertyValidationResult = new ValidationResult(isKubernetes);
             validate(child, propertySchema, schema, propertyValidationResult, matchingSchemas, isKubernetes);
             validationResult.mergePropertyMatch(propertyValidationResult);
@@ -1265,12 +1265,15 @@ function getSchemaSource(schema: JSONSchema, originalSchema: JSONSchema): string
       label = schema.title;
     } else if (originalSchema.title) {
       label = originalSchema.title;
-    } else if (originalSchema.url) {
-      const url = URI.parse(originalSchema.url);
-      if (url.scheme === 'file') {
-        label = url.fsPath;
+    } else {
+      const uriString = schema.url ?? originalSchema.url;
+      if (uriString) {
+        const url = URI.parse(uriString);
+        if (url.scheme === 'file') {
+          label = url.fsPath;
+        }
+        label = url.toString();
       }
-      label = url.toString();
     }
     if (label) {
       return `yaml-schema: ${label}`;
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -52,11 +52,14 @@ const formats = {
   },
 };
 
+export const YAML_SOURCE = 'YAML';
+
 export interface IProblem {
   location: IRange;
   severity: DiagnosticSeverity;
   code?: ErrorCode;
   message: string;
+  source?: string;
 }
 
 export abstract class ASTNodeImpl {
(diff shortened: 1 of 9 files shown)


JPinkney:
I think this is a really cool feature. I've been playing around with it and it seems like it works perfectly when theres one schema associated to the file but it has a bit of trouble when there's multiple schemas associated to a file. 

E.g.
```
"yaml.schemas": {
    "kubernetes": "test.yaml",
    "https://json.schemastore.org/composer": "test.yaml"
}
```
and then with content
```
abandoned: v1
archive:
  exclude:
    asd: asd
```

you'll get `yaml-schema: schemaservice://combinedschema/file%3A///home/joshpinkney/Documents/Work/Schemas/test.yaml` in the error message when you'd expect the composer schema.

Also if you did:
```
"yaml.schemas": {
    "kubernetes": "prometheus.yaml"
},
```
with content
```
apiVersion: v1
kind: Pod
metadata:
  name: False
```
you'll get `yaml-schema:`

evidolob:
@JPinkney Thx, I miss that file may have multiple schemas, going to fix that

evidolob:
@JPinkney 
Now it should resolve schema correctly:
Screenshot 2020-10-13 at 12 20 56

evidolob:
@gorkem @JPinkney @joshuawilson 
Full k8s url:
Screenshot 2020-10-13 at 12 34 24

test/schemaValidation.test.ts
```diff
@@ -815,4 +841,52 @@ suite('Validation Tests', () => {
         .then(done, done);
     });
   });
+
+  describe('Schema with title', () => {
+    it('validator use schema title instead of url', async () => {
```
JPinkney, line 846:
```suggestion
    it('validator uses schema title instead of url', async () => {
```

src/languageservice/parser/jsonParser07.ts
```diff
@@ -1218,3 +1257,28 @@ function validate(
...
+    if (schema.title) {
+      label = schema.title;
+    } else if (originalSchema.title) {
+      label = originalSchema.title;
+    } else {
+      const uriString = schema.url ?? originalSchema.url;
+      if (uriString) {
+        const url = URI.parse(uriString);
+        if (url.scheme === 'file') {
+          label = url.fsPath;
```
JPinkney, line 1273: In this case won't label always just end up being whatever url.toString() is?

JPinkney (commented):
Generally, it looks good to me. I tried it out locally and everything seems to be working almost perfectly. I only issue I ran into was in a test scenario with a `.drone.yml` file where I had yaml.settings as:
```
"yaml.schemas": {
    "kubernetes": ".drone.yml"
},
```
with the content:
```
apiVersion: v1
kind: Deployment
```
where I was getting error message `Property apiVersion is not allowed.YAML` when hovering apiVersion and when hovering deployment I got
```
Value is not accepted. Valid values: "pipeline".: YAML
```

maybe in this case the validator matches a little bit of the drone ci and the kubernetes schema and cannot decide which one it is matching? I'm not sure

src/languageservice/parser/jsonParser07.ts
evidolob: Yes, it is done for local files, I assume the path `/home/user/folder/schema.json` will be more informative than `file:///home/user/folder/schema.json`. If URI has different schema than it used as is.

evidolob:
@JPinkney I fix that, can you check again?

JPinkney (approved):
Tested again and now everything seems to be working as expected!

gorkem (approved):

Issue #317, opened 2020-09-15
Error when file has "Type" attribute --- yamlValidation


caobincb:
Request textDocument/completion failed.
  Message: Request textDocument/completion failed with message: Cannot read property 'type' of undefined
  Code: -32603

evidolob:
@caobincb Can you provide yaml sample on witch you has that error?

GhostLyrics:
Create the following folder file as `roles/example/tasks/main.yml` to simulate using Ansible and open the top level folder with e.g. Visual Studio Code with yaml-language-server enabled:

```
---
- 
```

After typing the list indicator, type any string and you'll get the error message for every key you hit. Example:

```
---
- na
```

Result:
```
[Error - 8:34:51 PM] Request textDocument/completion failed.
  Message: Request textDocument/completion failed with message: Cannot read property 'type' of undefined
  Code: -32603 
```

bendem:
Thanks god for that fix ❤️

PR #328, opened 2020-10-02
Fix error if schema doesn't contain type

evidolob:
Fix: #317


joshuawilson:
@evidolob github actions pointed out that you have added a function to import that isn't being used.

JPinkney (approved):

joshuawilson (approved):

PR #332, opened 2020-10-06
Fix completion indentation for object in array

### LINKED ISSUE #329: Bug: autogenerated snippet for keys that contain an array of objects is badly indented
When I validate auto-completion for a key that contain an array of objects, the snippet generated contains 2 extra spaces for each sub-key generated.

Here's an example in `kubernetes.yml`:

Given that I have this file content:
```
metadata:
    ownerReferences
```
Then I go to end of `ownerReferences`, press `Ctrl+Tab` and `[ENTER]` to validate the suggested completion.  
I get this result:
```
metadata:
    ownerReferences:
        - apiVersion: 
            kind: 
            name: 
            uid: 
```

The problem is that "kind", "name" and "uid" keys have 2 extra spaces in front.

evidolob:
This is firs PR, fix completion snippet generation for object in array. We also have problem with settings synchronisation, in case if `editor.tabSize` set up in language scope `[yaml]` we not take in account it. This PR fix that, but it also requires fix in `vscode-yaml` which I'm going to provide.

For #329

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -149,7 +149,7 @@ export class YAMLCompletion extends JSONCompletion {
           suggestion.label = label;
           proposed[label] = suggestion;
           result.items.push(suggestion);
-        } else if (!existing.documentation) {
+        } else if (!existing.documentation && suggestion.documentation) {
           existing.documentation = suggestion.documentation;
         }
       },
@@ -260,7 +260,7 @@ export class YAMLCompletion extends JSONCompletion {
     addValue: boolean,
     separatorAfter: string,
     collector: CompletionsCollector,
-    document
+    document: TextDocument
   ): void {
     const matchingSchemas = doc.getMatchingSchemas(schema.schema);
     matchingSchemas.forEach((s) => {
@@ -723,12 +723,7 @@ export class YAMLCompletion extends JSONCompletion {
             break;
           case 'array':
             {
-              const arrayInsertResult = this.getInsertTextForArray(
-                propertySchema.items,
-                separatorAfter,
-                `${indent}${this.indentation}`,
-                insertIndex++
-              );
+              const arrayInsertResult = this.getInsertTextForArray(propertySchema.items, separatorAfter, insertIndex++);
               insertIndex = arrayInsertResult.insertIndex;
               insertText += `${indent}${key}:\n${indent}${this.indentation}- ${arrayInsertResult.insertText}\n`;
             }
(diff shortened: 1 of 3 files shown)


src/server.ts
```diff
@@ -511,7 +511,9 @@ connection.onDidChangeConfiguration((change) => {
 
   schemaConfigurationSettings = [];
 
-  if (settings.editor?.tabSize) {
+  if (settings['[yaml]'] && settings['[yaml]']['editor.tabSize']) {
+    indentation = ' '.repeat(settings['[yaml]']['editor.tabSize']);
```
JPinkney, line 515: Should we document this somewhere just so people know that if their editor supports yaml specific tabSizes then it takes precedence over the regular editor.tabSize

JPinkney (approved):
Tested it out and it works well

src/server.ts
evidolob: Hm, not sure, do we have some description of settings that we use ?

JPinkney: We only really have https://github.com/redhat-developer/yaml-language-server#language-server-settings but since the number of spaces is grabbed from editor.tabSize we should probably add a little writeup there. I can do it in a separate PR if you want

evidolob: Yes, that helps, pleas include also `proxy` and `proxyStrictSSL` settings usage

src/languageservice/services/yamlCompletion.ts
```diff
@@ -149,7 +149,7 @@ export class YAMLCompletion extends JSONCompletion {
           suggestion.label = label;
           proposed[label] = suggestion;
           result.items.push(suggestion);
-        } else if (!existing.documentation) {
+        } else if (!existing.documentation && suggestion.documentation) {
```
joshuawilson, line 152: How is this related? How is it tested?

src/languageservice/services/yamlCompletion.ts
```diff
@@ -260,7 +260,7 @@ export class YAMLCompletion extends JSONCompletion {
     addValue: boolean,
     separatorAfter: string,
     collector: CompletionsCollector,
-    document
+    document: TextDocument
```
joshuawilson, line 263: Did you know that `TextDocument` is deprecated?

src/languageservice/services/yamlCompletion.ts
```diff
@@ -797,12 +792,7 @@ export class YAMLCompletion extends JSONCompletion {
         break;
       case 'object':
         {
-          const objectInsertResult = this.getInsertTextForObject(
-            schema,
-            separatorAfter,
-            `${indent}${this.indentation}`,
-            insertIndex++
-          );
+          const objectInsertResult = this.getInsertTextForObject(schema, separatorAfter, `${this.indentation}  `, insertIndex++);
```
joshuawilson, line 795: Do you need to also make this change for line 736 above?

src/languageservice/services/yamlCompletion.ts
evidolob: Yes, but we use it across all project.

src/languageservice/services/yamlCompletion.ts
evidolob: No

joshuawilson (approved):

PR #333, opened 2020-10-07
Publish to npm on each commit in master

evidolob:
This PR add GH Action workflow which publish changes in master branch to npm registry with `next` version tag.


README.md
```diff
@@ -225,3 +225,8 @@ We have included the option to connect to the language server via [stdio](https:
 ### ESM and UMD Modules
 
 Building the YAML Language Server produces [CommonJS](http://www.commonjs.org/) modules in the `/out/server/src` directory. In addition, a build also produces [UMD](https://github.com/umdjs/umd) (Universal Module Definition) modules and [ES Modules](https://tc39.es/ecma262/#sec-modules) (ESM) in the `/lib` directory. That gives you choices in using the YAML Language Server with different module loaders on the server side and in the browser with bundlers like webpack.
+
+### CI
+
+We use GitHub Action to publish each change in the `master` branch to npm registry, such packages tagged with `next` tag.
```
JPinkney, line 231:
```suggestion
We use a GitHub Action to publish each change in the `master` branch to [npm registry](https://www.npmjs.com/package/yaml-language-server) with the `next` tag.
```

JPinkney (approved):

joshuawilson:
If you merge this you will have 2 workflow files that will both fire when merge to master is detected. They do not have a dependency so failing the test will not keep it from releasing. 
You will either need to find a way to chain them. (after a quick search I do not see how to do that)
OR you will have to put everything in one workflow file and use conditionals there. (not sure this will work either)

evidolob:
@joshuawilson  nice catch, yes 
We can run tests in CD workflow or I can move everything in to one workflow

joshuawilson:
You could have one for `on push to master` and another for `on PR`. Both run the tests but only one will release.

JPinkney:
Yeah good catch, I completely forgot about that

joshuawilson:
Chaining would be cleaner as it would not duplicate the test and build steps but I can not find a way to do it.

evidolob:
OK, I think I know how to do that, I will update PR soon, all steps will be in one workflow

.github/workflows/CI.yaml
```diff
@@ -50,3 +51,9 @@ jobs:
         uses: coverallsapp/github-action@master
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
+      
+      - name: Publish 
+        if: ${{ success() && github.event_name == 'push' && github.ref == 'refs/heads/master'}}
```
joshuawilson, line 56: How do you ensure that the other workflow runs first?

evidolob: Which one? We have only one workflow.

joshuawilson (approved):

.github/workflows/CI.yaml
joshuawilson: Sorry, I thought this was an update to the CD workflow. I missed that are only using one now. 
This is a nice solution, good work.

PR #340, opened 2020-10-22
Fix example string of type integer gets pasted as int

evidolob:
### What does this PR do?
Fix issue when strings from example/default value pasted as numbers i.e. `'1'` was pasted like `1`, this PR change that, so now `"1"` will be pasted.

Fix: https://github.com/redhat-developer/vscode-yaml/issues/371

### Is it tested? How?
With schemas provided [there](https://github.com/redhat-developer/vscode-yaml/issues/371#issuecomment-701626267) and [there](https://github.com/redhat-developer/vscode-yaml/issues/371#issuecomment-702086392)


JPinkney (approved):
LGTM

gorkem (approved):

PR #341, opened 2020-10-23
Use next yaml parser

evidolob:
### What does this PR do?
Change yaml parser library to `yaml-language-server-parser`. In master brunch we will use `next` version tag.
Also it removes Travis CI, and add GH Action workflow to publish release.


JPinkney (approved):

.github/workflows/release.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - name: Install dependencies
+      run: npm install
+
+      # Build extension
+    - name: Run build
+      run: npm run build
+
+      # Run tests
+    - name: Run Test
+      run: npm run test
```
joshuawilson, line 43: Should this be `npm` or `yarn`?
And the ones above.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+      # Publish to npm
+    - run: |
+        yarn check-dependencies
+        yarn publish --access public
```
joshuawilson, line 48: What is the difference between what happens here and the on in CI.yaml?
Why not combine them?

yarn.lock
```diff
@@ -2345,10 +2345,10 @@ yallist@^2.1.2:
   resolved "https://registry.yarnpkg.com/yallist/-/yallist-2.1.2.tgz#1c11f9218f076089a47dd512f93c6699a6a81d52"
   integrity sha1-HBH5IY8HYImkfdUS+TxmmaaoHVI=
 
-yaml-ast-parser-custom-tags@0.0.43:
-  version "0.0.43"
-  resolved "https://registry.yarnpkg.com/yaml-ast-parser-custom-tags/-/yaml-ast-parser-custom-tags-0.0.43.tgz#46968145ce4e24cb03c3312057f0f141b93a7d02"
-  integrity sha512-R5063FF/JSAN6qXCmylwjt9PcDH6M0ExEme/nJBzLspc6FJDmHHIqM7xh2WfEmsTJqClF79A9VkXjkAqmZw9SQ==
+yaml-language-server-parser@next:
+  version "0.1.1-2b9337f.0"
```
joshuawilson, line 2349: Is this what you expect, that using `next` will create a version that is higher then the one released?

JPinkney: It's the next version on npm: https://www.npmjs.com/package/yaml-language-server-parser/v/0.1.1-2b9337f.0

.github/workflows/release.yaml
JPinkney: On CI.yaml the action is run on every commit to push to next whereas here it's only ran when a tag is created and pushed to the repo when

joshuawilson: CI.yaml reads to me that if the push is successful, then it will create the tag. Why not just add this to that section? 
Maybe I'm missing it.

evidolob: We do not create tag from CI, we create tag manually when we want to do release.
`release.yaml` workflow will run only on that tag and made release on npm registry.

yarn.lock
evidolob: `next` is only version tag, we publish package with some real version, which include hash of the commit, and point `next` to that version.

joshuawilson (approved):

Issue #239, opened 2020-02-18
Auto completion can't suggest string enums correctly in Flow Style content.


chuyangguangshin:
When write a piece of yaml in flow style, Auto completion can't suggest string enums.

Here is a demo:
![image](https://user-images.githubusercontent.com/6897710/74714867-64d45400-5266-11ea-8af8-354b08ff7d65.png)

while in indent mode, it can suggest right choices:
![image](https://user-images.githubusercontent.com/6897710/74714984-9816e300-5266-11ea-890f-2bc7d4574000.png)

However, if you already input some characters , the suggestion just works again event in flow style :
![image](https://user-images.githubusercontent.com/6897710/74715233-0c518680-5267-11ea-9956-2843cb68b17a.png)

schema.json used in the demo:

`{
    "$schema": "http://json-schema.org/draft-07/schema",
    "type":"object",
    "properties": {
        "A" : {
            "type" : "string",
            "enum": [
                "a1",
                "a2"
            ]
        },
        "B" : {
            "type":"string",
            "enum": [
                "b1",
                "b2"
            ]
        }
    }
}`

PR #342, opened 2020-10-27
Fix completion on flow style mapping

evidolob:
### What does this PR do?
Fix completion in one line flow mapping, like `{A: <completion invoke there>, B: b1}`.

Fix: #239
Depends on: https://github.com/redhat-developer/yaml-ast-parser/pull/10

### Is it tested? How?
See #239, it contains YAML and JSON Schema to test this.


src/languageservice/services/yamlCompletion.ts
```diff
@@ -195,6 +195,16 @@ export class YAMLCompletion extends JSONCompletion {
             }
           }
         }
+        if (node.type === 'null') {
```
joshuawilson, line 198: Are you checking if there is a String "null" or if it actually `null`?

evidolob: I'm checking on string `'null'`

joshuawilson: ok

joshuawilson (approved):

JPinkney (approved):

PR #347, opened 2020-11-09
Add isKubernetes option to doHover

JPinkney:

### What does this PR do?
This PR adds isKubernetes option to doHover. Basically what isKubernetes does is allows you to use a [special comparator](https://github.com/redhat-developer/yaml-language-server/blob/master/src/languageservice/parser/jsonParser07.ts#L585)  when evaluating oneOf/anyOf. This is needed because the [generic comparator](https://github.com/redhat-developer/yaml-language-server/blob/master/src/languageservice/parser/jsonParser07.ts#L1243) provided works well in the general case but breaks down when you have a ton of anyOf/oneOf like in the kubernetes schema: 

Part of https://bugzilla.redhat.com/show_bug.cgi?id=1888874

### Is it tested? How?

You'll need to set isKubernetes=true in yamlHover.ts before building then set:
```
"yaml.schemas": {
 "https://gist.githubusercontent.com/JPinkney/fbbb95f3708e8b33f81267b243981ff6/raw/09af433aff3972f5d72cfe5c69951e94c1497875/kubernetes-console.json": "test.yaml"
}
```
and use:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
```
as the yaml. Then hover over replicas, selector, etc and see that the hover is working as expected

I can write a test but the json schema for kubernetes is 70000 lines long so I could either have the test pull from the URL specified above or bring the schema into the project though checkout times would increase.

### DIFF
--- a/src/languageservice/parser/isKubernetes.ts
+++ b/src/languageservice/parser/isKubernetes.ts
@@ -1,0 +1,7 @@
+import * as Parser from './jsonParser07';
+
+export function setKubernetesParserOption(jsonDocuments: Parser.JSONDocument[], option: boolean): void {
+  for (const jsonDoc of jsonDocuments) {
+    jsonDoc.isKubernetes = option;
+  }
+}
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -30,6 +30,7 @@ import { ClientCapabilities } from 'vscode-languageserver-protocol';
 import { stringifyObject, StringifySettings } from '../utils/json';
 import { guessIndentation } from '../utils/indentationGuesser';
 import { TextBuffer } from '../utils/textBuffer';
+import { setKubernetesParserOption } from '../parser/isKubernetes';
 const localize = nls.loadMessageBundle();
 
 export class YAMLCompletion extends JSONCompletion {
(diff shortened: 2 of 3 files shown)


src/languageservice/parser/isKubernetes.ts
```diff
@@ -0,0 +1,7 @@
+import * as Parser from './jsonParser07';
+
+export function setKubernetesParserOption(jsonDocuments: Parser.JSONDocument[], option: boolean): void {
+  for (const jsonDoc in jsonDocuments) {
```
joshuawilson, line 4: `for...in` is designed for `Objects`. While it does work with `Arrays` it is usually better to use `Array.prototype.forEach()` or `for...of`.

joshuawilson (approved):

eddiezane:
@JPinkney just curious how you generated the schema in the gist of your testing example.

JPinkney:
@eddiezane The openshift console project creates the schema here: https://github.com/openshift/console/blob/master/frontend/public/module/k8s/openapi-to-json-schema.ts#L92 and I just saved it then uploaded it as a gist

evidolob (approved):

Issue #138, opened 2019-03-18
Use built-in Promises instead of custom Thenable


redborian:
Throughout the code base, custom Thenables are used. Now that Node has built-in Promises,  we should be able to leverage that.

PR #352, opened 2020-11-13
Rename all applicable Thenables to Promise

JPinkney:

### What does this PR do?
This PR renames all Thenables to Promises and removes a few pieces of code that are never called. There is 1 lingering Thenable that I didn't get rid of in server.ts. The type returned for formatterRegistration is a Thenable (all promises are thenable but not all thenables are promises) so we can't convert it.

https://github.com/redhat-developer/yaml-language-server/issues/138

### Is it tested? How?

Tested to see that the yamlValidation contribution, json schema store schemas, setting a schema manually, and the custom contributor api all still work and then ran npm run test

### DIFF
--- a/src/languageservice/services/schemaRequestHandler.ts
+++ b/src/languageservice/services/schemaRequestHandler.ts
@@ -3,14 +3,14 @@ import { IConnection } from 'vscode-languageserver';
 import { xhr, XHRResponse, getErrorStatusDescription } from 'request-light';
 import * as fs from 'fs';
 
-import { VSCodeContentRequest, CustomSchemaContentRequest } from '../../requestTypes';
+import { CustomSchemaContentRequest } from '../../requestTypes';
 import { isRelativePath, relativeToAbsolutePath } from '../utils/paths';
 
 /**
  * Handles schema content requests given the schema URI
  * @param uri can be a local file, vscode request, http(s) request or a custom request
  */
-export const schemaRequestHandler = (connection: IConnection, uri: string): Thenable<string> => {
+export const schemaRequestHandler = (connection: IConnection, uri: string): Promise<string> => {
   if (!uri) {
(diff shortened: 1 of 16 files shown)


src/languageservice/services/schemaRequestHandler.ts
```diff
@@ -45,20 +45,6 @@ export const schemaRequestHandler = (connection: IConnection, uri: string): Then
...
-    return connection.sendRequest(VSCodeContentRequest.type, uri).then(
-      (responseText) => {
-        return responseText;
-      },
-      (error) => {
-        return error.message;
-      }
-    );
-  }
-
```
evidolob, line 61: Can you elaborate more about deletting this code?

test/utils/testHelper.ts
```diff
@@ -59,23 +52,14 @@ export const workspaceContext = {
...
     });
-  } else if (Strings.startsWith(uri, 'vscode://')) {
-    return connection.sendRequest(VSCodeContentRequest.type, uri).then(
-      (responseText) => {
-        return responseText;
-      },
-      (error) => {
-        return error.message;
-      }
-    );
```
evidolob, line 78: Same there, is shemaas with `vscode://` protocol is not used anymore?

evidolob (approved):
LGTM

src/languageservice/services/schemaRequestHandler.ts
JPinkney: Yeah, I found out it's not actually used anywhere. I think it must have came in when I originally had copied the schemaRequestHandler over from the json language server. I looked into what it's used for on VSCode and it's just used for setting their schemas for the settings, keybindings, etc. Since YAML ls doesn't need any of those specific things it doesn't make sense to keep

joshuawilson (approved):

PR #359, opened 2020-11-20
Register a non-standard LSP content request for resolving schemas on the client side

JPinkney:

### What does this PR do?
This PR registers a non-standard LSP request used specifically for VSCode. This LSP request gets triggered when an HTTP/HTTPS request to a schema is being made. Instead of the server resolving the schema the language server will send the URI to the client-side to be resolved. This has to be done in order to circumvent signed certificate errors that happen when you are trying to fetch the schemas outside of VSCode. From my understanding, VSCode is using Electron which uses Chromium which has certs already installed so this is why we have to do it this way and can't just fetch the schemas normally. I assume this is also why vscode-json-languageservice does the same thing

https://github.com/redhat-developer/vscode-yaml/issues/107

### Is it tested? How?

The only way I know how to test this is on the Red Hat VPN.

You'll need the client-side changes: https://github.com/redhat-developer/vscode-yaml/pull/395

Once you connect to the VPN these schemas should resolve correctly
```
"yaml.schemas": {
    "https://gitlab.cee.redhat.com/cpaas/service/-/raw/development/resources/schemas/release-root.json": "release.yml",
    "https://gitlab.cee.redhat.com/cpaas/service/-/raw/development/resources/schemas/product-root.json": "product.yml"
  }
```

### DIFF
--- a/src/languageservice/services/schemaRequestHandler.ts
+++ b/src/languageservice/services/schemaRequestHandler.ts
@@ -1,16 +1,22 @@
 import { URI } from 'vscode-uri';
-import { IConnection } from 'vscode-languageserver';
+import { IConnection, WorkspaceFolder } from 'vscode-languageserver';
 import { xhr, XHRResponse, getErrorStatusDescription } from 'request-light';
 import * as fs from 'fs';
 
-import { CustomSchemaContentRequest } from '../../requestTypes';
+import { CustomSchemaContentRequest, VSCodeContentRequest } from '../../requestTypes';
 import { isRelativePath, relativeToAbsolutePath } from '../utils/paths';
 
 /**
  * Handles schema content requests given the schema URI
  * @param uri can be a local file, vscode request, http(s) request or a custom request
  */
-export const schemaRequestHandler = (connection: IConnection, uri: string): Promise<string> => {
+export const schemaRequestHandler = (
+  connection: IConnection,
+  uri: string,
+  workspaceFolders: WorkspaceFolder[],
+  workspaceRoot: URI,
+  useVSCodeContentRequest: boolean
+): Promise<string> => {
   if (!uri) {
     return Promise.reject('No schema specified');
   }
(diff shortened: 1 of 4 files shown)


evidolob:
> From my understanding, VSCode is using Electron which uses Chromium which has certs already installed so this is why we have to do it this way and can't just fetch the schemas normally.

That is true, but only for main vscode code, all extensions works in regular Node.JS instance, which spawn from Electron process. Maybe VSCode has some code which set up cert for extensions node.js process. Need to look on that.

JPinkney:
Hmm, it looks like `ps aux` only shows the language server's running in their own process, I can't find any of the client-side side stuff. Do you know of any way to find the VSCode extension client-side process?

evidolob:
Yes, try `ps -ax | grep bootstrap-fork` and look on `--type=extensionHost` option.
[Link on source ](https://github.com/Microsoft/vscode/blob/1273299f3d5849b0966cb3187035b378871889f3/src/vs/workbench/services/extensions/electron-browser/localProcessExtensionHost.ts#L224)
It seems that use `fork` function to spawn process for extensions.

JPinkney:
@aeschli I see that you brought the json schema fetching for http and https into the ext host quite a long time ago (https://github.com/microsoft/vscode/commit/849d9417eab3d8f85792070247bdc99247ce174a), was your reasoning because of self-signed certificates like ours? User's keep running into self signed certificate issues and I'm not sure of a clear way to resolve this other then grabbing the schemas from the extension host as well

evidolob:
@JPinkney I made some investigation on this.
According to [doc](https://code.visualstudio.com/docs/setup/network#_ssl-certificates) VSCode uses Chromium mechanisms to work with TLS. 
It seems that this PR is only reliable way to handle the situation with VPN or self signed certificates.

> What is still currius for me, that vscode run extensions and yaml-ls in same buildin node.js (in different process) with [fork](https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options)  function. So for me, is still unclear, what make the difference in https requests.

evidolob (approved):

Issue #344, opened 2020-11-01
Issue with UI5 YAML  in 0.12.0


AlexanderDemidovich74:
## Describe the bug
No completion for https://sap.github.io/ui5-tooling/schema/ui5.yaml.json

It works in version 0.11.1 in VSCode (though, not completely).  Now the only first object specVersion is suggested...
ajv-cli validates the project's ui5.yaml against this schema without any problems.

## Expected Behavior
Completion support (ctrl+space)

## Environment
VS Code.

evidolob:
@AlexanderDemidovich74 This bug may be fixed with https://github.com/redhat-developer/yaml-language-server/pull/357 as ui5 schema use `if/then/else` conditions. Can you verify?

evidolob:
Fille free to reopen this if you still has a problem

PR #360, opened 2020-11-20
Fix error on completion 'null' value

evidolob:
### What does this PR do?
Fix error during completion when schema contains `null` value in enum, example:
```json
{
   "type": "object",
   "properties": {
      "kind": {
        "enum": [
          "project",
          null
       ],
     }
   }
}
```
With this PR error is gone, but I have doubt about behaviour. 
We generate completion item with label `null` but insert only new line`\n`.
According to yaml spec
```yaml
kind:
```
and 
```yaml
kind: null
```
are equal, but maybe it would be better if we will insert `null\n` instead of just `\n`?
Found during https://github.com/redhat-developer/yaml-language-server/issues/344 investigation.

### Is it tested? How?
Use example schema and yaml content:
```yaml
kind: 
```
call completion after `kind: `, you should see completion list with two items `project` and `null`.

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -899,6 +899,9 @@ export class YAMLCompletion extends JSONCompletion {
   }
 
   private getLabelForValue(value: string): string {
+    if (value === null) {
+      return 'null'; // return string with 'null' value if schema contains null as possible value
+    }
     return value;
   }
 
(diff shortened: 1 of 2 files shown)


JPinkney:
> According to yaml spec
> 
> kind:
> and
> 
> kind: null
> are equal, but maybe it would be better if we will insert null\n instead of just \n?

I think I like the idea of inserting `null\n` a bit better. I tried with it just inserting `\n` but I'm not sure when anyone would actually use the completionitem for that rather then just pressing enter but I could be wrong

evidolob:
@JPinkney Done, just note, to alight with other string completion we will insert just `null` string, without `\n`

joshuawilson (approved):

JPinkney (approved):

PR #362, opened 2020-11-26
Determine which schemas to use based off of priority

JPinkney:

### What does this PR do?
This PR implements schema priority. The priority level is modeline > settings schema > schema association > schema store, meaning that if you have a modeline in the file then it overrides everything else associated with the file. Likewise, if you have schema set on the file manually through yaml.settings it will override schema associations and the schema store.

### Is it tested? How?

You can test manually by creating a .travis.yml file. Observe that it is associated with the .travis.yml schema. Then associate a schema manually to .travis.yml
```
"yaml.schemas": {
    "https://json.schemastore.org/composer": ".travis.yml"
}
```
and observe that the composer schema is now picked up and the travis schema from the schema store is not


ssbarnea:
I was surprised today to observe that the extension decided to wrongly use "phpdeploy" schema from unknown location regardless the fact that I did configure it to use ansible-playbook schema for that file pattern.

I think it should have being obvious that user-configured `yaml.schemas` in settings.json should take precendence.

evidolob (approved):

PR #363, opened 2020-11-26
Keep space before word after inserting completion

evidolob:
### What does this PR do?
When completion invoked on field which has value:
`name: <cursor>aName` and file has scheme which contains `defaultSnippets` for `name`: 
```json
"properties": {
      "name": {
          "type": "string",
          "defaultSnippets": [
              {
                  "label": "some",
                  "body": "some"
              }
          ]
      },
}
```
it contains one option `some`, but inserting of that item will break yaml, as it 'eats' space after `:`:
`name:someaName`, this PR fix that, and now it no break yaml, so editor should contain `name: someaName`.

Found during work on https://github.com/redhat-developer/vscode-tekton/issues/429

### Is it tested? How?
You may use schema example from this description, or try complete any yaml mapping value. Completion shouldn't breck  yaml syntax.

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -63,6 +63,7 @@ export class YAMLCompletion extends JSONCompletion {
     if (!this.completion) {
       return Promise.resolve(result);
     }
+    const originalPosition = Position.create(position.line, position.character);
     const completionFix = this.completionHelper(document, position);
     const newText = completionFix.newText;
     const doc = parseYAML(newText);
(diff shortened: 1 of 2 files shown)


JPinkney (approved):
The test case worked for me. I just found out we have another semi-related issue that is present on master as well though

If we have an empty document and then auto complete at the root with the schema given we will get:
```
name:some
```
instead of
```
name: some
```

It's unrelated to this PR but I found it while testing :smile:. I'll open an issue for it

EDIT: The issue is here: https://github.com/redhat-developer/yaml-language-server/issues/364

joshuawilson (approved):

PR #365, opened 2020-11-27
Boolean value should use YAML 1.2 spec not 1.1

JPinkney:

### What does this PR do?
This PR makes it so that boolean values are parsed in accordance with the YAML 1.2 spec and not the YAML 1.1 spec. The difference is that in the [YAML 1.2 spec](https://yaml.org/spec/1.2/spec.html) under heading `10.2.1.2. Boolean` only true/false are booleans whereas in the [YAML 1.1](https://yaml.org/type/bool.html) spec no, yes, on, off are boolean values as well

https://github.com/redhat-developer/vscode-yaml/issues/396

### Is it tested? How?

I've modified the existing tests to make sure that the YAML 1.2 spec is used not the YAML 1.1 spec


gorkem:
Is this going to break the users that depend on the YAML 1.1 behaviour? Is there a need for us to implement this in a way that those users can choose to use 1.1 behaviour?

JPinkney:
Technically yes. We could add detection for something like `%YAML 1.1` at the beginning of the file but then this would be the only 1.1 spec feature and other users would complain that their 1.1 feature isn't working. (There was already a request for that before).

I think the only way we could actually support all yaml specs realistically would be switching to: https://github.com/eemeli/yaml. When I originally looked at that project a few years ago it was very new and nobody was using it but now it was support for YAML 1.0, 1.1, 1.2, it's fully tested against all three of those specs, and has an active-ish community around it

ssbarnea:
While Ansible itself is still using YAML 1.1 specification, including in their examples, many Ansible users are using tools like yamllint/ansible-lint to ensure the more restrictive YAML 1.2 spec, which is backwards compatible.

I personally salute this move, especially around the booleans and I would not complain if the language server would be more restrictive than the consumers.

ssbarnea (approved):
Why was not this approved yet? Those 1.1 spec booleans are pure EVIL.

evidolob:
I think we can close this, as we almost there with new yaml parser https://github.com/redhat-developer/yaml-language-server/pull/442 (there are only 3 test left to fix). And thous parser is more strict and has support of `1.1`  and `1.2` yaml spec. We will have preference to chose default spec version. By default we will be in `1.2` mode.

evidolob:
@JPinkney We can close this, as #442 is merged.

PR #372, opened 2020-12-10
Fix empty completion list on empty file

evidolob:
### What does this PR do?
Fixes empty completion list on empty file. I found that we add `holder: \n` text in to text document, if line on which completion called is empty(contains only white spaces and line delimiter). 
Such behaviour do not to allow to choose proper schema, as some(k8's) schemas has `additionalProperties: false`, 
but AST contains property `holder`.
This PR fix that, so if file is empty we add `{}`(empty object)

Fix: https://github.com/redhat-developer/vscode-yaml/issues/349

### Is it tested? How?
Configure vscode-yaml with 
```json
"yaml.schemas": {
    "kubernetes": [
      "kubernetes.yaml"
    ]
}
```
then open `kubernetes.yaml`, file should be empty, and call code completion, you should see :
Screenshot 2020-12-10 at 14 31 13

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -936,11 +936,7 @@ export class YAMLCompletion extends JSONCompletion {
     if (document.getText().trim().length === 0) {
       return {
         // add empty object to be compatible with JSON
-        newText:
-          document.getText().substring(0, start + textLine.length) +
-          (textLine.trim()[0] === '-' && !textLine.endsWith(' ') ? ' ' : '') +
-          '{}\n' +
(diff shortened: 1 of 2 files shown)


JPinkney (commented):
I just played around with it and it works great when you have autocomplete on the first line and last line but for some reason it doesn't seem to work if you try to autocomplete on a line in the middle of the file (though I'm not sure if thats related to this PR or not, it might not have been working before)

E.g.
```

<- autocompleteting here

```
seems to not get any results

evidolob:
@JPinkney Yes, that is problem for this PR, going to fix it

evidolob:
@JPinkney Now completion should work on any line of "empty" file

JPinkney (approved):
Works great!

Issue #377, opened 2020-12-15
Server is sending client/registerCapability even if the client doesn't support dynamic capability registration


hugopl:
## Server is sending `client/registerCapability` even if the client doesn't support dynamic capability registration.

Here's a dump of the JSON RPC communication, using yaml-language-server version 0.13.0 on [Tijolo](https://github.com/hugopl/tijolo) editor from git.

```
LSP.Yaml: Starting LSP for "yaml-language-server --stdio" on pid 9566

LSP.Yaml: ==> {"jsonrpc":"2.0","id":0,"method":"initialize","params":{"processId":9552,"rootPath":"/home/hugo/src/tijolo","rootUri":"file:///home/hugo/src/tijolo","capabilities":{"workspace":{"applyEdit":true,"workspaceEdit":{"documentChanges":true},"didChangeConfiguration":{"dynamicRegistration":false},"didChangeWatchedFiles":{"dynamicRegistration":false},"workspaceFolders":true},"textDocument":{"synchronization":{"dynamicRegistration":false,"willSave":true,"willSaveWaitUntil":false,"didSave":true},"definition":{"dynamicRegistration":false},"documentHighlight":{"dynamicRegistration":false},"documentSymbol":{"dynamicRegistration":false,"symbolKind":{"valueSet":[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26]},"hierarchicalDocumentSymbolSupport":false}}},"trace":"off","workspaceFolders":[{"uri":"file:///home/hugo/src/tijolo","name":"project"}]}}

LSP.Yaml: <== {"jsonrpc":"2.0","id":0,"result":{"capabilities":{"textDocumentSync":1,"completionProvider":{"resolveProvider":false},"hoverProvider":true,"documentSymbolProvider":true,"documentFormattingProvider":false,"documentRangeFormattingProvider":false,"documentLinkProvider":{},"workspace":{"workspaceFolders":{"changeNotifications":true,"supported":true}}}}}

LSP.Yaml: ==> {"method":"initialized","params":{},"jsonrpc":"2.0"}

LSP.Yaml: <== {"jsonrpc":"2.0","id":0,"method":"client/registerCapability","params":{"registrations":[{"id":"d9f128c6-616e-40bd-a944-9d463f0c846b","method":"workspace/didChangeWorkspaceFolders","registerOptions":{}}]}}

[...]

Issue #370, opened 2020-12-09
Completion not working with neovim/neovim-lspconfig


david0:
## Describe the bug
I'm trying to use yaml-language-server with neovim + [neovim-lspconfig](https://github.com/neovim/nvim-lspconfig#yamlls).

It looks like its not working at all.

## Expected Behavior
Completion for kubernetes yaml files etc

## Current Behavior
No completion is provided, e.g. if I try to complete a kubernetes yaml file:
![vim](https://user-images.githubusercontent.com/1685610/101676054-e944cd00-3a5a-11eb-8ed7-d1f3f64d9163.png)

there is an error in my lsp.log file:
```
[ ERROR ] 2020-12-09T19:55:06+0100 ] ...ovim/HEAD-8fb786e/share/nvim/runtime/lua/vim/lsp/rpc.lua:455 ]  "rpc"   "/Users/david/.cache/nvim/lspconfig/yamlls/node_modules/.bin/yaml-language-server"      "stderr"        "/Users/david/.cache/nvim/lspconfig/yamlls/node_modules/vscode-jsonrpc/lib/main.js:436
                        responsePromise.reject(new messages_1.ResponseError(error.code, error.message, error.data));
                                               ^

ResponseError: MethodNotFound\n    at handleResponse (/Users/david/.cache/nvim/lspconfig/yamlls/node_modules/vscode-jsonrpc/lib/main.js:436:48)
    at processMessageQueue (/Users/david/.cache/nvim/lspconfig/yamlls/node_modules/vscode-jsonrpc/lib/main.js:263:17)
    at Immediate.<anonymous> (/Users/david/.cache/nvim/lspconfig/yamlls/node_modules/vscode-jsonrpc/lib/main.js:247:13)
    at processImmediate (node:internal/timers:463:21) {
  code: -32601,
  data: undefined
}"
```
see my [full lsp.log](https://github.com/redhat-developer/yaml-language-server/files/5668075/lsp.log) for more detail

## Steps to Reproduce
 NVIM v0.5.0-8fb786e41 + recent 
1. install neovim (i have  NVIM v0.5.0-8fb786e41) 
2. ~/.config/nvim/init.vim:
```
call plug#begin('~/.local/share/nvim/plugged')
Plug 'neovim/nvim-lspconfig'
call plug#end()

autocmd FileType * set omnifunc=v:lua.vim.lsp.omnifunc completeopt=noinsert,menuone
lua << EOF
[...]

PR #378, opened 2020-12-15
Check if dynamic registration is enabled before executing onDidChangeWorkspaceFolders

JPinkney:

### What does this PR do?
It seems that [vscode-languageserver](https://github.com/microsoft/vscode-languageserver-node/blob/master/server/src/common/workspaceFolders.ts#L39) automatically sends a `client/registerCapability` to the client if you use `onDidChangeWorkspaceFolders` on the server side. This means that currently the server will then send `client/registerCapability` even if the client doesn't support it. Since workspaceFolders don't have dynamicRegistration in the initialization according to the LSP spec, there's no way to tell if we should be using `onDidChangeWorkspaceFolders` and thus sending the `client/registerCapability` on the server side. This PR makes it so that `onDidChangeWorkspaceFolders` is only registered when the formatter dynamic registration is available

https://github.com/redhat-developer/yaml-language-server/issues/370
https://github.com/redhat-developer/yaml-language-server/issues/377

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -446,7 +446,7 @@ connection.onInitialize(
 );
 
 connection.onInitialized(() => {
-  if (hasWorkspaceFolderCapability) {
+  if (hasWorkspaceFolderCapability && clientDynamicRegisterSupport) {
     connection.workspace.onDidChangeWorkspaceFolders((changedFolders) => {
       workspaceFolders = workspaceFoldersChanged(workspaceFolders, changedFolders);
     });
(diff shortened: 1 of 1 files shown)


joshuawilson (approved):

joshuawilson:
@JPinkney I can't think of how to test this but I'm going to ask anyway, is there a way to write a test that covers this?

JPinkney:
>  is there a way to write a test that covers this?

Not currently, I have some code that completely refactors the language server so that you can test almost everything but I won't have it fully ready until friday-ish

joshuawilson:
I figured as much, that is why I approved without requiring the test.

evidolob (approved):

Issue #376, opened 2020-12-14
Array indentation in autocomplete is broken after upgrade to 0.13


yazaabed:
Hallo everyone! I hope you are having a good time! 

We have upgraded our implementation on top of yaml-language-server to latest version. And after successfully finish all needed things, we noticed that the array required fields autocomplete is broken. After debugging we found out that it is because of a small change in the completion service [here](https://github.com/redhat-developer/yaml-language-server/blob/b9fda68469a775253aaa4d507e8d56f5d749baaa/src/languageservice/services/yamlCompletion.ts#L803) and the code was changed to [this](https://github.com/redhat-developer/yaml-language-server/blob/b9b52253b93d460e1928da8e5c553f6cfcb0316f/src/languageservice/services/yamlCompletion.ts#L776)

Can you please give us a way to fix this issue? Or is it a bug from your side?

Thank you and Best regards!

JPinkney:
Do you have any particular schema and yaml that can reproduce the issue? I've just tried on VSCode with:
```
sample:
  - name: test
    email: test
    homepage: test
    role: test
``` 
and
```json
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "name": "Package",
    "type": "object",
    "properties": {
        "sample": {
            "type": "array",
            "items": {
                "type": "object",
                "required": [
                    "name",
                    "email",
                    "homepage",
                    "role"
                ],
                "properties": {
                    "name": {
                        "type": "string"
                    },
                    "email": {
                        "type": "string"
                    },
                    "homepage": {
                        "type": "string"
                    },
                    "role": {
                        "type": "string"
                    }
                }
            }
        }
    }
}
```
and haven't been able to reproduce yet

yazaabed:
Yes for first level array it is working fine but for nested arrays it is not working in the new version. For example we have this yaml.

```
components:
  - id: jsakdh
    type: list
    settings:
        data:
          arrayItems:
            - show: true
        id: asdgjajkdg <- this id here should be autocompleted under show not under data.
```

Schema
```
{
    "type": "object",
    "properties": {
        "components": {
            "type": "array",
            "items": {
                "type": "object",
                "properties": {
                    "id": {
                        "type": "string"
                    },
                    "type": {
                        "type": "string"
                    },
                    "settings": {
                        "type": "object",
                        "properties": {
                            "data": {
                                "type": "object",
                                "properties": {
                                    "arrayItems": {
                                        "type": "array",
[...]

PR #379, opened 2020-12-16
Fix indentation for object in array completion

evidolob:
### What does this PR do?
Fixes wrong indentation in case if object has array with object and it belongs to upper array, see schema in https://github.com/redhat-developer/yaml-language-server/issues/376#issuecomment-745850083

#376

### Is it tested? How?
Use schema from https://github.com/redhat-developer/yaml-language-server/issues/376#issuecomment-745850083
and yaml file content: 
```yaml
components:
  - id: ss
    type: dd
    se
```
call code completion after `se`, and insert `settings` completion item.
You should have same result as on:
ezgif com-gif-maker (7)


yazaabed:
Looks working fine. Here is a test example after applying the fix locally:

test

JPinkney (approved):

joshuawilson (approved):

Issue #323, opened 2020-09-25
Refactor 'server.ts' to make it testable


evidolob:
As discussed in https://github.com/redhat-developer/yaml-language-server/pull/322#issuecomment-698430029 we need to refactor `server.ts` to make it more suitable for unit tests.

JPinkney:
This should be fixed by https://github.com/redhat-developer/yaml-language-server/pull/381 so I'm going to close it

PR #381, opened 2020-12-18
Refactor the entire language server side so that everything can be tested

JPinkney:
### What does this PR do?
This PR refactors everything on the language server side so that it can be fully tested. It's worth to note that I didn't modify the way any of the language server-side code actually works, I just moved them into different files.

#### Changes in the first commit:
`server.ts` - now the server starter
`yamlServerInit.ts` - initializes the language server and registers features

`languageserver/handlers` - contains all the handlers for events.
- `languageHandlers` -> specific to language requests
- `notificationHandlers` -> specific to notification requests
- `requestHandlers` -> specific to request requests
- `settingsHandler` -> specific to handling/updating the settings
- `validationHandler` -> specific to validating (this is its own file instead of inside of languageHandlers because multiple different places in the language server need to call validation)

#### Changes in the second commit:
All references of textdocument have been updated to use textdocument from `vscode-languageserver-textdocument`

#### Changes in the third commit:
All tests have been updated to make it so that test the feature directly through the language server rather than the language service. This essentially means that before running a test you can set the state of the language server to be whatever you want, making a lot more tests possible.

https://github.com/redhat-developer/yaml-language-server/issues/323

### Is it tested? How?
Manually tested with VSCode and also ran all the tests


evidolob:
@JPinkney I got error when try to test this:
```
[Error - 9:36:29 AM] Notification handler 'workspace/didChangeConfiguration' failed with message: Cannot read property 'workspaceFolders' of undefined
```

I use same project to tests yaml-ls, and without this changes it was works normally.

src/yamlSettings.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export interface JSONSchemaSettings {
+  fileMatch?: string[];
+  url?: string;
+  schema?: JSONSchema;
+}
+
+// This class is responsible for handling all the settings
+export class SettingsState {
+  public yamlConfigurationSettings: JSONSchemaSettings[] = undefined;
```
evidolob, line 38: It is matter of taste, but I think we can remove all `public` keyword, as all fields/method are public by default in TS

JPinkney:
`[Error - 9:36:29 AM] Notification handler 'workspace/didChangeConfiguration' failed with message: Cannot read property 'workspaceFolders' of undefined` should be fixed now

evidolob:
@JPinkney Yes, it is fixed, everything seems to work

evidolob (approved):

PR #382, opened 2020-12-22
Hover doesn't work when there is now symbol after property

p-spacek:
### What does this PR do?
Fix jsonParser when objects is null and doesn't continue with any symbol in yaml file.

no issue reference
problem:

nullPropHover

Hover doesn't work when there is now symbol after property.
On this example no hover will be displayed on `objProperty`
```
obj:
  objProperty:
```

### Is it tested? How?
`hover.test.ts: 'Hover on null property'`


JPinkney (approved):
Just tested it out and it works great! Thanks for the PR

evidolob (approved):

PR #385, opened 2021-01-04
fix: codeCompletion-with-defaultSnippet-and-makdown

p-spacek:
### What does this PR do?
enable defaultSnippet markdown description in schemas for code completion service

when schema has defaultSnippet and markdown description parameter set, jsonCompletion will ignores this markdown description because of undefined parameter `supportsMarkdown`

### Is it tested? How?
added unit-test

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -31,7 +31,6 @@ import {
   WorkspaceContextService,
   SchemaConfiguration,
   SchemaPriority,
-  LanguageService,
 } from './languageservice/yamlLanguageService';
 import * as nls from 'vscode-nls';
 import {
--- a/src/server.ts
+++ b/src/server.ts
@@ -31,6 +31,7 @@ import {
   WorkspaceContextService,
   SchemaConfiguration,
   SchemaPriority,
+  LanguageService,
 } from './languageservice/yamlLanguageService';
 import * as nls from 'vscode-nls';
 import {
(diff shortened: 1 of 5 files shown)


JPinkney (commented):
I think we will probably have to check the client capabilities similar to how vscode-json-languageservice does it: https://github.com/microsoft/vscode-json-languageservice/blob/36f8016a33df9eff3b6edd584605aea349b73745/src/services/jsonCompletion.ts#L968. Just to make sure that the client supports markdown before setting `supportsMarkdown = true`

p-spacek:
You are absolutely right!

So I used functionality from original json-service.
But I'm not sure if it safe to move initialization of `customLanguageService = getCustomLanguageService` into `connection.onInitialize(`

src/server.ts
```diff
@@ -394,7 +395,7 @@ const schemaRequestHandlerWrapper = (connection: IConnection, uri: string): Prom
 
 const schemaRequestService = schemaRequestHandlerWrapper.bind(this, connection);
 
-export const customLanguageService = getCustomLanguageService(schemaRequestService, workspaceContext);
+let customLanguageService: LanguageService;
```
JPinkney, line 398: I think reassigning inside of onInitiailize seems to work. That's what vscode-json-languageserver does: https://github.com/vscode-langservers/vscode-json-languageserver/blob/master/src/jsonServer.ts#L80. I think we'd probably still have to do:
```suggestion
let customLanguageService = getCustomLanguageService(schemaRequestService, workspaceContext, ClientCapabilities.LATEST);
```
just in case though

src/server.ts
```diff
@@ -31,6 +31,7 @@ import {
   WorkspaceContextService,
   SchemaConfiguration,
   SchemaPriority,
+  LanguageService,
```
evidolob, line 34: `LanguageService` is unused, can you delete it?

JPinkney: Once this is removed the PR LGTM

JPinkney (approved):

evidolob (approved):

PR #388, opened 2021-01-11
feat: add new params to completion snippet

p-spacek:
### What does this PR do?
When object is type anyOf (array: null | string[]) some code completion doesn't work properly.
For example snippets on array it's not working very well.
```yaml
"anyOf_arrayObj": {
  "anyOf": [
    {
      "type": "array",
      "items": {
        "type": "object"
      }
    },
    {
      "type": "null"
    }
  ],
  "defaultSnippets": [
    {
      "label": "- (array item)",
      "type": "array",
      "suggestionKind": 9,
      "body": {
        "key": ""
      }
    }
  ]
},
```
Helps also in scenarios where yaml parser doesn't know that parent is array:
```yaml
anyOf_arrayObj:
  # invoke array completion here
```

This new parameters can specify:
 - `type`: type of snippet.
 - `symbolKind`: id of symbol that is displayed on completion panel. 
 
This feature probably doesn't solve the cause of the problem. But offers workaround.

[Completion-Snippet: add parameters to defaultSnippets to specify the type and simbolKind](https://github.com/p-spacek/yaml-language-server/issues/7)

### Is it tested? How?
added unit test


src/languageservice/jsonSchema.ts
```diff
@@ -1,3 +1,5 @@
+import { CompletionItemKind } from 'vscode-json-languageservice';
```
evidolob, line 1: Could you move import under license header

p-spacek: sorry, done

p-spacek (commented):
done

src/languageservice/services/yamlCompletion.ts
```diff
@@ -535,7 +535,7 @@ export class YAMLCompletion extends JSONCompletion {
           filterText = insertText.replace(/[\n]/g, ''); // remove new lines
         }
         collector.add({
-          kind: this.getSuggestionKind(type),
+          kind: s.suggestionKind || this.getSuggestionKind(type),
```
JPinkney, line 538: Is this change needed for this PR? If so I think we should have a simple test case covering it just to make sure everything works

p-spacek: I'm using it with custom array snippets where I need to set custom symbol.
I added UT for this case.

If you think that custom suggestionKind shouldn't be there. I can remove it and add it locally.

JPinkney (approved):

src/languageservice/services/yamlCompletion.ts
JPinkney: I think you can keep it in

evidolob:
@p-spacek Sorry for the delay, there are some tests fail. Could you fix them ASAP? I want to include this PR in to next release.

p-spacek:
@evidolob, I fixed the test.

evidolob (approved):

PR #389, opened 2021-01-12
Add 'doDocumentOnTypeFormatting' implementation

evidolob:
### What does this PR do?
Add implementation for [doDocumentOnTypeFormatting](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_onTypeFormatting) feature.
It allows to format text as user typing. This implementation reacts on new line character and add proper indentation. Also it automatically add `- ` for array items if previous line contains list simple array item(not map or array).

Demo:

https://user-images.githubusercontent.com/929743/104292252-43ef9f80-54c5-11eb-96bb-d70383c8ba73.mov

https://github.com/redhat-developer/vscode-yaml/issues/179

### Is it tested? How?
If you want to test it manually you need to make sure that you enable `"editor.formatOnType"` for yaml files, see first few second on demo.


JPinkney:
This looks like a really cool feature. I usually never use formatOnType but in this case it seems really useful

test/yamlOnTypeFormatting.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  Position,
+  Range,
+  TextDocument,
+  TextEdit,
+} from 'vscode-languageserver';
+import { TextDocumentIdentifier } from '../src';
+import { doDocumentOnTypeFormatting } from '../src/languageservice/services/yamlOnTypeFormatting';
+
+const FILE_URI = 'file://some/file.yaml';
+function createDocument(content: string): TextDocument {
```
JPinkney, line 18: I think you can probably re-use the ones from testHelper https://github.com/redhat-developer/yaml-language-server/blob/master/test/utils/testHelper.ts#L32

JPinkney (commented):
LGTM, I think you just need to rebase it

JPinkney (approved):

PR #395, opened 2021-01-18
Add code action to open json schema

evidolob:
### What does this PR do?
Adds code action to open json schema, this code action will be available only for JSON Schema specific error/warnings, only for schemas which URI has `file`(local schemas) or `https` (remote, usually for schema store).

Demo:
ezgif com-gif-maker (10)

ezgif com-gif-maker (9)

> Note: to be able to open files from LS this PR also update `vscode-languageclient` to `7.0.0` version

VSCode extension part: https://github.com/redhat-developer/vscode-yaml/pull/418

https://github.com/redhat-developer/vscode-yaml/issues/174

### Is it tested? How?

To test this you need https://github.com/redhat-developer/vscode-yaml/pull/418 also, just build both and run extension, open any yaml file with schema associated, make some error in it, after error appeared  in editor, there are yellow bulb should appeared for line with error, just click on it, an click on `Jump to schema location` and you should see JSON scheme opened in editor.

### DIFF
--- a/src/languageservice/services/yamlCodeActions.ts
+++ b/src/languageservice/services/yamlCodeActions.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { TextDocument } from 'vscode-languageserver-textdocument';
-import { CodeAction, CodeActionParams, Command, Connection, Diagnostic } from 'vscode-languageserver';
+import { ClientCapabilities, CodeAction, CodeActionParams, Command, Connection, Diagnostic } from 'vscode-languageserver';
 import { YamlCommands } from '../../commands';
 import { CommandExecutor } from '../../languageserver/commandExecutor';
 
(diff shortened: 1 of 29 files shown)


src/languageservice/services/yamlCodeActions.ts
```diff
@@ -0,0 +1,95 @@
...
+    return result;
+  }
+}
+
+/**
+ * Check if rangeA and rangeB is intersect
+ * @param rangeA
+ * @param rangeB
+ */
+function isIntersect(rangeA: Range, rangeB: Range): boolean {
```
JPinkney, line 66: I wonder if it it would be worth it to extract this to one of the files in `languageserver/utils`, its upto you

JPinkney (commented):
Other than my one comment and the github action comments it looks good to me

JPinkney (approved):
LGTM

Issue #244, opened 2020-03-11
Auto-Edit to turn indent tabs into indent spaces


mickaelistria:
See https://github.com/eclipse/wildwebdeveloper/issues/372 and https://yaml.org/spec/1.2/spec.html#id2777534 . YAML spec enforces indentation with spaces.
So the language server should provide an -opt-out (since it's spec'd, it should be default- auto conversion on indent tabs into indent spaces.

mickaelistria:
Any chance this can be addressed soon? This is one of the main issue users of Eclipse IDE are facing with Yaml and I'd rather avoid implementing the language-specific feature in the client if this can be addressed by the LS according to the language specification.

JPinkney:
The only way I can think of doing it is by providing enhancements to `OnTypeFormatting` once https://github.com/redhat-developer/yaml-language-server/pull/389 is merged

evidolob:
I think we can provide [CodeAction](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction) to convert tab indentation to space indentation.

mickaelistria:
OnTypeFormatting seems ideal (although it's not supported yet in LSP4E, so we'd have to add it, but it would be a good enhancement).
CodeActions are fine, but they require explicit user action, he I'd like things to happen by default as users type.
Another approach is to send some workspace/applyEdits from the server to the client.

evidolob:
@mickaelistria Hm, maybe I miss understand the issue. You want to convert tab to space when user type it?

mickaelistria:
@evidolob yes, exactly.

evidolob:
OK, now I'm, rebasing #389 , I can add such conversion

evidolob:
What we can do more on this, I think we can CodeAction to convert tabs in to space for existing files, as we generate error for tab character.

mickaelistria:
I think having a codeaction to fix an error (whichever one) is always welcome!

PR #396, opened 2021-01-20
Add on type conversion of tab char to spaces

evidolob:
### What does this PR do?
Add on type conversion of tab char to spaces

#244

### Is it tested? How?
With unit test. Unfortunately vscode doesn't trigger onTypeFormatting when `\\t` character is typed, maybe I need to report an issue for it.

### DIFF
--- a/src/languageservice/services/yamlOnTypeFormatting.ts
+++ b/src/languageservice/services/yamlOnTypeFormatting.ts
@@ -45,5 +45,14 @@ export function doDocumentOnTypeFormatting(
       return [TextEdit.insert(position, '  ')];
     }
   }
+
+  if (params.ch === '\t' && params.options.insertSpaces) {
+    return [
+      TextEdit.replace(
+        Range.create(position.line, position.character - 1, position.line, position.character),
+        ' '.repeat(params.options.tabSize)
+      ),
+    ];
+  }
   return;
 }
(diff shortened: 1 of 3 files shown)


evidolob:
@mickaelistria Can you check this PR, as I won't able to test it in vscode?

mickaelistria:
Unfortunately, LSP4E doesn't yet support onTypeFormatting ( https://bugs.eclipse.org/bugs/show_bug.cgi?id=537154 ) so I'm not able to test it either at the moment...

gorkem:
Closing as the original issue is closed and this is no longer worked on

mickaelistria:
Can someone ( @msivasubramaniaan maybe? ) please reopen this issue and consider reviewing/merging such code? This feature is extremely helpful for editors that do not have particular language-specific indentation configuration built-in, and allow to get proper Yaml directly with the LS (without need to configure other editor parts).

src/languageservice/services/yamlOnTypeFormatting.ts
```diff
@@ -45,5 +45,14 @@ export function doDocumentOnTypeFormatting(
       return [TextEdit.insert(position, '  ')];
     }
   }
+
+  if (params.ch === '\t' && params.options.insertSpaces) {
```
mickaelistria, line 49: I think the condition should be refined so that the replacement only happens before the line payload (ie no previous character on line or all previous chars on line are whitespaces)

datho7561:
I managed to rebase this PR locally and get the tests passing.

> Unfortunately vscode doesn't trigger onTypeFormatting when \\t character is typed, maybe I need to report an issue for it.

I started digging into the VS Code codebase to try and see why tab doesn't work as the trigger character, and didn't find any obvious reason. Space and newline work fine.

I'll file this upstream; it might take a while because they likely want a minimal reproducer. Until the issue is resolved, this change won't change anything in VS Code. So, we'll need to test this feature against a different LSP client.

datho7561:
See https://github.com/microsoft/vscode/issues/247651

PR #410, opened 2021-02-23
feat: multiple schema distinction in validation

p-spacek:
### What does this PR do?
When there are more possible warning messages from different schemas or objects connected to specific problem, validator takes only the first one and it's hard to recognize the core of the problem.

before:MissingProp-Before
after:  MissingProp-After

before:IncorrectType-Before
after:  IncorrectType-After
after:Const-After

jump to multiple schemas:
before:JumpToSchema-Before
after:  JumpToSchema-After

https://github.com/p-spacek/yaml-language-server/issues/10

### Is it tested? How?
Unit tests


src/languageservice/parser/jsonParser07.ts
```diff
@@ -23,6 +24,8 @@ import { URI } from 'vscode-uri';
 import { DiagnosticSeverity, Range } from 'vscode-languageserver-types';
 import { TextDocument } from 'vscode-languageserver-textdocument';
 import { Diagnostic } from 'vscode-languageserver';
+import { MissingRequiredPropWarning, TypeMismatchWarning, ConstWarning } from '../../../test/utils/errorMessages';
```
evidolob, line 27: If you are need types form tests code, then it would be better to extract them in to `src` folder

test/fixtures/testMultipleSimilarSchema.json
```diff
@@ -0,0 +1,79 @@
...
+          },
+          {
+            "$ref": "#/definitions/type3"
+          }
+        ]
+      }
+    },
+    "type": "object"
+  }
+}
```
evidolob, line 79: Could you add new line there

src/languageservice/parser/jsonParser07.ts
p-spacek: I removed this import, didn't notice it before.
Values are defined also in jsonParser07 file without reference to test file...

evidolob (approved):
Working good!

JPinkney (approved):
:+1:

p-spacek:
Thank you for really quick review and merge!!

PR #417, opened 2021-03-01
Fix 'fileMatch' pattern to match whole file name

evidolob:
### What does this PR do?
It add `/` to file match pattern, if pattern doesn't include it.
The original problem in `vscode-json-languageservice`, it the way how file match implemented there.
Basically it match only file name end(i.e. `arc.yaml` and `.mocharc.yaml` file patterns will match to `/some/path/.mocharc.yaml`) 
This PR is basically, trying to workaround that behaviour, by forcing matching against full file name.

`vscode-json-languageservice` has similar issues:

- https://github.com/microsoft/vscode/issues/67191
- https://github.com/microsoft/vscode/issues/44135

Demo:
Screenshot 2021-03-01 at 14 16 14

Fix: https://github.com/redhat-developer/vscode-yaml/issues/354
and fix: https://github.com/redhat-developer/vscode-yaml/issues/397

### Is it tested? How?
Just create `.mocharc.yaml` file, open it and make sure that JSON Schema selected properly.

### DIFF
--- a/src/languageserver/handlers/settingsHandlers.ts
+++ b/src/languageserver/handlers/settingsHandlers.ts
@@ -116,18 +116,17 @@ export class SettingsHandler {
    * AND the schema store setting is enabled. If the schema store setting
    * is not enabled we need to clear the schemas.
    */
-  public setSchemaStoreSettingsIfNotSet(): void {
+  public async setSchemaStoreSettingsIfNotSet(): Promise<void> {
     const schemaStoreIsSet = this.yamlSettings.schemaStoreSettings.length !== 0;
 
     if (this.yamlSettings.schemaStoreEnabled && !schemaStoreIsSet) {
-      this.getSchemaStoreMatchingSchemas()
-        .then((schemaStore) => {
-          this.yamlSettings.schemaStoreSettings = schemaStore.schemas;
-          this.updateConfiguration();
-        })
-        .catch(() => {
-          // ignore
-        });
+      try {
+        const schemaStore = await this.getSchemaStoreMatchingSchemas();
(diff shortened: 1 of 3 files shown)


JPinkney (approved):

joshuawilson:
Does not look like you added a test for this.

src/languageserver/handlers/settingsHandlers.ts
```diff
@@ -116,18 +116,17 @@ export class SettingsHandler {
...
-      this.getSchemaStoreMatchingSchemas()
-        .then((schemaStore) => {
-          this.yamlSettings.schemaStoreSettings = schemaStore.schemas;
-          this.updateConfiguration();
-        })
-        .catch(() => {
-          // ignore
-        });
+      try {
+        const schemaStore = await this.getSchemaStoreMatchingSchemas();
```
joshuawilson, line 124: I do not see where you are stubbing this promise in the test.

evidolob: Because I don't stub it, I stub `xhr` function https://github.com/redhat-developer/yaml-language-server/pull/417/files#diff-9c120b6ffef26acf5a9470c86608e41b18520b4ebdb317aded945ab236739631R40 which used in `getSchemaStoreMatchingSchemas` as the reasl code which I want to test is inside `getSchemaStoreMatchingSchemas`.

joshuawilson: I was wondering if that might be the case. Thanks for clearing that up.

joshuawilson (approved):

JPinkney (approved):

PR #418, opened 2021-03-01
Re-add schema priority levels

JPinkney:

### What does this PR do?
This PR re-adds in schema priority levels. Apparently, they got lost when I refactored everything


ssbarnea:
I am really interested about this change because I currently have a serious issue caused by overlapping file matching patterns, where vscode does not respect a given order. By being able to set priorities I can help it pick the right schema.

This has the potential to address https://github.com/ansible-community/vscode-ansible/issues/36 with minor changes on ansible extension. At this moment I cannot do anything to make it pick right schema.

evidolob (approved):
LGTM

Issue #412, opened 2021-02-24
Problems reported by the linter must include information about originating schema


ssbarnea:
### Is your enhancement related to a problem? Please describe.

Based on discussions from https://github.com/redhat-developer/yaml-language-server/issues/310#issuecomment-784198962 we identified a real need to disclose the schema that was used for a particular file.

### Describe the solution you would like
I will list several approaches with the note that addressing any of them should be enough.

Information we need to expose:
* Schema name (that is not the title of the element from within the schema that generated the issues)
* Schema link
* What determined the use of the schema: which match pattern, if it came from file-header, settings.json or schemaStore

We should not underestimate the chance of using the wrong schema due to wrong patterns, so all of these will help user identify the source of the problem and avoid creating

UI Locations:
* Find a way to expose schema used on the status bar, either by adding this information into the tooltip from YAML block, which now display a static text "Select Language Mode", or by adding a new status bar item that can display that info if the other cannot be altered
* Output category YAML Support should log a "message" level entry when is processing data that includes schema identification (link). Right now all we can see is "Sending response 'custom/schema/request" and activating verbose mode tracing does produce such a verbosity that makes impossible to use in practice (even so schema is still not mentioned)
* Problems reported by YAML extension do not include any contextual options (lightbulb), which would be another way to expose information about schema, like a goto schema option. I do see other extension exposing that. 

### Additional context
![example-of-lightbulb-menu](https://sbarnea.com/ss/Screen-Shot-2021-02-24-08-47-30.20.png)

evidolob:
>Find a way to expose schema used on the status bar, either by adding this information into the tooltip from YAML block, which now display a static text "Select Language Mode", or by adding a new status bar item that can display that info if the other cannot be altered

It is not possible with Language Server(the only way is adding custom extension to protocol)

I'm going to experiment with [CodeLens](https://code.visualstudio.com/api/language-extensions/programmatic-language-features#codelens-show-actionable-context-information-within-source-code) 
I want to add it on first line of file, it will contain schema info and on click it will open schema itself.

ssbarnea:
I am not sure how it would look using CodeLens. I am asking because I want to assure that when someone clicks on an entry from the found problems panel, I do expect to see somewhere which schema was used for that document. The file is opened and the error line is selected automatically, but if codelends display the schema at the top of the file, it will mean that is not visible to the user.

[...]

PR #424, opened 2021-03-16
Show all json schemas for yaml file in codelens

evidolob:
### What does this PR do?
Add CodeLens at the beginning of the file with all applied JSON Schema's, click on that CodeLans will open JSON Schema in separate tab:

https://user-images.githubusercontent.com/929743/111322351-81ada780-8671-11eb-94c7-1f98f0c0fe4c.mov

Als it improves opening schemas, now all possible type of schema is supported, including schema contributed from vscode-yaml api.

>Note: I still not able to add navigation to specific part of JSON Schema which cause an error, as we are losing information about positions during parsing JSON.

#412 

### Is it tested? How?
With tests and manual, just open any yaml file which has JSON Schema, CodeLens with name and file name should appears above first line.

### DIFF
--- a/src/languageservice/jsonSchema.ts
+++ b/src/languageservice/jsonSchema.ts
@@ -92,5 +92,5 @@ export interface JSONSchemaMap {
 }
 
 export function isJSONSchema(schema: JSONSchemaRef): schema is JSONSchema {
-  return !(schema instanceof Boolean);
+  return typeof schema !== 'boolean';
 }
--- a/src/languageservice/jsonSchema.ts
+++ b/src/languageservice/jsonSchema.ts
@@ -90,3 +90,7 @@ export interface JSONSchema {
 export interface JSONSchemaMap {
   [name: string]: JSONSchemaRef;
 }
+
+export function isJSONSchema(schema: JSONSchemaRef): schema is JSONSchema {
+  return !(schema instanceof Boolean);
+}
--- a/src/languageservice/jsonSchema.ts
+++ b/src/languageservice/jsonSchema.ts
@@ -90,7 +90,3 @@ export interface JSONSchema {
 export interface JSONSchemaMap {
   [name: string]: JSONSchemaRef;
 }
-
-export function isJSONSchema(schema: JSONSchemaRef): schema is JSONSchema {
-  return typeof schema !== 'boolean';
-}
--- a/src/languageservice/services/yamlCodeActions.ts
+++ b/src/languageservice/services/yamlCodeActions.ts
@@ -55,7 +55,7 @@ export class YamlCodeActions {
     for (const diagnostic of diagnostics) {
       const schemaUri = (diagnostic.data as YamlDiagnosticData)?.schemaUri || [];
       for (const schemaUriStr of schemaUri) {
-        if (schemaUriStr /* && (schemaUriStr.startsWith('file') || schemaUriStr.startsWith('https')) */) {
+        if (schemaUriStr) {
           if (!schemaUriToDiagnostic.has(schemaUriStr)) {
             schemaUriToDiagnostic.set(schemaUriStr, []);
           }
--- a/src/languageservice/services/yamlCodeActions.ts
+++ b/src/languageservice/services/yamlCodeActions.ts
@@ -10,7 +10,6 @@ import {
   CodeActionKind,
   CodeActionParams,
   Command,
-  Connection,
   Diagnostic,
   Position,
   Range,
@@ -19,7 +18,6 @@ import {
 } from 'vscode-languageserver';
 import { YamlCommands } from '../../commands';
 import * as path from 'path';
-import { CommandExecutor } from '../../languageserver/commandExecutor';
 import { TextBuffer } from '../utils/textBuffer';
 import { LanguageSettings } from '../yamlLanguageService';
 
@@ -29,21 +27,7 @@ interface YamlDiagnosticData {
 export class YamlCodeActions {
   private indentation = '  ';
 
-  constructor(commandExecutor: CommandExecutor, connection: Connection, private readonly clientCapabilities: ClientCapabilities) {
-    commandExecutor.registerCommand(YamlCommands.JUMP_TO_SCHEMA, async (uri: string) => {
-      if (!uri) {
-        return;
-      }
-      if (!uri.startsWith('file')) {
-        uri = 'json-schema' + uri.substring(uri.indexOf('://'), uri.length);
-      }
-
-      const result = await connection.window.showDocument({ uri: uri, external: false, takeFocus: true });
-      if (!result) {
-        connection.window.showErrorMessage(`Cannot open ${uri}`);
-      }
-    });
-  }
+  constructor(private readonly clientCapabilities: ClientCapabilities) {}
 
   configure(settings: LanguageSettings): void {
     this.indentation = settings.indentation;
@@ -71,7 +55,7 @@ export class YamlCodeActions {
     for (const diagnostic of diagnostics) {
       const schemaUri = (diagnostic.data as YamlDiagnosticData)?.schemaUri || [];
       for (const schemaUriStr of schemaUri) {
-        if (schemaUriStr && (schemaUriStr.startsWith('file') || schemaUriStr.startsWith('https'))) {
+        if (schemaUriStr /* && (schemaUriStr.startsWith('file') || schemaUriStr.startsWith('https')) */) {
           if (!schemaUriToDiagnostic.has(schemaUriStr)) {
             schemaUriToDiagnostic.set(schemaUriStr, []);
           }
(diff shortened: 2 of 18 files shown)


src/languageservice/services/yamlCodeActions.ts
```diff
@@ -71,7 +55,7 @@ export class YamlCodeActions {
     for (const diagnostic of diagnostics) {
       const schemaUri = (diagnostic.data as YamlDiagnosticData)?.schemaUri || [];
       for (const schemaUriStr of schemaUri) {
-        if (schemaUriStr && (schemaUriStr.startsWith('file') || schemaUriStr.startsWith('https'))) {
+        if (schemaUriStr /* && (schemaUriStr.startsWith('file') || schemaUriStr.startsWith('https')) */) {
```
JPinkney, line 58: Was this purposely commented out? If so does it make sense to remove it?

src/languageservice/services/yamlCodeLens.ts
```diff
@@ -0,0 +1,102 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Red Hat, Inc. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { TextDocument } from 'vscode-languageserver-textdocument';
+import { CodeLens, Range } from 'vscode-languageserver-types';
+import { YamlCommands } from '../../commands';
+import { yamlDocumentsCache } from '../parser/yaml-documents';
+import { YAMLSchemaService } from './yamlSchemaService';
```
JPinkney, line 10: I wonder if we should start making a standard for exported classes/interfaces that start with yaml. Right now we have 3 different styles from lines 8-10:
```
Yaml
yaml
YAML
```

test/yamlCodeLens.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    const schema: JSONSchema = {
+      url: 'some://url/to/schema.json',
+    };
+    yamlSchemaService.getSchemaForResource.resolves({ schema });
+    const codeLens = new YamlCodeLens((yamlSchemaService as unknown) as YAMLSchemaService);
+    const result = await codeLens.getCodeLens(doc, { textDocument: { uri: doc.uri } });
+    expect(result[0].range).is.deep.equal(Range.create(0, 0, 0, 0));
+  });
+
+  it('command name should contains schema name', async () => {
```
JPinkney, line 68: Is this test and the test underneath it the same?

src/languageservice/services/yamlCodeLens.ts
joshuawilson: If we treat it as a word that is pronounced then I think it should be `yaml` and `Yaml` depending if the first letter is capitalized. 
I know that I do not say Y A M L so lowercase works for me.

src/languageservice/jsonSchema.ts
```diff
@@ -90,3 +90,7 @@ export interface JSONSchema {
 export interface JSONSchemaMap {
   [name: string]: JSONSchemaRef;
 }
+
+export function isJSONSchema(schema: JSONSchemaRef): schema is JSONSchema {
+  return !(schema instanceof Boolean);
```
joshuawilson, line 95: I'm not sure this will work as expected. It might, I have not tested it. Capital B Boolean is the class Boolean. 
The var `schema` has a type of `JSONSchemaRef`, can it be a Boolean. Might need to make `schema` a union type. 
If it is a simple boolean then using `typeof` is better.

Since you are asking for the NOT of `instanceof`, it might always return the wrong answer. This needs thorough testing.

src/languageservice/services/yamlCodeLens.ts
evidolob: Yes we need naming conventions, not only for classes/interfaces but for TS file names also.
But I don't think that it should be a part of this PR. In this PR I add `yamlDocumentsCache` which is const instance of `YamlDocuments` class. `YAMLSchemaService` was created a way before, I think it name collide with `JSONSchemaService` which it extends.

JPinkney: Yeah, I don't think we need to address it in this PR, we can address it in a different one. It was just something I noticed when looking at the imports

JPinkney (approved):

src/languageservice/jsonSchema.ts
```diff
@@ -90,3 +90,7 @@ export interface JSONSchema {
 export interface JSONSchemaMap {
   [name: string]: JSONSchemaRef;
 }
+
+export function isJSONSchema(schema: JSONSchemaRef): schema is JSONSchema {
+  return typeof schema !== 'boolean';
```
joshuawilson, line 95: This check does not make sense to me. Maybe you can explain the idea of what you are trying to do.

Could `schema` ever be a boolean? It looks like it would be either be `JSONSchemaRef` or `undefined`. 

It looks like this might be a type-guard function. If so then maybe something like this would work:
```
export function isJSONSchema(schema: any): schema is JSONSchema {
  return schema instanceof JSONSchema;
}
```

joshuawilson: I can see below that this is indeed a type-guard. I do not know if the `JSONSchemaRef` is actually a boolean, I'm guessing it is not. If it is not then checking that it is not a boolean will show you nothing. It will always be true.

You need unit tests that tests both sides of this.

JPinkney: The JSONSchemaRef type is: `export type JSONSchemaRef = JSONSchema | boolean;`

gorkem:
I am not sure about the overall experience of this. I do not think there are many users out there who are interested to navigate to json schemas. Now this path is putting navigation to schema to the front of every yaml file for the purpose of listing the schemas that the yaml is validated against. I do not think code lenses is a good fit for that purpose.

evidolob:
@gorkem I don't think that we have much options on this, we have a lot of issues where users reports unexpected validation errors, just because yaml-ls uses wrong schema to validate file. The main goal for this PR is to show which JSON schemas used for specific yaml file, and navigation to thous schemas is optional.

An alternative way to show which schema used is to add new vscode statusbar item, which will be vscode only and requires new non standard LSP extension.

src/languageservice/jsonSchema.ts
evidolob: `isJSONSchema` is a type-guard function.
`JSONSchema` is interface, I cannot use `instanceof` to check  that `schema` type is `JSONSchema`

joshuawilson: This function is saying, if you check that it is NOT boolean then as long as it is anything but a boolean then it must be `JSONSchema`. Logically that does not make sense.

evidolob: But we only two choices, `schema`  can be boolean or `JSONSchema`
It would be enough to check that  `schema` is not boolean, as rest check will do TS compiler, it will not allow to pass anything to `isJSONSchema` which not looks like boolean or  `JSONSchema`.

src/languageservice/services/yamlCodeLens.ts
```diff
@@ -93,7 +94,7 @@ function addSchemasForOf(schema: JSONSchema, result: Map<string, JSONSchema>): v
 
 function addInnerSchemaUrls(schemas: JSONSchemaRef[], result: Map<string, JSONSchema>): void {
   for (const subSchema of schemas) {
-    if (isJSONSchema(subSchema)) {
+    if (!isBoolean(subSchema)) {
```
joshuawilson, line 97: This still does not confirm that it is a JSONSchema.

"I'm looking for a dog. Is a bird? No, it's not a bird. OK, it's not a bird so it must be a dog" That does not work.

evidolob: We have only two choices, `subSchema` can be boolean or JSONSchema
It would be enough to check that schema is not boolean, as rest check will do TS compiler.

joshuawilson: OK, since you are not using an `any` I guess that could work. 

I was thinking that if you wanted to confirm that matched the interface then test for expected mandatory fields. But I'll leave it up to you.

joshuawilson (approved):

andsens:
Is there any way this feature can be disabled? Sometimes it takes a while before the extension figures out the schema. When opening a file it can take 1-3 seconds before I find the line I am looking for, only for the codelens to pop in at that exact moment, shifting all the lines one down and completely ruining my concentration. It's... frustrating to say the least.

evidolob:
@andsens Just put 

```json
    "[yaml]": {
        "editor.codeLens": false
    }
```
in to your settings

gorkem:
@evidolob this info should be on README. on vscode-yaml

evidolob:
@gorkem  https://github.com/redhat-developer/vscode-yaml/pull/489

Issue #414, opened 2021-02-24
Wrong completion


jeffmaury:
## Describe the bug
Created a devfile.yaml whose corresponding schema is here: https://github.com/devfile/api/releases/download/v2.0.0/devfile.json and with the following content:

`schemaVersion: 2`

then remove the `n` before the `:` and ask for code completion: schemaVersion is proposed, accept it but content is now:

`schemaVersion: : 2`

## Expected Behavior
Generated content should be valid

## Current Behavior
Generated content is `schemaVersion: : 2`

## Steps to Reproduce
1. add in yaml.schemas: "https://github.com/devfile/api/releases/download/v2.0.0/devfile.json":"devfile.y*ml"
2. create a file devfile.yaml
3. enter `schemaVersion: 2`
4. remove the `n` before the `:`
5. trigger code completion
6. accept proposition

## Environment
- [X] Windows
- [ ] Mac
- [ ] Linux
- [ ] other (please specify)

PR #428, opened 2021-03-22
 Fix completion for existing property

evidolob:
### What does this PR do?
 Fix completion for existing property.
Demo:

https://user-images.githubusercontent.com/929743/111995901-26bcfa00-8b22-11eb-8c15-ee0c5f89ad41.mov

Fix: #414

### Is it tested? How?
With test.


JPinkney:
I just tested it out and this is the result I get when trying the changes:
Peek 2021-03-24 08-38
For some reason with mine it still adds the extra `:`

evidolob:
@JPinkney Good catch, that should be fixed now

JPinkney (approved):

PR #431, opened 2021-03-25
Use 'markdownDescription' for property completion

evidolob:
### What does this PR do?
Use 'markdownDescription' for property completion.
Demo:
Screenshot 2021-03-25 at 12 05 03

Fix: https://github.com/redhat-developer/vscode-yaml/issues/417

### Is it tested? How?
With test

### DIFF
--- a/test/autoCompletion.test.ts
+++ b/test/autoCompletion.test.ts
@@ -9,7 +9,7 @@ import assert = require('assert');
 import path = require('path');
 import { createExpectedCompletion } from './utils/verifyError';
 import { ServiceSetup } from './utils/serviceSetup';
-import { CompletionList, InsertTextFormat, MarkupContent } from 'vscode-languageserver';
+import { CompletionList, InsertTextFormat, MarkupContent, MarkupKind } from 'vscode-languageserver';
 import { expect } from 'chai';
 import { SettingsState, TextDocumentTestManager } from '../src/yamlSettings';
 import { LanguageService } from '../src';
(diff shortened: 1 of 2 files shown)


test/autoCompletion.test.ts
```diff
@@ -1840,6 +1840,33 @@ describe('Auto Completion Tests', () => {
...
+            type: 'string',
+            description: 'Kind is a string value representing the REST',
+            markdownDescription:
+              '**kind** (string)\n\nKind is a string value representing the REST resource this object represents.',
+          },
+        },
+        required: ['kind'],
+      });
+
+      const content = 'kin';
```
joshuawilson, line 1858: Should `kin` be `kind`?

evidolob: It should be `kin` to test completion to `kind`

joshuawilson: ok

joshuawilson (approved):

JPinkney (commented):
I think we might have to add a `CompletionClientCapabilities` check for the documentationFormat. If I'm understanding the spec correctly, under https://microsoft.github.io/language-server-protocol/specification#textDocument_completion it looks like clients can specify what documentation formats they support in `completionItem.documentationFormat`

evidolob:
@JPinkney In all cases we use [fromMarkup](https://github.com/microsoft/vscode-json-languageservice/blob/main/src/services/jsonCompletion.ts#L958) Which already has such check.
I will double check that we use that method in all cases.

Issue #353, opened 2020-11-14
Inconsistent way to generate whole property snippet and value snippet, when it contains \"


fbaligand:
Given that I have this property:
``` json
"properties": {
  "property1": {
    "type": "string",
    "default": "\"mydefault\""
  }
}
```

If I write `property1: ` and ask completion (Ctrl+tab), then choose first option, I get this generated:
`property1: "mydefault"`
That is nice.

But if I write `property1` and ask completion (Ctrl+tab), then I get this generated:
`property1: \"mydefault\"`

So the second case is inconsistent and is an issue.
Could you fix it?

fbaligand:
Hi,

I just tried YAML extension v0.18.0, and unfortunately, I still reproduce the issue :(
So can you reopen this issue?

evidolob:
@fbaligand Sure, that strange. Can you share schema where you have this?

fbaligand:
Hi @evidolob,

Thanks for reopening the issue.

My test schema to reproduce this issue is exactly the schema I provided in issue description.
The whole schema is:
``` json
{
	"$schema": "http://json-schema.org/draft-07/schema#",
	"type": "object",
	"properties": {
		"property1": {
			"type": "string",
			"default": "\"mydefault\""
		}
	}
}
```

fbaligand:
That said, to reproduce the issue, all characters are important:
- *If I write `property1: ` and ask completion (Ctrl+tab)* => there is a space after ":"
- *But if I write property1 and ask completion (Ctrl+tab)* => no ":" at end

fbaligand:
@evidolob 
Thanks for the fix!
Is there a (easy) way to try the fix before next release?

evidolob:
@fbaligand Yes you can follow this https://github.com/redhat-developer/vscode-yaml/blob/master/CONTRIBUTING.md#developing-the-client-and-server-together to build and run

PR #434, opened 2021-04-07
Remove escape character from default values

evidolob:
### What does this PR do?
Remove escape character `\"` from default values.
Demo:

https://user-images.githubusercontent.com/929743/113821333-a256b200-9784-11eb-8b1a-92e091a4b9d0.mov

#353 

### Is it tested? How?
With test and manually, use schema provided in #353 description, and try to complete `property1` value, it should be inserted without `\"` just `"`

### DIFF
--- a/test/autoCompletion.test.ts
+++ b/test/autoCompletion.test.ts
@@ -152,7 +152,8 @@ describe('Auto Completion Tests', () => {
           properties: {
             name: {
               type: 'string',
-              default: '\\"yaml\\"',
+              // eslint-disable-next-line prettier/prettier, no-useless-escape
+              default: '\"yaml\"',
             },
           },
         });
--- a/test/autoCompletion.test.ts
+++ b/test/autoCompletion.test.ts
@@ -146,6 +146,27 @@ describe('Auto Completion Tests', () => {
           .then(done, done);
       });
 
+      it('Autocomplete on default value with \\"', async () => {
+        languageService.addSchema(SCHEMA_ID, {
+          type: 'object',
+          properties: {
+            name: {
+              type: 'string',
+              default: '\\"yaml\\"',
+            },
+          },
+        });
+        const content = 'name: ';
+        const completion = await parseSetup(content, 6);
+        assert.strictEqual(completion.items.length, 1);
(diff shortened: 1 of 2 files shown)


test/autoCompletion.test.ts
```diff
@@ -146,6 +146,27 @@ describe('Auto Completion Tests', () => {
           .then(done, done);
       });
 
+      it('Autocomplete on default value with \\"', async () => {
+        languageService.addSchema(SCHEMA_ID, {
+          type: 'object',
+          properties: {
+            name: {
+              type: 'string',
+              default: '\\"yaml\\"',
```
joshuawilson, line 155: The regex is looking for `\"`. This example does not match the video. 
It does match the expected, unless the expected resolves one slash as an escape character. I'm just not sure.

evidolob: You need to test with schema:
```json
"properties": {
  "property1": {
    "type": "string",
    "default": "\"mydefault\""
  }
}
```
In test that is not JSON, it is JS object, in JS you need to escape `\` in string literals, so if I want to have `\"` in string literal I need to have `\\"`.

joshuawilson: That explains my confusion. Thanks.

joshuawilson (approved):

test/autoCompletion.test.ts
evidolob: No, I was wrong, it `eslint` requires to double escape `\` as it complains about unnecessary escaping. But anyway the result is the same. As for test I really need to have `\"some\"` string

JPinkney (approved):
LGTM

test/autoCompletion.test.ts
joshuawilson: that makes me feel better

evidolob: I found a better way

joshuawilson: That keeps it more accurate. Nicely done.

JPinkney: Yeah the new way makes a bit more sense!

PR #436, opened 2021-04-08
Invalidate document cache if custom tags provided

evidolob:
### What does this PR do?
Fixes errors related to custom tags

>Also it add fix which removes warnings on test output window related to sourcemap if launch test with VSCode launch

https://github.com/redhat-developer/vscode-yaml/issues/461

### Is it tested? How?
With tests and manually, just follow https://github.com/redhat-developer/vscode-yaml/issues/461 add custom tags and open any yaml file which contains custom tags, there are no errors should present.

### DIFF
--- a/src/languageservice/utils/arrUtils.ts
+++ b/src/languageservice/utils/arrUtils.ts
@@ -73,7 +73,7 @@ export function filterInvalidCustomTags(customTags: string[]): string[] {
   });
 }
 export function isArrayEqual(fst: Array<unknown>, snd: Array<unknown>): boolean {
-  if (!snd) {
+  if (!snd || !fst) {
     return false;
   }
   if (snd.length !== fst.length) {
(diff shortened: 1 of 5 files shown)


src/languageservice/utils/arrUtils.ts
```diff
@@ -73,7 +73,7 @@ export function filterInvalidCustomTags(customTags: string[]): string[] {
   });
 }
 export function isArrayEqual(fst: Array<unknown>, snd: Array<unknown>): boolean {
-  if (!snd) {
+  if (!snd || !fst) {
```
joshuawilson, line 76: Is this being tested? I don't see this function being called in the test.

NOT logic is easy to get wrong and cause unforeseen issues. It is better to test it.

evidolob: This is tested but not directly. I just found that `isArrayEqual` not handle situation when first argument is `undefined` or `null`. I do not think that providing test for  `arrUtils.ts` is a subject for this PR.

joshuawilson (approved):

src/languageservice/utils/arrUtils.ts
joshuawilson: I approved it. 
It is better to keep changes out of a commit / PR that do not directly address the issue. They should have their own commit/PR. If you include it then testing it is expected too.

JPinkney (approved):

PR #441, opened 2021-04-19
Remove jenkinsfile since it's no longer needed

JPinkney:

### What does this PR do?
This PR removes the jenkinsfile because it's no longer used for anything


joshuawilson:
You no longer use the Red Hat CI server to build/release this?

JPinkney:
Not as far as I know, unless @evidolob added them to a different jenkins instance. I think the only reason we had the Jenkinsfile was to build a tar for Che 6. The last successful build was August 6th 2020

evidolob (approved):
We do not use this `Jenkinsfile` anymore.

evidolob:
> Not as far as I know, unless @evidolob added them to a different jenkins instance. I think the only reason we had the Jenkinsfile was to build a tar for Che 6. The last successful build was August 6th 2020

The only `Jenkinsfile` which we use is on `vscode-yaml`, we use it for release

joshuawilson (approved):

PR #442, opened 2021-04-19
Initial work on adopting new parser

gorkem:
Replacing the parser with eemeli yaml parser AST to internal AST.


src/languageservice/parser/custom-tag-provider.ts
```diff
@@ -0,0 +1,65 @@
+import { Tags, isSeq, isMap, YAMLMap, YAMLSeq } from 'yaml'
+import { filterInvalidCustomTags } from '../utils/arrUtils';
+
+class CommonTagImpl {
+  public tag: string;
+  public readonly type: string;
+  default: never;
```
joshuawilson, line 7: `public` is the default. If you left it off to make it private it will not work. It would be better to be consistent and use public on all public fields or none of them.

Danielguo99 (approved):

evidolob:
@gorkem @JPinkney I think we are ready to merge this.
Could you do final review/test?

gorkem:
I can not really approve this PR since I am the author but please go ahead and merge.

PR #450, opened 2021-04-27
Handle nested anchors in validation expansion (Fixes #273)

### LINKED ISSUE #273: Anchor on property which uses alias fails validation

metcalf:
### What does this PR do?
Primarily, this adds support for recursively expanding multiple levels of aliases. In the process I made some other minor changes:
* `validateTextDocument` was swallowing errors which made my test pass when it should have failed. I think propagating errors could result in "unhandled promise rejection" errors in logs but I think that's better than ignoring them?
* Double-negative typos in a few tests I saw
* I added a test for a very particular case of anchor error rendering. I initially thought this was broken, but it appears intentional so I wanted to encode it in a test since it seems easy to regress.

#273

### Is it tested? How?
Wrote a test. Ran this in development against a YAML file/schema that was giving me trouble.

### DIFF
--- a/src/languageserver/handlers/validationHandlers.ts
+++ b/src/languageserver/handlers/validationHandlers.ts
@@ -51,24 +51,19 @@ export class ValidationHandler {
 
     return this.languageService
       .doValidation(textDocument, isKubernetesAssociatedDocument(textDocument, this.yamlSettings.specificValidatorPaths))
-      .then(
-        (diagnosticResults) => {
-          const diagnostics = [];
-          for (const diagnosticItem in diagnosticResults) {
-            diagnosticResults[diagnosticItem].severity = 1; //Convert all warnings to errors
-            diagnostics.push(diagnosticResults[diagnosticItem]);
-          }
-
-          const removeDuplicatesDiagnostics = removeDuplicatesObj(diagnostics);
-          this.connection.sendDiagnostics({
-            uri: textDocument.uri,
-            diagnostics: removeDuplicatesDiagnostics,
-          });
-          return removeDuplicatesDiagnostics;
-        },
-        () => {
-          return [];
+      .then((diagnosticResults) => {
+        const diagnostics = [];
+        for (const diagnosticItem in diagnosticResults) {
+          diagnosticResults[diagnosticItem].severity = 1; //Convert all warnings to errors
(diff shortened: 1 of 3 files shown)


src/languageserver/handlers/validationHandlers.ts
```diff
@@ -51,24 +51,19 @@ export class ValidationHandler {
...
-          const removeDuplicatesDiagnostics = removeDuplicatesObj(diagnostics);
-          this.connection.sendDiagnostics({
-            uri: textDocument.uri,
-            diagnostics: removeDuplicatesDiagnostics,
-          });
-          return removeDuplicatesDiagnostics;
-        },
-        () => {
-          return [];
+      .then((diagnosticResults) => {
```
metcalf, line 54: Review with `?w=1` if you want to confirm no semantic changes other than not swallowing the promise rejection.

evidolob (approved):
Working good. Not sure about swallowing promise rejection in `validateTextDocument`, maybe @JPinkney has some points.

JPinkney (approved):
I think we were swallowing errors because i'm not sure how other language clients respond to being sent an "unhandled promise rejection". vscode-json-languageservice seems to not swallow them so I'm OK with the change

PR #451, opened 2021-04-28
Fix range convert to offset+length

evidolob:
### What does this PR do?
Fix build.
Fix new parser AST range convert to offset + length.
This is fix all Document Symbols tests.

>Note: This PR changed convert of `Alias` node, nowit converted to `StringASTNodeImpl` instead of resolving to declaration node. As JSON AST don't support such nodes, Document Symbols calculation is breaking LSP rules on `SelectionRange`

### DIFF
--- a/src/languageservice/parser/ast-converter.ts
+++ b/src/languageservice/parser/ast-converter.ts
@@ -1,4 +1,19 @@
-import { Node, isScalar, Scalar, isMap, YAMLMap, isPair, Pair, isSeq, YAMLSeq, isNode, isAlias, Alias, Document } from 'yaml';
+import {
+  Node,
+  isScalar,
+  Scalar,
+  isMap,
+  YAMLMap,
+  isPair,
+  Pair,
+  isSeq,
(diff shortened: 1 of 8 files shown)


src/languageservice/parser/ast-converter.ts
```diff
@@ -1,4 +1,19 @@
...
+  YAMLMap,
+  isPair,
+  Pair,
+  isSeq,
+  YAMLSeq,
+  isNode,
+  isAlias,
+  Alias,
+  Document,
+  LineCounter,
```
joshuawilson, line 15: If you are going to the trouble of putting them each on a line, it might be worth alphabetizing it. 
It's up to you.

evidolob: That is `prettier` does, as line with import becomes to long

joshuawilson (approved):

JPinkney (approved):

gorkem (approved):

PR #452, opened 2021-04-28
Feat: Add config parameter to disable additional properties

p-spacek:
### What does this PR do?
Add `disableAdditionalProperties` into yaml configuration and if is set to true yaml validation won’t allow extra props that are not in schema (where type == 'object').

By default we can achieve same behave with setting each of schema object by property `additionalProperties = false`.
But this global `disableAdditionalParameters` set this globally for all object schemas.

For my purpose I'm using this to quickly switch between strict validation mode and not-strict.

no ref

### Is it tested? How?
added tests


evidolob:
@p-spacek Could you provide PR for vscode-yaml with adding settings for `disableAdditionalProperties` ?

p-spacek:
Created PR in vscode-yaml repo
 - just added this property into package contributes configuration.

evidolob (approved):

JPinkney (approved):

evidolob:
@p-spacek This PR has conflict, could you resolve it?

PR #459, opened 2021-05-12
feat: add option to specify custom schema store

catthehacker:
### What does this PR do?
Adds option to specify custom schema store schema

https://github.com/redhat-developer/vscode-yaml/issues/294
https://github.com/redhat-developer/vscode-yaml/issues/406

### Is it tested? How?

This change is published here: https://www.npmjs.com/package/@hackercat/yaml-language-server (with few changes in tests and my own schema store mirror) since I've tested it few times with custom store and clearing cache.
I think that the change is fairly small and easy to review. I'd appreciate any pointers since I never had any experience (except JS) with NodeJS/TS/NPM/etc. before and this whole thing was hacked in few hours (90% of time trying to figure out how to build this whole thing with `vscode-yaml` extension, it was a lot of `yarn this`, `yarn that`, I've not run `yarn test` for `vscode-yaml` extension since `vscode` breaks on my Alpine machine with `musl` instead of `glibc`).

### DIFF
--- a/src/yamlSettings.ts
+++ b/src/yamlSettings.ts
@@ -4,6 +4,7 @@ import { ISchemaAssociations } from './requestTypes';
 import { URI } from 'vscode-uri';
 import { JSONSchema } from './languageservice/jsonSchema';
 import { TextDocument } from 'vscode-languageserver-textdocument';
+import { JSON_SCHEMASTORE_URL } from './languageservice/utils/schemaUrls';
 
 // Client settings interface to grab settings relevant for the language server
 export interface Settings {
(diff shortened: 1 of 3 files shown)


src/yamlSettings.ts
```diff
@@ -54,6 +55,7 @@ export class SettingsState {
   schemaStoreSettings = [];
   customTags = [];
   schemaStoreEnabled = true;
+  schemaStoreUrl = 'https://www.schemastore.org/api/json/catalog.json';
```
evidolob, line 58: I think it worth to use `JSON_SCHEMASTORE_URL` constant to have single place where that URL defined.

evidolob (approved):
Sorry for long response. 
Could you also preference in `vscode-yaml`  for schema store url?

src/yamlSettings.ts
catthehacker: Fixed!

JPinkney (approved):

joshuawilson (approved):

PR #462, opened 2021-05-14
Trigger autocomplete event

kamok:
### What does this PR do?
Tries to trigger the event in vscode to automatically show the autocomplete modal.

Closes https://github.com/redhat-developer/vscode-yaml/issues/495

### Is it tested? How?
No, I don't know how to run this. I tried following the instructions on the repo and failed.

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -31,6 +31,7 @@ import { guessIndentation } from '../utils/indentationGuesser';
 import { TextBuffer } from '../utils/textBuffer';
 import { setKubernetesParserOption } from '../parser/isKubernetes';
 import { ClientCapabilities, MarkupContent } from 'vscode-languageserver';
+import { endsWith } from '../utils/strings';
 const localize = nls.loadMessageBundle();
 
 const doubleQuotesEscapeRegExp = /[\\]+"/g;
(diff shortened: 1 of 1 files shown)


evidolob:
@kamok could you fix build?

kamok:
I will fix it.

kamok:
How do you want to fix these tests? For example, there is a test that tests for "should use markdownDescription for property completion"

Here's the expectation:
```
expect(completion.items[0]).eql(
        createExpectedCompletion('kind', 'kind: $1', 0, 0, 0, 3, 10, InsertTextFormat.Snippet, {
          documentation: {
            kind: MarkupKind.Markdown,
            value: '**kind** (string)\n\nKind is a string value representing the REST resource this object represents.',
          },
        })
      );
```

With my change, it would also return a `command` to trigger autocomplete like the json library. So the new expectation would be:

```
expect(completion.items[0]).eql(
        createExpectedCompletion('kind', 'kind: $1', 0, 0, 0, 3, 10, InsertTextFormat.Snippet, {
          documentation: {
            kind: MarkupKind.Markdown,
            value: '**kind** (string)\n\nKind is a string value representing the REST resource this object represents.',
          },
          command: {
            command: "editor.action.triggerSuggest",
            title: "Suggest"
          }
        })
      );
```

This is not optimal since it would seem weird to expect something like `command: "editor.action.triggerSuggest",` in a test about `documentation`. 

In other test frameworks, there's something like testing if an object contains certain key / value, which would fit this problem without coding in the triggerSuggest command in unit tests that are unrelated. 

https://stackoverflow.com/questions/35101998/chai-assertion-testing-whether-object-structure-contains-at-least-other-object-s

Do you guys want to pull in a new library and tweak these test that way? Is there another convention your team prefers?

kamok:
@JPinkney, @evidolob, any thoughts on my question above?

JPinkney:
I'm personally a fan of just checking the object explicitly, without checking a subset. That way when we are checking a specific completion request we know it doesn't have any extra fields we aren't expecting.

@evidolob WDYT?

evidolob:
I depends, but mostly I prefer checking the object explicitly.

PR #505, opened 2021-07-09
Fix inline schema with url handling

evidolob:
### What does this PR do?
Fixes inline schema URL handling.

Fix: https://github.com/redhat-developer/vscode-yaml/issues/550

### Is it tested? How?
with test

### DIFF
--- a/test/autoCompletion.test.ts
+++ b/test/autoCompletion.test.ts
@@ -1591,20 +1591,6 @@ describe('Auto Completion Tests', () => {
       });
       assert.strictEqual(result.items.length, 3, `Expecting 3 items in completion but found ${result.items.length}`);
     });
-
-    it('should handle http(s) url', async () => {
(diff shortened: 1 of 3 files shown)


test/autoCompletion.test.ts
```diff
@@ -1591,6 +1591,20 @@ describe('Auto Completion Tests', () => {
       });
       assert.strictEqual(result.items.length, 3, `Expecting 3 items in completion but found ${result.items.length}`);
     });
+
+    it('should handle http(s) url', async () => {
+      const documentContent = `# yaml-language-server: $schema=http://json-schema.org/draft-07/schema# anothermodeline=value\n`;
```
JPinkney, line 1596: I assume this test is reaching out to the internet to grab the schema? If so, is it possible to mock the return of the schema so that we don't have to touch the internet

JPinkney (approved):

PR #507, opened 2021-07-13
Fix more tests

evidolob:
### What does this PR do?
Fix almost all tests, except one - `Snippet in array schema should autocomplete correctly on array level`

The problem with last test, is that, during codecompletion we modify original yaml, by adding `holder:\n` in cursor position(see https://github.com/redhat-developer/yaml-language-server/blob/main/src/languageservice/services/yamlCompletion.ts#L1026 ),
and current parser has a bug, it generate wrong ast node for `holder`(node has no text and wrong position)  and we calculate completion items from that, not correct AST tree. New parser doesn’t  have such bug, and AST differs.

### DIFF
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -82,21 +82,30 @@ export interface IProblem {
 }
 
 export abstract class ASTNodeImpl {
-  public abstract readonly type: 'object' | 'property' | 'array' | 'number' | 'boolean' | 'null' | 'string';
+  abstract readonly type: 'object' | 'property' | 'array' | 'number' | 'boolean' | 'null' | 'string';
 
-  public offset: number;
-  public length: number;
-  public readonly parent: ASTNode;
-  public location: string;
+  offset: number;
+  length: number;
+  readonly parent: ASTNode;
+  location: string;
 
   constructor(parent: ASTNode, offset: number, length?: number) {
     this.offset = offset;
     this.length = length;
     this.parent = parent;
+    this.updateParentPosition();
   }
 
-  public getNodeFromOffsetEndInclusive(offset: number): ASTNode {
-    const collector = [];
+  private updateParentPosition(): void {
+    if (this.parent && this.parent.offset + this.parent.length < this.offset + this.length) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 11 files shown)


JPinkney (approved):

src/languageservice/parser/jsonParser07.ts
```diff
@@ -82,21 +82,30 @@ export interface IProblem {
...
+  offset: number;
+  length: number;
+  readonly parent: ASTNode;
+  location: string;
 
   constructor(parent: ASTNode, offset: number, length?: number) {
     this.offset = offset;
     this.length = length;
     this.parent = parent;
+    this.updateParentPosition();
```
gorkem, line 96: I am not thrilled that a node needs to update its parent's position. Shouldn't this be the responsibility of the AST converter logic?

src/languageservice/parser/yaml-documents.ts
```diff
@@ -24,10 +24,12 @@ export class SingleYAMLDocument extends JSONDocument {
   public root: ASTNode;
   public currentDocIndex: number;
   private _lineComments: string[];
+  private text: string;
 
-  constructor(lineCounter?: LineCounter) {
+  constructor(text: string, lineCounter?: LineCounter) {
```
gorkem, line 29: We need to document what text is about.

gorkem (changes requested):
Should we make a plan to remove that holder logic? I am not sure, I fully grok we we are doing the reparsing?

src/languageservice/parser/jsonParser07.ts
evidolob: It should be, but to be compatible with previous AST, we need recalculate parent position, as some child may be have length bigger that new parser provide, and we extend  length in some cases there: https://github.com/redhat-developer/yaml-language-server/pull/507/files#diff-c61722cb0cd0930a1a34772189f9726eca5a9e95b69e23a6b07acfc40d219d39R166

evidolob:
Yes, whole https://github.com/redhat-developer/yaml-language-server/blob/main/src/languageservice/services/yamlCompletion.ts#L983 looks redundant, but when I try to delete it, a lot of codecomplation test starts failing, so it is necessary.

The BIG problem that adding `holder:\n` is relaying on bug in previous parser. 
I will start looking on how we can remove `completionHelper` completely. But, I'm afraid that will requires a full rewriting of `YAMLCompletion`  class.

PR #511, opened 2021-07-15
Fix fetching nested http settings

robb-j:
### What does this PR do?

This PR tweaks the way the `http` configuration section is pulled from the client. It pulls the entire `http` section at once then plucks out the values it needs. I do assume there is no sensitive client information other that `proxy` and `proxyStrictSSL` on the `http` section which shouldn't be pulled.

In my research I've found no solid definition for what a `section` in the LSP is so my thinking is that usage of it should be as simple as possible for most clients to support it.

[Nova](https://nova.app) doesn't currently have a way to configure how workspace settings are passed to the server and it does it in a bit of an odd way at the moment, which I've described in the issue below. Requesting `http.proxy` in Nova returns `{}` which was getting passed straight to `requests_light` and breaking it.

- https://github.com/robb-j/nova-yaml/issues/20

### Is it tested? How?

With unit tests. I updated the existing unit test to follow the new LSP message structure and added an assertion for the shape of settings passed to `setConfiguration`.

I ran the modified server in Nova using `npm link` and confirmed that schemas were being loaded and tooltips were showing again.

I can try running it with VSCode if you like but I don't currently have an extension development setup for it.

### DIFF
--- a/src/languageserver/handlers/settingsHandlers.ts
+++ b/src/languageserver/handlers/settingsHandlers.ts
@@ -34,17 +34,16 @@ export class SettingsHandler {
   async pullConfiguration(): Promise<void> {
     const config = await this.connection.workspace.getConfiguration([
       { section: 'yaml' },
-      { section: 'http.proxy' },
-      { section: 'http.proxyStrictSSL' },
+      { section: 'http' },
       { section: '[yaml]' },
     ]);
     const settings: Settings = {
       yaml: config[0],
       http: {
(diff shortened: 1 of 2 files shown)


evidolob:
@robb-j Could you fix build?

robb-j:
That's fixed, it was a code formatting issue in the test.

evidolob (approved):

JPinkney (approved):

PR #512, opened 2021-07-15
Add support of URL fragments in schema URL

evidolob:
### What does this PR do?
It add support of referencing to `#/definitions/{SomeObjectInSchema}`  in JSON Schema URL.
Before `0.21.0` we have support of such URL, but that support was because of bug in schema priority code.
In  `0.21.0` that bug was fixed and we loose of support. This PR add proper implementation of referencing in schema URL.

Resolve: https://github.com/redhat-developer/vscode-yaml/issues/557

### Is it tested? How?
With test. 
To test manually reproduce steps from https://github.com/redhat-developer/vscode-yaml/issues/557

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -24,7 +24,6 @@ import { SingleYAMLDocument } from '../parser/yamlParser07';
 import { JSONDocument } from '../parser/jsonParser07';
 import { load } from 'js-yaml';
 import * as path from 'path';
-import { settings } from 'cluster';
 
 const localize = nls.loadMessageBundle();
 
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -24,6 +24,7 @@ import { SingleYAMLDocument } from '../parser/yamlParser07';
 import { JSONDocument } from '../parser/jsonParser07';
 import { load } from 'js-yaml';
 import * as path from 'path';
+import { settings } from 'cluster';
 
 const localize = nls.loadMessageBundle();
 
@@ -256,6 +257,13 @@ export class YAMLSchemaService extends JSONSchemaService {
         collectMapEntries(next.definitions, next.properties, next.patternProperties, <JSONSchemaMap>next.dependencies);
         collectArrayEntries(next.anyOf, next.allOf, next.oneOf, <JSONSchema[]>next.items, next.schemaSequence);
       };
+
+      if (parentSchemaURL.indexOf('#') > 0) {
+        const segments = parentSchemaURL.split('#', 2);
+        if (segments[0].length > 0 && segments[1].length > 0) {
+          openPromises.push(resolveExternalLink(node, segments[0], segments[1], parentSchemaURL, parentSchemaDependencies));
+        }
+      }
 
       while (toWalk.length) {
         const next = toWalk.pop();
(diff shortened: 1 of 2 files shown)


src/languageservice/services/yamlSchemaService.ts
```diff
@@ -24,6 +24,7 @@ import { SingleYAMLDocument } from '../parser/yamlParser07';
 import { JSONDocument } from '../parser/jsonParser07';
 import { load } from 'js-yaml';
 import * as path from 'path';
+import { settings } from 'cluster';
```
JPinkney, line 27: Looks like this isn't used anywhere

JPinkney (approved):

remcohaszing:
When loading a JSON schema from JSON Schema Store that contains a URL fragment, that’s also broken, even in vscode-yaml version 0.20.0. I was the first to introduce such an entry to the catalog in SchemaStore/schemastore#1749. I’m not sure if that’s also covered by this PR or if I should create a new issue.

evidolob:
@remcohaszing It isn't broken, we newer has such feature. 
Previously it can work because of bug.

I'm going to support your schema in this PR.

evidolob:
@remcohaszing I found one bug in your schema:
you define:
```json
"SecurityRoleDefinition": {
        "type": "object",
        "properties": {
          "description": {
            "type": "string",
            "description": "The description of the role."
          },
          "defaultPage": {
            "$ref": "#/components/schemas/App/properties/definition/properties/defaultPage", // <- this path doesn't exists 
            "description": "The default page to redirect users with this role to."
          },
          "inherits": {
            "type": "array",
            "minItems": 1,
            "description": "The name of the role to inherit from.\n\nNote that this role must exist and can not inherit itself via this field or the `inherits` field\nof the referenced role.\n",
            "items": {
              "type": "string"
            }
          }
        }
      },
   ```
   The `$ref` : `"#/components/schemas/App/properties/definition/properties/defaultPage"` should be: "#/components/schemas/AppDefinition/properties/defaultPage" or something lake that, I'm not completely sure.
But `#/components/schemas/App/properties/definition/properties/defaultPage` path doesn't exist.

If you fix your schema, all references should work with this PR.

**Comment by remcohaszing** [comment]
Oh, super weird our tests didn’t catch that. :thinking: 

Thanks for pointing that out, I’m definitely going to fix that right now!

**Comment by remcohaszing** [comment]
It works great now. Thanks @evidolob!

**Comment by mgabeler-lee-6rs** [comment]
@evidolob Sadly it's not working for me :(

```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/schemas/v3.0/schema.json#/definitions/Schema
type: object
```
produces errors:
image

evidolob:
@mgabeler-lee-6rs But it works for me:
Screenshot 2021-07-29 at 17 38 22

remcohaszing:
Unfortunately I have to confirm I see the exact same issue as @mgabeler-lee-6rs, although schema validation works fine for `https://appsemble.app/api.json#/components/schemas/AppDefinition`.

The only difference I can think of is that the top level object of `https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/schemas/v3.0/schema.json` is a JSON schema and the top level object of `https://appsemble.app/api.json` is not.

mgabeler-lee-6rs:
It seems there may be something super-weird going on -- if I downgrade the vscode extension to 0.20.0 again (which is where it was working before), it's showing the same error, where it was working fine before I upgraded to 0.22.0 a couple hours ago.

Could there be a stray (wrong) version of the language server running somewhere, or not getting updated properly?

mgabeler-lee-6rs:
commenting on closed issues is usually pretty pointless, so I've opened #522 to follow up on this

PR #534, opened 2021-08-25
Fix error handling in hover and codelens

evidolob:
### What does this PR do?
It fix possible issue in `recursivelyBuildAst` which can cause https://github.com/redhat-developer/vscode-yaml/issues/583 and https://github.com/redhat-developer/vscode-yaml/issues/584

Also this PR enhance error handling in `hover` and `codelens` implementations.

Fix https://github.com/redhat-developer/vscode-yaml/issues/583 
Fix https://github.com/redhat-developer/vscode-yaml/issues/584

### Is it tested? How?
No, as currently telemetry doesn't have enough information about errors.


gorkem (approved):
lgtm

PR #536, opened 2021-09-02
Treat OpenAPI descriptions as markdown

remcohaszing:
### What does this PR do?

This treats descriptions in OpenAPI documents as markdown as per
https://spec.openapis.org/oas/latest.html#rich-text-formatting

It fixes a specific, yet common use case for https://github.com/redhat-developer/vscode-yaml/issues/454

### Is it tested? How?

A test was added.

### DIFF
--- a/src/languageservice/services/yamlHover.ts
+++ b/src/languageservice/services/yamlHover.ts
@@ -101,13 +101,17 @@ export class YAMLHover {
         matchingSchemas.every((s) => {
           if (s.node === node && !s.inverted && s.schema) {
             title = title || s.schema.title;
-            markdownDescription = markdownDescription || s.schema.markdownDescription || toMarkdown(s.schema.description);
+            markdownDescription =
+              markdownDescription ||
+              s.schema.markdownDescription ||
+              ('openapi' in schema.schema ? s.schema.description : toMarkdown(s.schema.description));
             if (s.schema.enum) {
               const idx = s.schema.enum.indexOf(getNodeValue(node));
               if (s.schema.markdownEnumDescriptions) {
                 markdownEnumValueDescription = s.schema.markdownEnumDescriptions[idx];
               } else if (s.schema.enumDescriptions) {
-                markdownEnumValueDescription = toMarkdown(s.schema.enumDescriptions[idx]);
+                markdownEnumValueDescription =
+                  'openapi' in schema.schema ? s.schema.description : toMarkdown(s.schema.enumDescriptions[idx]);
               }
               if (markdownEnumValueDescription) {
                 enumValue = s.schema.enum[idx];
--- a/test/hover.test.ts
+++ b/test/hover.test.ts
@@ -336,6 +336,29 @@ describe('Hover Tests', () => {
       );
     });
 
+    it('Hover treats descriptions as markdown for OpenAPI documents', async () => {
+      languageService.addSchema(SCHEMA_ID, {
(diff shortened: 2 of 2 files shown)


remcohaszing:
Would you also accept if I add a property to schema settings to force enable markdown descriptions? I see now that this is very easy.

```diff
  export interface SchemasSettings {
    priority?: SchemaPriority;document then highest priority wins
    fileMatch: string[];
+   markdownDescriptions?: boolean
    schema?: unknown;
    uri: string;
  }
```

```diff
        if (schema && node && !schema.errors.length) {
-         const descriptionsAreMarkdown = 'openapi' in schema.schema || 'swagger' in schema.schema;
+         const descriptionsAreMarkdown = schema.markdownDescriptions || 'openapi' in schema.schema || 'swagger' in schema.schema;
          const matchingSchemas = doc.getMatchingSchemas(schema.schema, node.offset);
```

This would make it possible to specify JSON schema descriptions are markdown in `monaco-yaml`. Also although it wouldn’t resolve the referenced issue yet, but it would be a first step towards a fix.

gorkem:
Can we take a different approach and make `toMarkdown` function more clever so that it does not make the changes that it applies when it is not necessary?

remcohaszing:
`toMarkdown` could simply take `descriptionsAreMarkdown` as a second argument, and return the raw input if it’s this property is true. Although IMO this doesn’t make the code simpler or more readable.

gorkem:
I was actually thinking if we could make `toMarkdown` to detect if the string is already Markdown and skip modifying it. Unfortunately, I was not successfully spot an easy way to do that with my glorious 3 min spend on Google :)

src/languageservice/services/yamlHover.ts
```diff
@@ -92,6 +92,7 @@ export class YAMLHover {
 
     return this.schemaService.getSchemaForResource(document.uri, doc).then((schema) => {
       if (schema && node && !schema.errors.length) {
+        const descriptionsAreMarkdown = 'openapi' in schema.schema || 'swagger' in schema.schema;
```
gorkem, line 95: Move this to its own function so that it is easier for us to spot where to add new schemas in the future

gorkem (changes requested):

remcohaszing:
> I was actually thinking if we could make `toMarkdown` to detect if the string is already Markdown and skip modifying it. Unfortunately, I was not successful to spot an easy way to do that with my glorious 3 min spend on Google :)

I’m pretty sure it’s not. :disappointed: It would have been awesome though.

The following clearly defines a continuous paragraph in markdown

```md
Penguins are awesome
birds.
```

However, in plain text the newline makes it ambiguous.

I have also been thinking maybe all descriptions could be treated as markdown, but that would mess up code sample for example.

gorkem:
I am a bit reluctant on this one because we are introducing a behavior based on schema. We did that in the early times with kubernetes and it has been haunting us since. We are actually planning to remove it. Can we think of a way to introduce this without yaml-language-server being aware of the schema?

remcohaszing:
The alternative is what I proposed in https://github.com/redhat-developer/yaml-language-server/pull/536#issuecomment-912499323. This makes the hover provider only aware of whether or not it should treat descriptions as markdown, but not why. This moves the responsibility of determining whether or not a schema has markdown descriptions up.

gorkem:
I am more comfortable with that one, instead of yaml-language-server being aware of the schemas. I am afraid the list of schemas may grow out of hand quickly.

PR #566, opened 2021-10-26
Provide completion for schema ids in modeline #560

### LINKED ISSUE #560: Provide completion for reachable schema

When getting started on a file which doesn't have a schema already associated based on name or another provider, it is not easy to know where to search a schema and what is the url for it.
see also https://github.com/redhat-developer/vscode-yaml/issues/567

it would nice to have a list of potential schemas from schemastore (somewhere else too? local?)
So when I call completion after `# yaml-language-server: $schema=`, there is a list of schema proposed.

apupier:
potential improvements:
- support range when there is already a value provided
- provide completion for local yaml schema files

Extracted a specific file for modeline related things.

### What does this PR do?

Provide completion for schema ids in modeline. So when we have `# yaml-language-server: $schema=`

fixes #560

### Is it tested? How?

i wasn't abel to test the yaml-language-server

### DIFF
--- a/src/languageservice/services/modelineUtil.ts
+++ b/src/languageservice/services/modelineUtil.ts
@@ -1,0 +1,37 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Red Hat, Inc. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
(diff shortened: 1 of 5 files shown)


apupier:
betterly handled by https://github.com/redhat-developer/yaml-language-server/pull/573

PR #573, opened 2021-11-02
Completion schema URI in modeline comment

evidolob:
### What does this PR do?
Completion schema URI in modeline comment, based on https://github.com/redhat-developer/yaml-language-server/pull/566

Screenshot 2021-11-02 at 12 03 13

none

### Is it tested? How?
with tests


evidolob:
To get this working, I need to change https://github.com/redhat-developer/vscode-yaml/blob/main/language-configuration.json#L37 `wordPattern` regexp, as currently it copied from JSON and not valid for yaml

evidolob:
As https://github.com/redhat-developer/vscode-yaml/pull/634 merged, this PR should work.

@JPinkney @apupier Could you review this?

apupier (approved):
Despite I was not able to test locally (for general reasons, not specific to this PR), reading code and seeing the screenshot leads me to think that it is a good PR.

JPinkney (approved):

PR #574, opened 2021-11-03
implemented the squeceItem identation based on the sequenceItemIndent…

msivasubramaniaan:
… value

### What does this PR do?

Doing indentation on sequeceItem based on the settings. If sequenceItemIndent is 0 then the sequenceItem not perform any indentation and it acquire it parent value itself. Suppose sequenceItemIndent contains any values, it will do the indentation on sequenceItem.

- When sequenceItemIndent is 0
image

- When sequenceItemIndent is 2
image

https://github.com/redhat-developer/vscode-yaml/issues/172

### Is it tested? How?
Yes. Using test cases

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ The following settings are supported:
 - `yaml.format.bracketSpacing`: Print spaces between brackets in objects
 - `yaml.format.proseWrap`: Always: wrap prose if it exeeds the print width, Never: never wrap the prose, Preserve: wrap prose as-is
 - `yaml.format.printWidth`: Specify the line length that the printer will wrap on
+- `yaml.format.sequenceItemIndent`: Specify the indent size of the sequence item from it parent
 - `yaml.validate`: Enable/disable validation feature
 - `yaml.hover`: Enable/disable hover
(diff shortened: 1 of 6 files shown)


evidolob (changes requested):
@msivasubramaniaan It doesn't work with:
```yaml
objA:
  - name: nameA1
 ```

and :
```yaml
objC:
  - { foo: bar }
  - { foo: bar2 }
 ```

evidolob:
And I'm not very happy with having own code which do formatting after executing formatter tool. I think it would be better if we change formatter tool or contribute to them new features.
@JPinkney @gorkem WDYT?

JPinkney:
I'd prefer not to own any more of the formatter stuff then we have to. With that said, prettier is very opinionated and probably wouldn't allow us to contribute this feature. Maybe we can switch to [prettier-x](https://github.com/brodybits/prettierx) which is a less opinionated version of prettier and can contribute this feature there

gorkem:
I agree, having to split the formatting logic between the extension and the upstream library is going to cause us more long term issues.

PR #589, opened 2021-11-18
Fix collecting comments from AST

evidolob:
### What does this PR do?
Fix collecting comments, in case when document contains several commented line in a row, like:
```yaml
#first comment
#second comment
foo: bar
```

Fix: https://github.com/redhat-developer/vscode-yaml/issues/629

### Is it tested? How?
With test


gorkem (approved):
LGTM

JPinkney (approved):

Issue #524, opened 2021-08-19
Consider extracting yaml-language-service from yaml-language-server


remcohaszing:
## Summary

I would like to be able to depend on the language service part without pulling in the language server part and all of its dependencies.

## Relevant information

I’m working on [`monaco-yaml`](https://github.com/remcohaszing/monaco-yaml). `monaco-yaml` depends on the YAML language service part in this package, but not on the language server. Currently a build process is needed to include the language service, but ideally I would like to be able to simply add it as a dependency without pulling in the language server part and its dependencies.

I believe extracting the language service also helps to better separate the two parts. `monaco-yaml` still depends on an older version of `yaml-language-server`, but I noticed the latest version of the language service also uses some imports from the language server.

I understand if this gets rejected, because it adds an additional maintenance burden, but I wanted to ask anyway to see what you think of it. I think using a yarn mono repository reduces the additional maintenance a lot.

evidolob:
@remcohaszing I'm not clear, what language service do you want to extract? Do you mean [`languageservice`](https://github.com/redhat-developer/yaml-language-server/tree/main/src/languageservice) folder?

remcohaszing:
Yes. I would like the [`languageservice`](https://github.com/redhat-developer/yaml-language-server/tree/main/src/languageservice) folder to be published as a separate package. I imagine it would be called `yaml-language-service`. The `yaml-language-server` package would then depend on `yaml-language-service`.

This is similar to [`vscode-json-languageserver`](https://www.npmjs.com/package/vscode-json-languageserver) which depends on [`vscode-json-languageservice`](https://www.npmjs.com/package/vscode-json-languageservice)

fantonangeli:
Hi all,
We also have a similar issue in [KIE-Tools](https://github.com/kiegroup/kie-tools). 
Our package [Serverless Logic Web Tools](https://github.com/kiegroup/kie-tools/tree/main/packages/serverless-logic-sandbox), which is a  web-app, we needed to validate a YAML using a schema. 
Using `yaml-language-server` from the Jest everything works, but Webpack cannot compile and we have errors with importing libraries, mainly `prettier`.
I tried importing `yaml-language-server` using CJS, ESM or UMD formats but it didn't work.

To solve the issue I bundled `yaml-language-server` with Esbuild, like @remcohaszing  in `monaco-yaml`, in a different package and used that package in Webpack.
https://github.com/kiegroup/kie-tools/tree/main/packages/yaml-language-server
[...]

PR #593, opened 2021-11-22
Prefer vscode-languageserver-types imports

remcohaszing:
### What does this PR do?

Members from `vscode-languageserver-types` are re-exported through `vscode-languageserver-protocol`. The members of
`vscode-languageserver-protocol` are re-exported through `vscode-languageserver`.

This changes all of these imports to be imported from their base package. This saves the need to rewrite imports for packages which only depend on the language service part ([monaco-yaml](https://github.com/remcohaszing/monaco-yaml/blob/934eb8d1e302d2e2fffa117b3d0e1b78cd42baf7/build.js#L23-L28)).

Alternatively all of these imports could come from `vscode-languageserver`. In that case the explicit dependencies `vscode-languageserver-types` and `vscode-languageserver-textdocument` can be removed from `package.json`.

Ideally if #524 is implemented, the language service would depend on `vscode-languageserver-types` and `vscode-languageserver-protocol` and the language server would import on `vscode-languageserver`.

### Is it tested? How?

```sh
yarn build
```


src/languageservice/services/yamlValidation.ts
```diff
@@ -5,7 +5,7 @@
  *--------------------------------------------------------------------------------------------*/
 'use strict';
 
-import { Diagnostic, Position } from 'vscode-languageserver';
+import { Diagnostic, Position } from 'vscode-languageserver-types';
 import { LanguageSettings } from '../yamlLanguageService';
 import { YAMLDocument, YamlVersion } from '../parser/yamlParser07';
 import { SingleYAMLDocument } from '../parser/yamlParser07';
```
gorkem, line 11: Not really related to this PR but yamlParser07 is imported multiple times.

gorkem (approved):
I wish there was an eslint rule for us to enforce this in the future otherwise LGTM.

remcohaszing:
> I wish there was an eslint rule for us to enforce this in the future otherwise LGTM.

This (actually the exact opposite) is possible using ([import/no-extraneous-dependencies](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md))!

If `vscode-languageserver-textdocument` and `vscode-languageserver-types` are removed from `package.json`, this rule can enforce all rules are imported from `vscode-languageserver`.

If the language service is extracted into its own package which depends on `vscode-languageserver-textdocument` and `vscode-languageserver-types`, this rule can be used to disallow imports from `vscode-languageserver` in the language service part.

src/languageservice/services/yamlValidation.ts
remcohaszing: This could be linted using ESLint rule [import/no-duplicates](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md). Actually this plugin could help maintain more consistency, such as import sorting.

ssbarnea:
Any chance to rebase this?

ssbarnea:
@remcohaszing Sorry to ping you again...

remcohaszing:
No problem! Rebased :smile:

PR #594, opened 2021-11-22
Fix document symbols, when key is not string

evidolob:
### What does this PR do?
Fix `yaml.hierarchicalDocumentSymbols.error` error message on telemetry.
That can be caused when mapping key is not string.

none

### Is it tested? How?
with tests.

### DIFF
--- a/src/languageservice/services/documentSymbols.ts
+++ b/src/languageservice/services/documentSymbols.ts
@@ -12,7 +12,7 @@ import { DocumentSymbolsContext } from 'vscode-json-languageservice/lib/umd/json
 import { TextDocument } from 'vscode-languageserver-textdocument';
 import { yamlDocumentsCache } from '../parser/yaml-documents';
 import { Telemetry } from '../../languageserver/telemetry';
-import { isMap, isScalar, isSeq } from 'yaml';
+import { isMap, isSeq, Node } from 'yaml';
 
 export class YAMLDocumentSymbols {
   private jsonDocumentSymbols;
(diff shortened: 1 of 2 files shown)


JPinkney (approved):

src/languageservice/services/documentSymbols.ts
```diff
@@ -12,22 +12,36 @@ import { DocumentSymbolsContext } from 'vscode-json-languageservice/lib/umd/json
...
-    const origKeyLabel = this.jsonDocumentSymbols.getKeyLabel;
 
     // override 'getKeyLabel' to handle complex mapping
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     this.jsonDocumentSymbols.getKeyLabel = (property: any) => {
-      if (typeof property.keyNode.value === 'object') {
-        return property.keyNode.value.value;
+      const keyNode = property.keyNode.internalNode;
+      let name = '';
+      if (isScalar(keyNode)) {
```
gorkem, line 28: Can we remove this first `if`? It is probably covered by else at the end.

src/languageservice/services/documentSymbols.ts
```diff
@@ -12,22 +12,36 @@ import { DocumentSymbolsContext } from 'vscode-json-languageservice/lib/umd/json
...
+        name = keyNode.source;
       }
+
+      if (name) {
+        name = name.replace(/[\n]/g, '↵');
+      }
+      if (name && name.trim()) {
+        return name;
+      }
+      return `"${name}"`;
```
gorkem, line 44: What is the case to reach this one? Can we have a test for it ?

gorkem (changes requested):

gorkem (approved):

Issue #583, opened 2021-11-10
LSP triggering`registerCapability` despite `dynamicRegistration` set to false


WhyNotHugo:
## Describe the bug

When using `yaml-language-server` I see this printed on my logs:

The language server yamlls triggers a registerCapability handler despite dynamicRegistration set to false. Report upstream.

The LSP exits soon after -- I've no idea if both issues are related, but this issue still does make sense.

## Expected Behavior

Should not triggers a registerCapability if dynamicRegistration set to false.

## Current Behavior

Triggers a registerCapability if dynamicRegistration set to false.

## Steps to Reproduce

No special steps merely configured nvim (using nvim-lspconfig) and opened a YAML file.

## Environment
- Linux

msivasubramaniaan:
@WhyNotHugo can you please share the **.vim** config file which you used. It would help us to reproduce the issue at the earliest.

PR #598, opened 2021-11-26
Fix client/registerCapability request when workspaces/didChangeWorksp…

msivasubramaniaan:
…aceFolder even if dynamicRegistration set to false

### What does this PR do?
It checked the client workspace dynamicRegistration true/false on didChangeWatchedFiles. If false, then disable the client/registerCapability call from server to client

https://github.com/redhat-developer/yaml-language-server/issues/583

### Is it tested? How?
Yes, tested with tests and with latest changes on neovim.

### DIFF
--- a/src/yamlSettings.ts
+++ b/src/yamlSettings.ts
@@ -84,7 +84,7 @@ export class SettingsState {
   hasConfigurationCapability = false;
   useVSCodeContentRequest = false;
   yamlVersion: YamlVersion = '1.2';
-  hasWsChangeWathedFileDynamicRegistraton = false;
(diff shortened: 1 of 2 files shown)


src/yamlSettings.ts
```diff
@@ -84,6 +84,7 @@ export class SettingsState {
   hasConfigurationCapability = false;
   useVSCodeContentRequest = false;
   yamlVersion: YamlVersion = '1.2';
+  hasWsChangeWathedFileDynamicRegistraton = false;
```
evidolob, line 87: It is better to rename to `hasWsChangeWatchedFileDynamicRegistration`

evidolob (approved):

PR #602, opened 2021-12-02
Improve telemetry errors

evidolob:
### What does this PR do?
It ensures that we send errors to telemetry with stack traces. 
Also it add `uncaughtException` error handler for  yaml-ls `process` to track errors and sand them with stack trace to telemetry. That should add possibility to detect `Maximum call stack size exceeded` errors.

None, but it help with issues like https://github.com/redhat-developer/vscode-yaml/issues/652 where we have only error message, but we don't have any trace.

### Is it tested? How?
manually


gorkem (changes requested):
It is missing test cases. I have added some on https://github.com/evidolob/yaml-language-server/pull/1 also handled a null check.

gorkem (approved):

JPinkney (approved):

PR #603, opened 2021-12-02
Fix relative file paths with fragments

codeclown:
### What does this PR do?
Fixes fragments when schema URL is a local file.

Example of such a reference:

```
# yaml-language-server: $schema=../../schemas/example.schema.json#/definitions/CustomSpec
```

Currently this doesn't work in VS Code:

image

Error on hover is:

```
Unable to load schema from '/path/to/schemas/example.schema.json#/definitions/CustomSpec': No content.YAML(768)
```

The issue is caused by "#" being turned into "%23" during some path resolve operations. This PR fixes it by excluding the fragment during those operations.

Couldn't find a GitHub issue.
Relates somewhat to old PR which added fragment support: #530

### Is it tested? How?
Includes automated test in existing suite.
I don't know how to test this with VS Code directly (if you could advice, I'll happily test locally; as a sidenote, CONTRIBUTING.md in this regard would be helpful).

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -328,11 +328,23 @@ export class YAMLSchemaService extends JSONSchemaService {
       let schemaFromModeline = getSchemaFromModeline(doc);
       if (schemaFromModeline !== undefined) {
         if (!schemaFromModeline.startsWith('file:') && !schemaFromModeline.startsWith('http')) {
+          // If path contains a fragment and it is left intact, "#" will be
+          // considered part of the filename and converted to "%23" by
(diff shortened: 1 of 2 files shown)


evidolob:
@codeclown Thx for contribution.
To test this in vscode you can follow https://github.com/redhat-developer/vscode-yaml/blob/main/CONTRIBUTING.md

codeclown:
@evidolob Thanks for the tip! I managed to test the change locally, and it works as expected.

Also added an if-else to the test to account for Windows paths.

evidolob (approved):

JPinkney (approved):

PR #606, opened 2021-12-07
feat: disable default props

p-spacek:
### What does this PR do?
Add config to exclude not required default props from completion.
fix: required prop with default value or with const

The reason why users appreciated this feature is that users don't have to specify default properties. Only if the default value needs to be changed, it makes sense to add them (but not all of them automatically).

no ref

### Is it tested? How?
add unit tests

### DIFF
--- a/test/autoCompletion.test.ts
+++ b/test/autoCompletion.test.ts
@@ -156,7 +156,7 @@ describe('Auto Completion Tests', () => {
             name: {
               type: 'string',
               // eslint-disable-next-line prettier/prettier, no-useless-escape
-              default: '\"yaml\"',
+              default: '"yaml"',
             },
           },
         });
@@ -178,7 +178,7 @@ describe('Auto Completion Tests', () => {
             name: {
               type: 'string',
               // eslint-disable-next-line prettier/prettier, no-useless-escape
-              default: '\"yaml\"',
+              default: '"yaml"',
             },
           },
         });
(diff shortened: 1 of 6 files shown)


evidolob:
@p-spacek It would be nice if you add description for new settings item in to [README.md](https://github.com/redhat-developer/yaml-language-server#language-server-settings)

test/autoCompletion.test.ts
```diff
@@ -386,6 +386,62 @@ describe('Auto Completion Tests', () => {
...
+      it('Autocomplete without default value - not required', async () => {
+        const languageSettingsSetup = new ServiceSetup().withCompletion();
+        languageSettingsSetup.languageSettings.disableDefaultProperties = true;
+        languageService.configure(languageSettingsSetup.languageSettings);
+        languageService.addSchema(SCHEMA_ID, {
+          type: 'object',
+          properties: {
+            scripts: {
+              type: 'object',
+              properties: {
```
gorkem, line 398: Can we possibly have a second non-string property added to this test case?

test/autoCompletion.test.ts
```diff
@@ -386,6 +386,62 @@ describe('Auto Completion Tests', () => {
...
+      it('Autocomplete without default value - required', async () => {
+        const languageSettingsSetup = new ServiceSetup().withCompletion();
+        languageSettingsSetup.languageSettings.disableDefaultProperties = true;
+        languageService.configure(languageSettingsSetup.languageSettings);
+        languageService.addSchema(SCHEMA_ID, {
+          type: 'object',
+          properties: {
+            scripts: {
+              type: 'object',
+              properties: {
```
gorkem, line 425: Can we have a non-string property added?

gorkem (changes requested):
A small improvement to the test cases but otherwise LGTM. Is there a corresponding PR needed for the vscode-yaml repo?

test/autoCompletion.test.ts
p-spacek: I'm sorry for the longer delay...
I modified the UT. Have a look if I understood correctly your note, please.
vscode-yaml PR is here https://github.com/redhat-developer/vscode-yaml/pull/667

gorkem (approved):
This looks ready to go. I am not sure why CI is unhappy with the new test though

evidolob:
@p-spacek Test fails because of https://github.com/redhat-developer/yaml-language-server/pull/609 , could you update this?

p-spacek:
> @p-spacek Test fails because of #609 , could you update this?

oh sure.
updated

evidolob (approved):

PR #609, opened 2021-12-07
feat: trim end $1 from completion

p-spacek:
### What does this PR do?
Simplify simple (with one variable) code completion when the text ends with $1.
For the user, it's quite complicated to not forget to hit `tab` instead of `enter`.

Note: not sure if it makes sense for your users. I would like to know your opinion. For me/users it's not very handy to use tab and enter together in these simple completion scenarios.

Before:

https://user-images.githubusercontent.com/38421337/144215088-c30c8663-3ee9-4c24-8ea2-d1fa7fa857f1.mov

After: 

https://user-images.githubusercontent.com/38421337/144215152-46e0f44d-3aea-47e8-9336-bb752b91f00b.mov

no ref

### Is it tested? How?
modified UTs


evidolob (approved):
LGTM
@JPinkney WDYT?

JPinkney (approved):

Issue #612, opened 2021-12-08
If maxProperties is set, completion does not work for the last property


sfalmo:
## Describe the bug
This is likely an off-by-one error.
Suppose a schema sets `maxProperties: 3` for a certain object. In the yaml file using this schema, completion works for the first and second property of this object, but no completion suggestions are shown if one wants to add a third property.

## Expected Behavior
The language server should suggest completions also in this case, since `maxProperties` sets an inclusive bound on the number of properties.

## Current Behavior
No completions are shown. This is especially frustrating if `maxProperties: 1`, since there is no way to see the available properties of an object at all in this case.

## Steps to Reproduce
Use the following example schema. Create `someObject` and add properties. You won't get completions for the third one.

```yaml
$schema: http://json-schema.org/draft-07/schema
title: Test maxProperties
description: Test schema for completion of properties if maxProperties is set
type: object
properties:
  someObject:
    type: object
    properties:
      prop1:
        type: string
      prop2:
        type: string
      prop3:
        type: string
      prop4:
        type: string
    maxProperties: 3
```
![prop1](https://user-images.githubusercontent.com/52037539/145193226-6439c258-f46c-45d5-a32b-fbc994c94115.png)

![prop2](https://user-images.githubusercontent.com/52037539/145193241-dcf3247b-ce04-416c-bee5-24900ffa84bb.png)

![prop3](https://user-images.githubusercontent.com/52037539/145193252-a6a81b7f-7cf1-47ba-b323-04a17e4ff9aa.png)

## Environment
- [ ] Windows
- [ ] Mac
- [x] Linux
- [ ] other (please specify)

PR #614, opened 2021-12-08
Fix completion of last property when maxProperties is set

sfalmo:
### What does this PR do?
Fix completion of properties if maxProperties is set and last allowed property is reached.

#612 

### Is it tested? How?
Added test for completion after beginning to type the last allowed property.


evidolob:
@sfalmo Could you update this PR with latest main branch, I have test fails on my side. 
They related to https://github.com/redhat-developer/yaml-language-server/pull/609

sfalmo:
I updated the test to the completion behavior that was introduced in #609, now it works again.

JPinkney (approved):

PR #619, opened 2021-12-14
Fix showing the snippets with description

msivasubramaniaan:
### What does this PR do?

The fix is showing the snippets with corresponding description

https://github.com/redhat-developer/vscode-yaml/issues/642

### Is it tested? How?
Yes with test cases


evidolob:
@msivasubramaniaan Could you add test for fix?

test/schema.test.ts
```diff
@@ -687,6 +689,20 @@ describe('JSON Schema', () => {
       assert.strictEqual(result.items.length, 1);
       assert.strictEqual(result.items[0].label, 'schemastore');
     });
+
+    it('Default snippet with descritpion', async () => {
```
JPinkney, line 693:
```suggestion
    it('Default snippet with description', async () => {
```

test/schema.test.ts
```diff
@@ -687,6 +689,20 @@ describe('JSON Schema', () => {
...
+      languageSettingsSetup.withSchemaFileMatch({
+        fileMatch: ['test.yaml'],
+        uri: TEST_URI,
+        priority: SchemaPriority.SchemaStore,
+        schema: schemaDefaultSnippetSample,
+      });
+      languageService.configure(languageSettingsSetup.languageSettings);
+      const testTextDocument = setupTextDocument('foo:  ');
+      const result = await languageService.doComplete(testTextDocument, Position.create(0, 5), false);
+      assert.strictEqual(result.items.length, 2);
```
JPinkney, line 703: Can you do `assert.equal` on the description value here so that we know it's set correctly

gorkem (changes requested):
Can you please update the author of these commits? Right now they are signed as `unknown` and I am not sure if `<msivasub@win.redhat.com>` is a valid mail address.

evidolob (approved):

Issue #586, opened 2021-11-12
Schema validation matches `@bitnami` as a `uri`-formatted `string`.


TBBle:
## Describe the bug

The [schema for helm Charts on SchemaStore](https://json.schemastore.org/chart.json) has the following for the `dependencies.repository` field:
```json
          "repository": {
            "description": "The repository URL or alias",
            "oneOf": [
              {
                "type": "string",
                "format": "uri"
              },
              {
                "type": "string",
                "pattern": "^@"
              }
            ]
          },
```

Given a valid Helm Chart.yaml:
```yaml
apiVersion: v2
name: services
description: Services

type: application

version: 0.1.1

appVersion: 0.1.0

dependencies:
- name: rabbitmq
  repository: '@bitnami'
  version: 8.15.2
- name: mongodb
  repository: '@bitnami'
  version: 8.3.0
```

validation fails, because the repository string `@bitnami` appears to match both of the elements of the `oneOf`.

However, `@bitnami` is not a valid `uri` format, as far as I can tell, per the [json-schema spec](https://json-schema.org/understanding-json-schema/reference/string.html#resource-identifiers). Per [RFC 3986 Secion 4.1](https://datatracker.ietf.org/doc/html/rfc3986#section-4.1) it'd be a valid `uri-reference`.

It's odd, because it appears yaml-language-server [_handles_ this case](https://github.com/redhat-developer/yaml-language-server/blob/1.2.0/src/languageservice/parser/jsonParser07.ts#L923-L930), but perhaps that logic is not mixing correctly with `oneOf`? (Or maybe this code isn't code path being used here, and the same logic is missing in some other validation)

## Expected Behavior
Valid Helm Chart.yaml validates.

## Current Behavior
```json
[{
	"resource": "/C:/Projects/Arsenal/services/charts/premspec-services/Chart.yaml",
	"owner": "_generated_diagnostic_collection_name_#2",
	"severity": 8,
	"message": "Matches multiple schemas when only one must validate.",
	"source": "yaml-schema: Helm Chart.yaml",
	"startLineNumber": 13,
	"startColumn": 15,
	"endLineNumber": 13,
	"endColumn": 16
},{
	"resource": "/C:/Projects/Arsenal/services/charts/premspec-services/Chart.yaml",
	"owner": "_generated_diagnostic_collection_name_#2",
	"severity": 8,
	"message": "Matches multiple schemas when only one must validate.",
[...]

Issue #312, opened 2020-09-02
More info for "Matches multiple schemas when only one must validate."


bukowa:
Hello, could this message tell us more about what schemas are matched?
````json
{
	"resource": "pathtofile.yaml",
	"owner": "_generated_diagnostic_collection_name_#1",
	"severity": 8,
	"message": "Matches multiple schemas when only one must validate.",
	"startLineNumber": 1,
	"startColumn": 1,
	"endLineNumber": 1,
	"endColumn": 2
}
````

**Comment by JPinkney** [comment]
It depends. In some cases that error occurs when a schema is written in a way that allows multiple subschemas to match one piece of content. For example, in kubernetes:
```
apiVersion: v1
```
matches multiple subschemas and it wouldn't really be feasible to show more info about the subschemas (since theres probably > 200 matching subschemas).

However, If there's multiple schemas associated to a file then I think it _should_ be possible to say which schemas are being validated against it and why its throwing that error

bukowa:
That's interesting that you mentioned `apiVersion` - i cannot get them to autocomplete is that possible? I think that the rest of the autocomplete for ex. `kind` should show only kinds for the `apiVersion` i specified, but even the `apiVersion` is not autocompleted.

JPinkney:
I think its an issue with the schema. The problem is there are so many subschemas inside of the kubernetes schema that it makes issues with the schema almost impossible to debug

rene-descartes2021:
I am experiencing this error on an array of objects distinquished by an oneOf enum string property. In the enum ["Foo", "FooBar"] this error is on each instance of "Foo". In general for every enumeration (in enum property set) if there exists another enumeration with a matching prefix then this error occurs on each instance of the former.
I will open new issue with minimal reproduction steps and link to it.

PR #623, opened 2021-12-22
updated maxMatch value based on oneOf size

msivasubramaniaan:
### What does this PR do?

This PR updated the maxMatch value based on the oneOf size. Because there may be a chances that node value matches all the schema which provided under oneOf 

https://github.com/redhat-developer/yaml-language-server/issues/586

### Is it tested? How?
Yes with test cases.

### DIFF
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -681,9 +681,13 @@ function validate(
       }
     }
 
-    const testAlternatives = (alternatives: JSONSchemaRef[], maxOneMatch: boolean): number => {
+    const testAlternatives = (alternatives: JSONSchemaRef[], oneOfMatch: boolean): number => {
       const matches = [];
-
+      let maxMatch = 1;
+      if (oneOfMatch) {
+        //there may be chances that node value matches all the schemas which provided under oneOf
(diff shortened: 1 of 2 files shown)


evidolob (approved):

src/languageservice/parser/jsonParser07.ts
```diff
@@ -707,11 +711,11 @@ function validate(
...
-          bestMatch = genericComparison(maxOneMatch, subValidationResult, bestMatch, subSchema, subMatchingSchemas);
+          bestMatch = genericComparison(oneOfMatch, subValidationResult, bestMatch, subSchema, subMatchingSchemas);
         }
       }
 
-      if (matches.length > 1 && maxOneMatch) {
+      if (matches.length > maxMatch && oneOfMatch) {
         validationResult.problems.push({
           location: { offset: node.offset, length: 1 },
           severity: DiagnosticSeverity.Warning,
```
gorkem, line 721: I do not think the following message is accurrate with this change anymore. It needs an updated message.

gorkem (changes requested):

gorkem (approved):

gorkem:
@evidolob do we need to still address #312?

Issue #615, opened 2021-12-08
Hover doesn't show up when a new line is behind nested prop with null value


p-spacek:
## Describe the bug
Hover doesn't show up when a new line is behind `prop` with null value.

```yaml
childObject:
  prop:

#emptyline
```

```
type: 'object',
properties: {
  childObject: {
    type: 'object',
    properties: {
      prop: {
        type: 'string',
        description: 'should return this description',
      },
    },
  },
},
```

## Expected Behavior
Hover will display on `prop` when there is no value and when new line is behind

## Current Behavior
No hover is displayed
Note: this yaml works ok
```yaml
childObject:
  prop:
```

## Environment
- [ ] Windows
- [x] Mac
- [ ] Linux
- [ ] other (please specify)

## Details
I'm seeing problem in `ast-converter.ts`
```ts
  if (start.line !== end.line && (lineCounter.lineStarts.length !== end.line || end.col === 1)) {
    result[1]--;
  }
```
basically, this line trim `\n` from objects/arrays. So root object is without ending newLine, but node
`let node = doc.getNodeFromOffset(offset);` contains this new line and has grater range then. So this condition inside `jsonParser07.ts` is not true
```ts
export function contains(node: ASTNode, offset: number, includeRightBound = false): boolean {
  return (
    (offset >= node.offset && offset <= node.offset + node.length) || (includeRightBound && offset === node.offset + node.length)
  );
}

// fails this one offset <= node.offset + node.length
```

Unit test that covers this:
```ts
it('Hover on null property in nested object', async () => {
      languageService.addSchema(SCHEMA_ID, {
        type: 'object',
        properties: {
          childObject: {
            type: 'object',
            properties: {
              prop: {
                type: 'string',
                description: 'should return this description',
              },
            },
          },
        },
      });
      const content = 'childObject:\n  prop: \n';

      const result = await parseSetup(content, content.indexOf('prop') + 1);
      console.log((result.contents as MarkupContent).value);

      assert.strictEqual(MarkupContent.is(result.contents), true);
      assert.strictEqual(
        (result.contents as MarkupContent).value,
        `should return this description\n\nSource: [${SCHEMA_ID}](file:///${SCHEMA_ID})`
      );
    });
```

I am not sure why there is this offset decreasing `result[1]--`
from UTs result it's related to documentSymbols, but why parser has to trim new line is not clear to me.

thanks a lot

PR #625, opened 2021-12-24
changed parser when last line is empty or null

msivasubramaniaan:
### What does this PR do?

This PR related to show the hover on the property when the next line is empty/null value

https://github.com/redhat-developer/yaml-language-server/issues/615

### Is it tested? How?
Yes. With test cases

### DIFF
--- a/src/languageservice/parser/yamlParser07.ts
+++ b/src/languageservice/parser/yamlParser07.ts
@@ -5,9 +5,11 @@
  *--------------------------------------------------------------------------------------------*/
 'use strict';
 
-import { Parser, Composer, Document, LineCounter, ParseOptions, DocumentOptions, SchemaOptions, CST, Lexer } from 'yaml';
+import { Parser, Composer, Document, LineCounter, ParseOptions, DocumentOptions, SchemaOptions } from 'yaml';
 import { YAMLDocument, SingleYAMLDocument } from './yaml-documents';
 import { getCustomTags } from './custom-tag-provider';
+import { TextDocument } from 'vscode-languageserver-textdocument';
+import { TextBuffer } from '../utils/textBuffer';
 
 export { YAMLDocument, SingleYAMLDocument };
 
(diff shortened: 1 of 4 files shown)


src/languageservice/parser/yamlParser07.ts
```diff
@@ -33,7 +33,8 @@ export function parse(text: string, parserOptions: ParserOptions = defaultOption
   };
   const composer = new Composer(options);
   const lineCounter = new LineCounter();
-  const parser = new Parser(lineCounter.addNewLine);
+  const regx = new RegExp('[\\n]+[\\s]*$', 'g');
```
evidolob, line 36: I don't really like this rexexp, could we achieve same behaviour without the regexp.

You apply that reg exp to file content, files can be big and it not very efficient to apply such regexps on each re parse.

msivasubramaniaan: Ya. I have changed the logic. Instead of using RegExp I have used Lexer from yaml for getting tokens of the text.

gorkem (approved):
lgtm

src/languageservice/parser/yamlParser07.ts
```diff
@@ -33,7 +33,9 @@ export function parse(text: string, parserOptions: ParserOptions = defaultOption
   };
   const composer = new Composer(options);
   const lineCounter = new LineCounter();
-  const parser = new Parser(lineCounter.addNewLine);
+  const lexerTokensArr = Array.from(new Lexer().lex(text));
```
evidolob, line 36: I still don't like this, as we parse/tokenize file content twice just to detect end of file.
I think it would be much better/efficient  if `parse` function receive `TextDocument` object instead of `string`,
by using `TextDocument` you can receive last line content and make all validation that you need.

evidolob (changes requested):

src/languageservice/parser/yamlParser07.ts
msivasubramaniaan: Using TextDocument now I am checking with the last line empty/not

evidolob (approved):

PR #627, opened 2021-12-28
when detect Indentation is false set the editior.tabSize based on settings value

msivasubramaniaan:
### What does this PR do?

https://user-images.githubusercontent.com/93245779/147567212-77383a76-bdfb-4701-b4fd-3bf9fbf0b99c.mp4

https://github.com/redhat-developer/vscode-yaml/issues/573

### Is it tested? How?
Yes. with test case


src/languageserver/handlers/settingsHandlers.ts
```diff
@@ -146,6 +153,17 @@ export class SettingsHandler {
     }
   }
 
+  public async updateTabSize(tabSize: number): Promise<boolean> {
```
evidolob, line 156: Not sure, why we need to update user settings?

msivasubramaniaan: As per the bug, when user changed the settings Editor: Detect Indentation is false, then we need to change the default tab size to Editor: Tab size settings value. For that we need to update the user settings from LS.

evidolob: I still not understand why we need to update property. If Editor: Detect Indentation is false formatter should use Editor: Tab, but we don't need to change any property for that

msivasubramaniaan: Removed the update property code.

evidolob (approved):

PR #628, opened 2022-01-01
feat: parent completion

p-spacek:
### What does this PR do?
Add the possibility to suggest and insert the whole object structure. So a user doesn't have to add required properties one by one.
Another advantage is when a user has to pick between a few 'anyOf' objects.
It also works with incomplete objects.

Display two possibilities (anyOf) of complex objects and insert the whole structure of all required props of the object. 

https://user-images.githubusercontent.com/38421337/147861510-b2aa75ea-b583-4fa8-8b07-8f213cddd7b2.mov

Suggest the rest of the 'parent' object (missing required props).

https://user-images.githubusercontent.com/38421337/147861490-bd2032eb-64e5-48c9-9a6f-02abeda0723f.mov

Added also feature that adds const value instead of $1. So instead of `prop: {$1:value}` is used prop: value. (beginning of the second video)

no ref

### Is it tested? How?
few unit tests

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -41,11 +41,16 @@ const localize = nls.loadMessageBundle();
 
 const doubleQuotesEscapeRegExp = /[\\]+"/g;
 
-interface CompletionItem extends CompletionItemBase {
-  schemaType?: string;
+const parentCompletionKind = CompletionItemKind.Class;
+
+interface ParentCompletionItemOptions {
+  schemaType: string;
   indent?: string;
-  isForParentSuggestion?: boolean;
-  isInlineObject?: boolean;
+  insertTexts?: string[];
+}
+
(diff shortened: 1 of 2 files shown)


src/languageservice/services/yamlCompletion.ts
```diff
@@ -394,9 +447,63 @@ export class YamlCompletion {
...
+    };
+
+    result.items.forEach((completionItem) => {
+      if (isParentCompletionItem(completionItem)) {
+        const indent = completionItem.parent.indent || '';
+
+        const reindexedTexts = reindexText(completionItem.parent.insertTexts);
+
+        // add indent to each object property and join completion item texts
+        // let insertText = reindexedTexts.map((text) => text.replace('\n', `${indent}\n`)).join(`\n${indent}`);
```
evidolob, line 482: Could you delete this line?

evidolob (approved):
Could you resolve merge conflict?

Issue #588, opened 2021-11-17
_PROXY environment and setting not honoured since 1.1.1


lcrockett:
## Describe the bug
Since version 1.1.1 PROXY environment variables are not honoured anymore. Referring to `HTTP_PROXY` and `HTTPS_PROXY`. This blocks the LS from downloading schema's. Setting a proxy via the settings in +1.1.1 does not work either. Due to a corporate firewall this prevents schema downloading functionality in +1.1.1.

## Expected Behavior
I would expect either that the PROXY environment variables or the proxy settings would work.

## Current Behavior
The error popping up in +1.1.1 when downloading schema's is `Unable to load schema from 'https://json.schemastore.org/pre-commit-config.json': connect ECONNREFUSED 20.42.128.105:443`

## Steps to Reproduce
1. Use settings block below
2. Ensure LS version is +1.1.1
3. Ensure a file is opened that matches one of the listings in the settings block (for instance: `.pre-commit-config.yaml`)
4. The error should pop up

## Environment
- [ ] Windows
- [ ] Mac
- [X ] Linux
- [ ] other (please specify)

NVIM v0.6.0-dev+589-gc0efe49e7 @ Ubuntu 20.04 LTS

## Additional information

Could this perhaps be cause by [this](https://github.com/redhat-developer/yaml-language-server/commit/27daa94333f7c0567978d891dfa5dbd334f8af09#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deL2169-L2171) ?

<details>
  <summary>LS settings</summary>
  
```lua
  settings = {
    http = {
      proxy = 'http://proxy.domain.tld:8080',
      proxyStrictSSL = false
    },
    redhat = {
      telemetry = { enabled = false }
    },
    yaml = {
      completion = true,
      editor = { formatOnType = true },
      format = {
        bracketSpacing = true,
        enable = true,
        proseWrap = 'preserve',
        singleQuote = true
      },
      hover = true,
      schemas = {
        ['https://raw.githubusercontent.com/awslabs/goformation/master/schema/cloudformation.schema.json'] = '**/cf-templates/**/*.{yml,yaml}',
        ['https://raw.githubusercontent.com/compose-spec/compose-spec/master/schema/compose-spec.json'] = 'docker-compose.yml.{yml,yaml}',
        ['https://json.schemastore.org/ansible-playbook'] = '{playbook,site}.{yaml,yml}',
[...]

PR #636, opened 2022-01-05
Fix proxy environment and settings not honored

msivasubramaniaan:
### What does this PR do?

The latest version of request-lib version 0.5.6 integrated. This version contains the proxy support

https://github.com/redhat-developer/yaml-language-server/issues/588

### Is it tested? How?
Yes with test cases


gorkem (changes requested):
Looks OK but should this also update the `yarn.lock` file ?
Also I have noticed that there is a new version of the `request-light` that is using the newer `http-proxy-agent`,

gorkem (approved):

Issue #585, opened 2021-11-11
Schema reference in URL hash can’t be resolved


remcohaszing:
## Describe the bug

When a schema uri uses a hash to select a nested schema, the language server can no longer resolve it. This was working in `yaml-language-server` 1.1.0 and below. It first appeared in 1.1.1.

## Expected Behavior

A nested schema can be referenced using a URL hash.

## Current Behavior

The following error occurs in VS Code:

```json
[{
	"resource": "/home/remco/Projects/appsemble/apps/person/app-definition.yaml",
	"owner": "_generated_diagnostic_collection_name_#0",
	"code": "768",
	"severity": 8,
	"message": "$ref '%2Fcomponents%2Fschemas%2FAppDefinition' in 'https://appsemble.app/api.json' can not be resolved.",
	"source": "YAML",
	"startLineNumber": 1,
	"startColumn": 1,
	"endLineNumber": 1,
	"endColumn": 2
}]
```

## Steps to Reproduce

1. Create a file named `app-definition.yaml`.
2. Open it in VS Code
3. Enter at least one character

## Environment
- [ ] Windows
- [ ] Mac
- [x] Linux
- [x] other (please specify) remcohaszing/monaco-yaml#126 ([demo](https://deploy-preview-126--monaco-yaml.netlify.app/))

remcohaszing:
This is a real blocker for me, so I decided to try and solve it myself.

I am not able to reproduce this issue by debugging by following [these steps](https://github.com/redhat-developer/vscode-yaml/blob/main/CONTRIBUTING.md), as in: It works as expected then using whatever versions. And yes, I **really** made sure I was using the correct versions, ran `yarn` and `yarn compile` in both projects in the correct order after doing a `git checkout` of several tags I tried.

Instead I decided to give it a try with [`vim-lsp`](https://github.com/prabirshrestha/vim-lsp). For this I used the tag `1.2.2`.

I use [`vim-lsp-settings`](https://github.com/mattn/vim-lsp-settings). This reproduces the error. I use the following `vimrc` snippet:

```viml
call plug#begin('~/.vim/plugged')

Plug 'prabirshrestha/vim-lsp'
Plug 'mattn/vim-lsp-settings'

call plug#end()
```

However, if I use a local clone of `yaml-language-server`, I can once again not reproduce the issue. For this I used the following `vimrc` snippet:

```viml
au User lsp_setup call lsp#register_server({
  \ 'name': 'yaml',
[...]

PR #637, opened 2022-01-07
Lock vscode-json-languageservice@4.1.8

remcohaszing:
### What does this PR do?

This is a around #585 by locking `vscode-json-languageservice` to version 4.1.8.

It’s a workaround, not a desirable fix.

#585\
microsoft/vscode-json-languageservice#123

### Is it tested? How?

All steps are run from the project root using `vim` and the following configuration:

```viml
au User lsp_setup call lsp#register_server({
  \ 'name': 'yaml',
  \ 'cmd': {server_info->['node', 'out/server/src/server.js', '--stdio']},
  \ 'allowlist': ['yaml'],
\})

call plug#begin('~/.vim/plugged')

Plug 'prabirshrestha/vim-lsp'

call plug#end()
```

Clone and setup the project

```sh
git clone git@github.com:redhat-developer/yaml-language-server.git
yarn
yarn compile
yarn add vscode-json-languageservice@4.1.8
```

Create a file named `app-definition.yaml` with the following content. Use vim to edit it.

```yaml
name: 123
```

This correctly shows that `name` is invalid and some properties are missing.

Now close vim and run the following command:

```sh
yarn add vscode-json-languageservice@4.1.9
```

Now when editing the same file again, it shows an error saying the JSON schema can’t be resolved.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   },
   "dependencies": {
     "request-light": "^0.5.5",
-    "vscode-json-languageservice": "^4.1.7",
+    "vscode-json-languageservice": "4.1.8",
     "vscode-languageserver": "^7.0.0",
     "vscode-languageserver-textdocument": "^1.0.1",
     "vscode-languageserver-types": "^3.16.0",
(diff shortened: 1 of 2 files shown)


gorkem (approved):
LGTM

PR #643, opened 2022-01-18
Fix errors reported by telemetry

evidolob:
### What does this PR do?
It provided fixes for error reported by telemetry, like https://github.com/redhat-developer/vscode-yaml/issues/653 and https://github.com/redhat-developer/vscode-yaml/issues/651
Also it fixes `TypeError: Cannot read property 'offset' of undefined`,
`TypeError: e.indexOf is not a function at w` and `TypeError: Cannot read property 'type' of undefined`

Fix: https://github.com/redhat-developer/vscode-yaml/issues/653
Fix: https://github.com/redhat-developer/vscode-yaml/issues/651

### Is it tested? How?
with tests

### DIFF
--- a/src/languageservice/parser/ast-converter.ts
+++ b/src/languageservice/parser/ast-converter.ts
@@ -138,6 +138,18 @@ function convertScalar(node: Scalar, parent: ASTNode): ASTNode {
       result.isInteger = Number.isInteger(result.value);
       return result;
     }
+    default: {
+      if (node.tag === 'tag:yaml.org,2002:binary') {
+        const result = new StringASTNodeImpl(parent, node, ...toOffsetLength(node.range));
+        result.value = node.source;
+        return result;
+      } else {
+        // fail safe converting, we need to return some node anyway
+        const result = new StringASTNodeImpl(parent, node, ...toOffsetLength(node.range));
+        result.value = node.source;
+        return result;
+      }
+    }
   }
 }
 
--- a/src/languageservice/parser/ast-converter.ts
+++ b/src/languageservice/parser/ast-converter.ts
@@ -143,16 +143,10 @@ function convertScalar(node: Scalar, parent: ASTNode): ASTNode {
       return result;
     }
     default: {
-      if (node.tag === 'tag:yaml.org,2002:binary') {
-        const result = new StringASTNodeImpl(parent, node, ...toOffsetLength(node.range));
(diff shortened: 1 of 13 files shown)


src/languageservice/parser/ast-converter.ts
```diff
@@ -138,6 +142,18 @@ function convertScalar(node: Scalar, parent: ASTNode): ASTNode {
       result.isInteger = Number.isInteger(result.value);
       return result;
     }
+    default: {
+      if (node.tag === 'tag:yaml.org,2002:binary') {
```
gorkem, line 146: Why do we need this `if` in here? It looks like we are constructing exactly the same result to return?

src/languageservice/parser/jsonParser07.ts
```diff
@@ -233,6 +233,16 @@ export function asSchema(schema: JSONSchemaRef): JSONSchema {
   if (isBoolean(schema)) {
     return schema ? {} : { not: {} };
   }
+  if (schema === undefined) {
```
gorkem, line 236: Shouldn't this `if` come before the one above and simplfy the body of it?

src/languageservice/parser/jsonParser07.ts
```diff
@@ -233,6 +233,16 @@ export function asSchema(schema: JSONSchemaRef): JSONSchema {
...
+    return schema;
+  }
+  if (typeof schema !== 'object') {
+    // we need to report this case as JSONSchemaRef MUST be an Object or Boolean
+    console.warn(`Wrong schema: ${JSON.stringify(schema)}, it MUST be an Object or Boolean`);
```
gorkem, line 241: What do we expect to happen when we log this ? Do we expect users to refer to the logs?

gorkem (changes requested):

src/languageservice/parser/jsonParser07.ts
evidolob: That is can happen when user use not entirely valid JSON Schema. 
I not sure how we should report that to user. As that check can be called multiple times, so I don't want to spam user with multiple error/warning notifications.

evidolob:
@gorkem Could you look on this?

gorkem (approved):

msivasubramaniaan (approved):

PR #645, opened 2022-01-20
Use copy of AST for completion

evidolob:
### What does this PR do?
It uses copy of AST for completion, as during completion we modify AST and that modification can trigger validation errors.

Fix: https://github.com/redhat-developer/vscode-yaml/issues/671

### Is it tested? How?
with test


gorkem:
What are the reasons that we modify the AST during completion? Is it only to fix incomplete parts?

evidolob:
Not only, we also modify AST to get proper sub JSON Schema match during code completion

gorkem (approved):

PR #648, opened 2022-02-02
fix mulitple schema matches in oneOf based on the propertiesMatches value

msivasubramaniaan:
### What does this PR do?

This PR checks the multiple  schema matches under oneOf based on the sub schema properties matches value

https://github.com/redhat-developer/vscode-yaml/issues/683
https://github.com/redhat-developer/yaml-language-server/issues/586

### Is it tested? How?
yes with test cases

### DIFF
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -682,12 +682,9 @@ function validate(
       }
     }
 
-    const testAlternatives = (alternatives: JSONSchemaRef[], minOneMatch: boolean): number => {
+    const testAlternatives = (alternatives: JSONSchemaRef[], maxOneMatch: boolean): number => {
       const matches = [];
-      let minMatch = 1;
-      if (minOneMatch) {
-        minMatch = alternatives.length;
-      }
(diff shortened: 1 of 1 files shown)


evidolob:
@msivasubramaniaan It seems that you forgot to add tests

msivasubramaniaan:
@evidolob The test case was added https://github.com/redhat-developer/yaml-language-server/pull/623/files as part of this PR.

evidolob:
Then I miss understand the case, if you add tests in #623 and with this changes tests not failing then your test is not correct, please make sure that we have test witch tests changes in this PR.

msivasubramaniaan:
> Then I miss understand the case, if you add tests in #623 and with this changes tests not failing then your test is not correct, please make sure that we have test witch tests changes in this PR.

As part of #623 the given YAML matched both oneOf cases so it didn't fail with the latest fix. But https://github.com/redhat-developer/vscode-yaml/issues/683 the YAML not matched both of given schema in oneOf and some properties are still missing. So with the latest fix the error count will not be 0. There are some test cases which covered the same scenario already

evidolob (approved):

Issue #652, opened 2022-02-04
TypeError: Cannot create property 'url' on string 


evidolob:
## Describe the bug

Reported by telemetry on 1.4.0 version.

Stack trace: 
```
TypeError: Cannot create property 'url' on string 'asdasd' at O (/Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.4.0/dist/languageserver.js:2:521466)
 at f.getMatchingSchemas (/Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.4.0/dist/languageserver.js:2:521229) 
at t.YamlCompletion.addPropertyCompletions (/Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.4.0/dist/languageserver.js:2:557523) 
at t.YamlCompletion.<anonymous> (/Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.4.0/dist/languageserver.js:2:556075) 
at Generator.next (<anonymous>) at s (/Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.4.0/dist/languageserver.js:2:549789)
```

And:
```
TypeError: Cannot create property 'url' on number '1.1' at /Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.4.0/dist/languageserver.js:2:582787
```

PR #656, opened 2022-02-04
Handle case when schema not JS object

evidolob:
### What does this PR do?
Add check if schema is `object` in validation function.

Fix: https://github.com/redhat-developer/yaml-language-server/issues/652

### Is it tested? How?
with test

### DIFF
--- a/test/schemaValidation.test.ts
+++ b/test/schemaValidation.test.ts
@@ -1560,6 +1560,15 @@ obj:
   });
 
   describe('Bug fixes', () => {
+    it('should handle "additionalItems" not schema or boolean', async () => {
+      const schema = 'Foo';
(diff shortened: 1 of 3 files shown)


test/schemaValidation.test.ts
```diff
@@ -1560,6 +1560,15 @@ obj:
   });
 
   describe('Bug fixes', () => {
+    it('should handle "additionalItems" not schema or boolean', async () => {
```
gorkem, line 1563: Is this an accurate description of the test?

evidolob: No, it wasn’t

gorkem (approved):

msivasubramaniaan (approved):

PR #659, opened 2022-02-11
fix/suggestion item title

p-spacek:
### What does this PR do?
 - The custom title of the suggestion item didn't work properly.
 - Add a description of the schema in 'object/parent' suggestion description. 

This PR fix logic of the `schema.title`.

I had a problem, that the previous `title` could be modified by title from parents. So `schema.title` could contain a title from its parent. And then I wasn't able to return the correct name of a suggestion item (display of the real title, name of the ref object, parent title).

So this PR adds a new `closestTitle` prop, that should cover previous functionality.

I went through all `schema.title` and tried to decide where to use `title` or `closesTitle`. Results of the UT are ok, but I will be better to check this from your side.

no ref

### Is it tested? How?
Modified one UT that covers this change

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,8 @@
     "build:libs": "yarn compile:umd && yarn compile:esm",
     "compile:umd": "tsc -p ./tsconfig.umd.json",
     "compile:esm": "tsc -p ./tsconfig.esm.json",
-    "check-dependencies": "node ./scripts/check-dependencies.js"
+    "check-dependencies": "node ./scripts/check-dependencies.js",
+    "pull-remote": "git pull https://github.com/redhat-developer/yaml-language-server.git main"
   },
   "nyc": {
     "extension": [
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -81,7 +81,8 @@
     "build:libs": "yarn compile:umd && yarn compile:esm",
     "compile:umd": "tsc -p ./tsconfig.umd.json",
     "compile:esm": "tsc -p ./tsconfig.esm.json",
-    "check-dependencies": "node ./scripts/check-dependencies.js"
+    "check-dependencies": "node ./scripts/check-dependencies.js",
+    "pull-remote": "git pull https://github.com/redhat-developer/yaml-language-server.git main"
```
p-spacek, line 85: this change shouldn't be there. 
I can remove this if you don't want this in scrips...

but this command is useful to merge changes from your repo into forked one...

gorkem: It is better to remove. It will confuse me :)

gorkem (approved):
Thank you

msivasubramaniaan (approved):

PR #660, opened 2022-02-13
Add: show examples on hover.

yassun4dev:
### What does this PR do?
If "examples" is defined in the schema, display it on the hover.

I think it useful to show what kind of input is expected for a string in user format.

### Is it tested? How?
A test was added.

### DIFF
--- a/src/languageservice/services/yamlHover.ts
+++ b/src/languageservice/services/yamlHover.ts
@@ -102,8 +102,10 @@ export class YAMLHover {
 
         let title: string | undefined = undefined;
         let markdownDescription: string | undefined = undefined;
-        let markdownEnumValueDescription: string | undefined = undefined,
-          enumValue: string | undefined = undefined;
+        let markdownEnumValueDescription: string | undefined = undefined;
+        let enumValue: string | undefined = undefined;
+        const markdownExamples: string[] = [];
+
         matchingSchemas.every((s) => {
           if (s.node === node && !s.inverted && s.schema) {
             title = title || s.schema.title;
@@ -121,6 +123,11 @@ export class YAMLHover {
                   enumValue = JSON.stringify(enumValue);
                 }
               }
+            }
+            if (s.schema.examples) {
+              s.schema.examples.forEach((example) => {
+                markdownExamples.push(JSON.stringify(example));
+              });
             }
           }
           return true;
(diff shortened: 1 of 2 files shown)


evidolob:
@yassun4dev We have test failing on this PR.
Also could you fix sign of the commit, it seems that you use two accounts.

yassun4dev:
@evidolob Oh, sorry...

I fixed sign of the commit, and check why the test failed on Linux only.
( I'm working on Mac, so it will take a little time. )

gorkem:
@yassun4dev Are there any JSON schemas that you can suggest for testing this functionality?

msivasubramaniaan (approved):

yassun4dev:
@gorkem I added one test case and put a simple schema in it.

You need more comprehensive testing as separate schema files?

gorkem:
Not necessarily for testing but wanted to see real world usages.

evidolob (approved):

yassun4dev:
@gorkem I see.

I have encountered some cases that require examples in real world.

1. It is too complicated to describe with 'pattern', and 'examples' assists the reading.

```jsonc
{
  "type": "object",
  "properties": {
    "range": {
      "type": "string",
      "description": "Sheet range of Excel.",
      "pattern": "^[A-Z]+[1-9][0-9]*:[A-Z]+[1-9][0-9]*$", // It's too complicated pattern.
      "examples": ["A1:B12"] // Easy to read and understand. 
    }
  }
}
```

2. Since the exception data already exists, "enum" and "pattern" cannot be used, but a recommended hint is needed.

src/languageservice/services/yamlHover.ts
```diff
@@ -141,7 +148,16 @@ export class YAMLHover {
           }
           result += `\`${toMarkdownCodeBlock(enumValue)}\`: ${markdownEnumValueDescription}`;
         }
-
+        if (markdownExamples.length !== 0) {
+          if (result.length > 0) {
+            result += '\n\n';
+          }
+          result += 'Examples: [\n\n';
```
gorkem, line 155: Not a big one but I am not sure if we need the `[` in here. The hovers looked better for me without them

gorkem (approved):
Thanks for the contribution. Just one comment on the format of the hover. It is up-to you if you want to address it or not.

src/languageservice/services/yamlHover.ts
yassun4dev: @gorkem I fixed hover format. I think it looks better. Thank you!!

evidolob (approved):

PR #661, opened 2022-02-13
Remove data from unused anchor diagnostics

remcohaszing:
### What does this PR do?

The anchor name was added as custom data. It’s unnecessary, because the name of the anchor can be determined from other values in the place where it’s used.

Because Monaco Editor marker data doesn’t support custom data, this blocks supports code actions in monaco-yaml.

### Is it tested? How?

Manually using monaco-yaml and using tests.


gorkem (approved):
Thanks for noticing.

PR #663, opened 2022-02-17
Schema validation should check the object's oneOf if it has instead of the object type

msivasubramaniaan:
### What does this PR do?

The PR will check the oneOf array and do the type check against the schemas mentioned over it if the particular schema object has, instead of directly check the schema object type.

https://github.com/redhat-developer/vscode-yaml/issues/692

### Is it tested? How?
Yes with UT

### DIFF
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -602,7 +602,8 @@ function validate(
   originalSchema: JSONSchema,
   validationResult: ValidationResult,
   matchingSchemas: ISchemaCollector,
-  options: Options
+  options: Options,
+  isAdditionalPropertiesCheck?: boolean
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
 ): any {
   const { isKubernetes } = options;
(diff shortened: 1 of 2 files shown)


msivasubramaniaan:
@evidolob please review this PR

src/languageservice/parser/jsonParser07.ts
```diff
@@ -646,6 +647,16 @@ function validate(
...
     }
 
+    function matchesSchemaType(schema: JSONSchema): boolean {
+      /* when schema type is object and it contains oneOf then needs to validate that particular types
+        reference issue 692
+      */
+      if (schema.type === 'object' && typeof schema.additionalProperties === 'object' && schema.additionalProperties.oneOf) {
+        return true;
+      }
+      return node.type === schema.type || (schema.type === 'integer' && node.type === 'number' && node.isInteger);
```
gorkem, line 657: Shouldn't this just call `matchesType` from above? It looks like it is duplicating the code.

gorkem (changes requested):

gorkem (approved):

evidolob (approved):

Issue #620, opened 2021-12-14
array object completion - should not suggest const


p-spacek:
## Describe the bug
array object completion - should not suggest const

schema:

```json
{
  "properties": {
    "test": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "constProp": {
            "type": "string",
            "const": "const1"
          }
       }
    }
  "type": "object",
}
```

yaml:
```yaml
test:
  - constProp
    # cursor here
```

![image](https://user-images.githubusercontent.com/38421337/145993582-65cc8aef-15c3-492d-9d42-fa0226fe0479.png)

this test fails
```ts
    it('array completion - should not suggest const', async () => {
      languageService.addSchema(SCHEMA_ID, {
        type: 'object',
        properties: {
          test: {
            type: 'array',
            items: {
              type: 'object',
              properties: {
                constProp: {
                  type: 'string',
                  const: 'const1',
                },
              },
            },
          },
        },
      });
      const content = 'test:\n  - constProp:\n    ';
      const result = await parseSetup(content, content.length);
      expect(result.items.length).to.be.equal(0);
    });
```

problem is probably here in yamlCompletion.ts
check for array is missing

```ts
// it will treated as a property key since `:` has been appended
if (nodeParent && isSeq(nodeParent) && schema.schema.type !== 'object') {
  this.addSchemaValueCompletions(schema.schema, separatorAfter, collector, {});
}
```

## Expected Behavior
should not suggest const value inside the array
But if it's not an array, just an object, completion works ok.
![image](https://user-images.githubusercontent.com/38421337/145994092-dd5cd633-b853-47e7-ab78-09e3b27b08e6.png)

## Environment
Mac

PR #670, opened 2022-03-02
fixed auto completion not suggest const for array

msivasubramaniaan:
### What does this PR do?

Auto completion should not suggest const value from schema for Array

https://github.com/redhat-developer/yaml-language-server/issues/620

### Is it tested? How?

Yes. with UT

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -538,7 +538,6 @@ export class YamlCompletion {
     const hasColon = lineContent.indexOf(':') !== -1;
 
     const nodeParent = doc.getParent(node);
-    const isArray = nodeParent && isSeq(nodeParent) && nodeParent.items.length > 0 ? true : false;
     const matchOriginal = matchingSchemas.find((it) => it.node.internalNode === originalNode && it.schema.properties);
     for (const schema of matchingSchemas) {
       if (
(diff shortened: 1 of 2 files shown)


src/languageservice/services/yamlCompletion.ts
```diff
@@ -663,7 +663,7 @@ export class YamlCompletion {
         //    - item1
         // it will treated as a property key since `:` has been appended
         if (nodeParent && isSeq(nodeParent) && schema.schema.type !== 'object') {
-          this.addSchemaValueCompletions(schema.schema, separatorAfter, collector, {});
+          this.addSchemaValueCompletions(schema.schema, separatorAfter, collector, {}, isArray);
```
gorkem, line 666: If this is the only place `isArray` is used can we delay its initialization to inside this `if` block?

gorkem (changes requested):

src/languageservice/services/yamlCompletion.ts
```diff
@@ -663,7 +662,8 @@ export class YamlCompletion {
         //    - item1
         // it will treated as a property key since `:` has been appended
         if (nodeParent && isSeq(nodeParent) && schema.schema.type !== 'object') {
-          this.addSchemaValueCompletions(schema.schema, separatorAfter, collector, {});
+          const isArray = nodeParent.items.length > 0;
```
evidolob, line 665: Better to use `Array.isArray()` to check is `nodeParent.items` array

evidolob (commented):
LGTM

evidolob (approved):
LGTM

gorkem (approved):
lgtm

PR #671, opened 2022-03-04
fix: defaultSnippets insert text

p-spacek:
### What does this PR do?
fix various snippets problems:
 - array of strings that was split into chars
 - remove extra line between objects inside array
 - indent from nested object inside array
 - remove extra spaces after property colon

add property `sortText` into defaultSnippets, so it's possible to sort snippets inside code-completion panel

no ref

### Is it tested? How?
added unit tests and fixed existing

### DIFF
--- a/src/languageservice/utils/json.ts
+++ b/src/languageservice/utils/json.ts
@@ -9,11 +9,15 @@ export interface StringifySettings {
   shouldIndentWithTab: boolean;
 }
 
+interface StringifySettingsInternal extends StringifySettings {
+  indentation: string;
+}
+
 export function stringifyObject(
   obj: unknown,
   indent: string,
   stringifyLiteral: (val: unknown) => string,
-  settings: StringifySettings,
+  settings: StringifySettingsInternal,
   depth = 0,
   consecutiveArrays = 0
 ): string {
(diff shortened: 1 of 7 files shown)


src/languageservice/utils/json.ts
```diff
@@ -50,18 +50,23 @@ export function stringifyObject(
       let result = (depth === 0 && settings.newLineFirst) || depth > 0 ? '\n' : '';
       for (let i = 0; i < keys.length; i++) {
         const key = keys[i];
+        const isObject = typeof obj[key] === 'object';
+        const colonDelimiter = isObject ? ':' : ': '; // add space only when value is primitive
+        const parentArrayCompensation = isObject && /^\s|-/.test(key) ? '  ' : ''; // add extra space if parent is an array
```
msivasubramaniaan, line 55: extra space should be calculated based on the indentation

msivasubramaniaan, line 55: extra space should be calculated based on the indentation

msivasubramaniaan (changes requested):

src/languageservice/utils/json.ts
p-spacek: Thanks, it makes sense
I fixed it

msivasubramaniaan (approved):

PR #673, opened 2022-03-07
Re-enable folding ranges

gorkem:
### What does this PR do?
Enables folding ranges. Adds the ability to fold documents when there is more than one document. Adds tests for making sure comments are not part of a region.

fixes https://github.com/redhat-developer/vscode-yaml/issues/413
fixes https://github.com/redhat-developer/vscode-yaml/issues/561

### Is it tested? How?

2 new tests are added to cover the case.


msivasubramaniaan (changes requested):
typo on the method name. Please refactor it.

msivasubramaniaan (approved):

Issue #665, opened 2022-02-20
"Missing property" is reported for one of required properties


kazinad:
## Describe the bug

I have yaml files with a custom schema (see below) where a "Missing property" problem is reported with the extension versions 1.0 and above. Earlier versions worked fine.

exmaple-schema.yml:

```yaml
# yaml-language-server: $schema=http://json-schema.org/draft-07/schema
$schema: http://json-schema.org/draft-07/schema
type: object
additionalProperties: false
properties:
  columns:
    title: Contains column definitions
    type: object
    patternProperties:
      "^[a-zA-Z]+$":
        title: A column definition
        type: object
        additionalProperties: false
        properties:
          int:
            title: Integer column type
            type: "null"
          long:
            title: Long column type
            type: "null"
          id:
            title: Identity column
            type: "null"
          unique:
            title: Defines single column database unique constraint
            type: "null"
        oneOf:
          - required:
              - int
          - required:
              - long
```

example.yml:

```yaml
# yaml-language-server: $schema=./example-schema.yml
columns:
  ColumnA: { int, id }
  ColumnB: { long, unique }
  ColumnC: { int: null, unique}
```

## Expected Behavior

No problems should be reported for "int" and "long" properties. Please note, this works good for "id" and "unique" properties.

## Current Behavior

![image](https://user-images.githubusercontent.com/39874391/154848387-870aabe0-a058-4bbd-9fe8-f3298e4aecb4.png)

## Steps to Reproduce

1. Ensure, newest version is installed of the extension
2. Save and open the above example files

## Environment
- [X] Windows
- [ ] Mac
- [ ] Linux
- [ ] other (please specify)

kazinad:
I confirm, the fix is working good with version 1.6 on my machine. Thank you guys! 👍

PR #678, opened 2022-03-09
changed on specific to undefined check instead of null and undefined check

msivasubramaniaan:
### What does this PR do?

This PR checks on the missing property validation that the expected property is undefined or not instead of null

https://github.com/redhat-developer/yaml-language-server/issues/665

### Is it tested? How?
Yes, added UT for this scenario


evidolob (approved):

gorkem (approved):
lgtm

PR #681, opened 2022-03-16
fix: differentiate similar schemas on completion

p-spacek:
### What does this PR do?
If multiple schemas inside of `anyOf` have the same final calculated schema name, 'parent/skeleton' autocompletion merge properties from different schema into one completion item.
 
wrong merge of multiple subschemas:
image

after fix:
image

note that a fixed example is not ideal. Schema creator should add titles into the schemas for a proper fix.
```json
"leftElement": {
  "anyOf": [
    {
      "type": "object",
      "properties": { ... },
      "title": "Option 1"
    },
    {
      "type": "object",
      "properties": { ... },
      "title": "Option 2"
    },
   ...
  ]  
```

### Is it tested? How?
modified Unit test
 - the previous unit test doesn't make much sense

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -45,8 +45,7 @@ const doubleQuotesEscapeRegExp = /[\\]+"/g;
 const parentCompletionKind = CompletionItemKind.Class;
 
 interface ParentCompletionItemOptions {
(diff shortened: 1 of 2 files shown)


gorkem (approved):
lgtm

msivasubramaniaan (approved):

PR #682, opened 2022-03-18
Add enable disable option on schema validation for file level

msivasubramaniaan:
### What does this PR do?

User can provide option on enable: true/false for schema validation on file path level

https://user-images.githubusercontent.com/93245779/159001884-74df1eeb-2f92-4c31-a73f-bbabd5a1e21e.mp4

https://github.com/redhat-developer/vscode-yaml/issues/649

### Is it tested? How?
yes added UT

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ The following settings are supported:
 - `yaml.hover`: Enable/disable hover
 - `yaml.completion`: Enable/disable autocompletion
 - `yaml.schemas`: Helps you associate schemas with files in a glob pattern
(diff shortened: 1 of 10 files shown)


gorkem:
I do not see the value of this feature. There are many and more effective ways to manage a list of URLs and update them to a  JSON file. This is making the code for YAML language intelligence more complicated because somebody may want to keep a list of YAML mappings on the settings.

msivasubramaniaan:
As @gorkem statement user can use plain txt file for handle this scenario. Hence closing this PR

PR #683, opened 2022-03-22
changed comparison logic on result match while showing error

msivasubramaniaan:
### What does this PR do?

This PR showing the opt matched errors while validating the YAML

https://github.com/redhat-developer/vscode-yaml/issues/692
https://github.com/redhat-developer/vscode-yaml/issues/708

### Is it tested? How?
Yes added UT and tested all the mentioned scenario


evidolob (approved):
LGTM

Issue #668, opened 2022-02-21
Add option to specify schema validation severity level


neodescis:
### Is your enhancement related to a problem? Please describe.
I am using monaco-editor and monaco-yaml. I need to configure the editor such that user input YAML is validated against a schema, with validation failures reported as errors to the user (as opposed to warnings).

### Describe the solution you would like
The built-in JSON support of monaco-editor provides a configurable severity for schema validation failures (see [here](https://github.com/microsoft/monaco-editor/blob/7826dee/website/typedoc/monaco.d.ts#L7535) ). I would suggest similar functionality for yaml-language-server. If this option is added, the monaco-yaml maintainer has expressed support for passing the option through: remcohaszing/monaco-yaml#149

### Describe alternatives you have considered
@remcohaszing and I also discussed adding a hook into monaco-yaml to allow post-processing of warnings/errors, but this approach is more consistent with monaco-editor's built-in JSON support.

gorkem:
I do not see a problem with adding this option—definitely a better choice than post-processing. The fastest way to implement this would be to provide a PR on it. Let me know if you would like to take a shot at providing one.

neodescis:
I will try to get a PR together, hopefully next week.

GrzegorzKaczan:
Any update on this? :)

PR #692, opened 2022-04-01
Add schema validation failure severity setting

neodescis:
### What does this PR do?
This change adds a setting to language services which allows the severity of a schema validation failure to be specified. The intent of the change is to allow the monaco-yaml package to set the severity level, and that package is using language services directly for its validation.

This does not change the language server, though maybe that should be changed as well? In the ValidationHandler's validateTextDocument function, I see the severity level being converted there.

#668 

### Is it tested? How?
Existing tests pass, as the severity defaults to Warning. Verified tests fail if set to a different value.

### DIFF
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -521,6 +521,7 @@ export function findNodeAtOffset(node: ASTNode, offset: number, includeRightBoun
 export class JSONDocument {
   public isKubernetes: boolean;
   public disableAdditionalProperties: boolean;
+  public validationFailureSeverity: DiagnosticSeverity;
 
   constructor(
     public readonly root: ASTNode,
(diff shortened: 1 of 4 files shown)


gorkem:
I am reluctant on this one because I feel like YAML errors and JSON Schema-based validation severity levels should be separately configurable.

gorkem:
Although I have not tested, I do not think your changes affect the severity of the YAML validation errors. Was it your intention not to change them? 

The errors and warnings for YAML are generated during parsing and come from `SingleYamlDocument` in [here](https://github.com/redhat-developer/yaml-language-server/blob/417a68eb6b46d49e9b39fab4171e1eb85c087b99/src/languageservice/services/yamlValidation.ts#L88).

gorkem:
Although I have not tested, I do not think your changes affect the severity of the YAML validation errors. Was it your intention not to change them? 

The errors and warnings for YAML are generated during parsing and come from `SingleYamlDocument` in [here](https://github.com/redhat-developer/yaml-language-server/blob/417a68eb6b46d49e9b39fab4171e1eb85c087b99/src/languageservice/services/yamlValidation.ts#L88).

neodescis:
This was what was generating the warning that I was trying to change to an error, but I can see making the JSON schema validation configurable would be useful as well. At this point though, it's been two months since I've looked at this, and we've decided to leave things as warnings.

PR #696, opened 2022-04-05
fix: completion in the middle of the empty text

p-spacek:
### What does this PR do?
fix completion when the cursor is in the middle of the empty text

https://user-images.githubusercontent.com/38421337/161766640-d52a9a27-5156-4ad9-abc4-234c1bf1f04c.mov

no ref

### Is it tested? How?
added UT

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -126,7 +126,7 @@ export class YamlCompletion {
 
     const currentWord = this.getCurrentWord(document, offset);
 
-    let overwriteRange = null;
+    let overwriteRange: Range = null;
     if (node && isScalar(node) && node.value === 'null') {
       const nodeStartPos = document.positionAt(node.range[0]);
       nodeStartPos.character += 1;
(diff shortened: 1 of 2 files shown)


gorkem (approved):
LGTM

gorkem:
Should this also work for boolean properties ?

p-spacek:
> Should this also work for boolean properties ?

https://user-images.githubusercontent.com/38421337/162401877-782b6498-4034-481e-aade-63b7041e084b.mov

it seems ok.

but there are probably more other similar problems with not vert valid YAML structure when users expect correct behaviour.
It's possible to fix one by one and add more and more conditions/fallbacks. But I thought that there could be also an approach to modify yaml before yamlCompletion.doComplete evaluates the result. 
For example modify (temporary without modification of the original file)
 - 'prop:' -> 'prop: ' then evaluation of the completion items
 - '-' -> '- '
 - 'prop:   #  ' -> 'prop: #' (#means cursor position)
 - 'prop:       ' -> 'prop: ' 

and others.

so maybe next time, we can try the cleanup befo parsing of the yaml?

gorkem:
Thanks for testing it.

msivasubramaniaan (approved):

msivasubramaniaan:
@p-spacek it needs a rebase.

PR #697, opened 2022-04-05
fix array completion in the middle of the empty text

p-spacek:
### What does this PR do?
fix array completion when the cursor is in the middle of the empty text

before:

https://user-images.githubusercontent.com/38421337/161768169-198a26ce-d30b-48f7-bd42-5dcf724a6468.mov

after:

https://user-images.githubusercontent.com/38421337/161768203-082356fc-063a-4dab-9f2c-4687594bb28a.mov

schema for this test
```yaml
"arrayObj": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "item1": {
            "type": "string"
          },
          "item2": {
            "type": "string",
          }
        },
        "required": [
          "item1",
          "item2"
        ]
      }
    },
```

no ref

### Is it tested? How?
added UT, 
modified existing one

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -313,8 +313,20 @@ export class YamlCompletion {
         }
       }
 
-      const originalNode = node;
+      let originalNode = node;
       if (node) {
+        // when the array item value is null but the cursor is between '-' and null value (cursor is not at the end of the line)
+        if (isSeq(node) && node.items.length && isScalar(node.items[0])) {
+          const nullNode = node.items[0];
+          if (
+            nullNode.value === null && // value is null
+            nullNode.range[0] > offset // cursor is before null
+          ) {
+            node = nullNode;
+            originalNode = node;
+            overwriteRange.end.character += nullNode.range[2] - offset; // extend range to the end of the null element
+          }
+        }
         if (lineContent.length === 0) {
           node = currentDoc.internalDocument.contents as Node;
         } else {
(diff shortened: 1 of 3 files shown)


test/autoCompletion.test.ts
```diff
@@ -2364,10 +2364,44 @@ describe('Auto Completion Tests', () => {
       });
 
       const content = 'test:\n  - and\n  - - ';
-      const completion = await parseSetup(content, 19);
+
+      const completion = await parseSetup(content, 20);
```
p-spacek, line 2368: The previous index 19 is probably not correct. index 20 is the end of the content

src/languageservice/services/yamlCompletion.ts
```diff
@@ -313,8 +313,20 @@ export class YamlCompletion {
...
+        // when the array item value is null but the cursor is between '-' and null value (cursor is not at the end of the line)
+        if (isSeq(node) && node.items.length && isScalar(node.items[0]) && lineContent.includes('-')) {
+          const nullNode = node.items[0];
+          if (
+            nullNode.value === null && // value is null
+            nullNode.range[0] > offset // cursor is before null
+          ) {
+            node = nullNode;
+            originalNode = node;
+            overwriteRange.end.character += nullNode.range[2] - offset; // extend range to the end of the null element
```
gorkem, line 327: I do not think this is going to be enough to insert a valid yaml. The `overwriteRange.start.character` should be adjusted so that there is no extra space between `-` and the inserted value.

gorkem (approved):
This has cases where it does not insert valid YAML

src/languageservice/services/yamlCompletion.ts
p-spacek: thanks, I'll have a look

p-spacek: Do you mean this one?

https://user-images.githubusercontent.com/38421337/162406730-3a49ed3e-1cac-41bc-8b71-a645addc038a.mov

hard to say, it looks ugly (extra indent), but the yaml is valid I think

gorkem: Yes, in my tests when I inject a single item editor marks it as invalid.

msivasubramaniaan (approved):

PR #698, opened 2022-04-06
change completion that is invoked just after array symbol

p-spacek:
### What does this PR do?
unify how codecompletion is displayed in this two situations:
 - ` - #curstor`
 - ` -#curstor`

current implementation:

https://user-images.githubusercontent.com/38421337/161982187-22b1a7c8-3e16-498d-abf4-9c1810997e45.mov

after this update:

https://user-images.githubusercontent.com/38421337/161982355-9bbd70cb-4caf-4abe-8ab8-3511af1fa4ab.mov

this PR extends:
https://github.com/redhat-developer/yaml-language-server/pull/696
https://github.com/redhat-developer/yaml-language-server/pull/697
but it's working separately 

### Is it tested? How?
modified existing unit tests
I also tested all 3 changes together. I didn't want to put it all together because it's quite different modifications.

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -71,6 +71,7 @@ export class YamlCompletion {
   private configuredIndentation: string | undefined;
   private yamlVersion: YamlVersion;
   private indentation: string;
+  private prefixIndentation = '';
   private supportsMarkdown: boolean | undefined;
   private disableDefaultProperties: boolean;
 
@@ -126,7 +127,8 @@ export class YamlCompletion {
 
     const currentWord = this.getCurrentWord(document, offset);
 
-    let overwriteRange = null;
+    this.prefixIndentation = '';
+    let overwriteRange: Range = null;
     if (node && isScalar(node) && node.value === 'null') {
       const nodeStartPos = document.positionAt(node.range[0]);
       nodeStartPos.character += 1;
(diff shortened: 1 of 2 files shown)


src/languageservice/services/yamlCompletion.ts
```diff
@@ -71,6 +71,7 @@ export class YamlCompletion {
   private configuredIndentation: string | undefined;
   private yamlVersion: YamlVersion;
   private indentation: string;
+  private prefixIndentation = '';
```
gorkem, line 74: Should this get a better name to indicate that this prefix is used for arrays only.

gorkem (approved):
One small comment, otherwise looks good.

src/languageservice/services/yamlCompletion.ts
p-spacek: I thought that this property can be used for very similar functionality in the future:
colon without space

```yaml
simpleProp:#invoke here without space
```

but I can change the name to `arrayPrefixIndentantion` and when this functionality is implemented in the future we can change it back to something more generic.

What do you think?

gorkem: Unless you are planning to provide a PR for it soon. I would prefer descriptive names.

p-spacek: Not sure when, so I renamed it.

msivasubramaniaan (approved):

PR #705, opened 2022-04-23
fix value type convert to string

dev-bz:
`value as string` in typescript is just `value` in javascript, so it's not really converted.

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -1374,7 +1374,7 @@ export class YamlCompletion {
     if (Array.isArray(value)) {
       return JSON.stringify(value);
     }
-    return value as string;
+    return '' + value;
   }
 
   private collectDefaultSnippets(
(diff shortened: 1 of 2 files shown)


dev-bz:
Hi @coveralls, I cancelled #703 and create this for a same problem, maybe you should see this commit.

gorkem:
Is there a test that you can add that demonstrates the issue fixed with this PR?

dev-bz:
> Is there a test that you can add that demonstrates the issue fixed with this PR?

here a simple code to show
```json
// filename test.schema.json
{
	"type": "object",
	"properties": {
		"test": {
			"type": ["string","boolean"],
			"enum": ["Always",true,false]
		}
	}
}
```
```yaml
# yaml-language-server: $schema=test.schema.json
test: f
#     ^ false
```
~~~

gorkem:
Will you be able to add tests for this?

gorkem (approved):
LGTM

msivasubramaniaan (approved):

PR #716, opened 2022-05-18
Enable eslint import checks

ssbarnea:
- Enable eslint plugin to check imports
- Fix current problems identified
- Add option to treat warnings as error so we avoid regressions
- Add `package-lock.json` to exclude as current projects uses yarn

That was suggested on https://github.com/redhat-developer/yaml-language-server/pull/593#issuecomment-976297139


gorkem (approved):
lgtm

msivasubramaniaan (approved):

PR #717, opened 2022-05-19
fix: object with additional properties completion

p-spacek:
### What does this PR do?
add support for completion for an object that has a definition in `additionalProperties`

```
{
  type: 'object',
  additionalProperties: {
    anyOf: [
      {
        type: 'string',
        const: 'test1',
      },
    ],
  },
};
```

typescript that produces this schema:
```ts
Record<string, 'test1' | InterfaceXYZ>
```
no ref

### Is it tested? How?
add UnitTest


msivasubramaniaan (approved):
LGTM

gorkem (approved):
Thanks for the contribution.

PR #718, opened 2022-05-24
Add hover settings

nikalexxx:
### Add hover settings

View by default
Screen Shot 2022-05-24 at 18 28 50

View with `{showTitle: false, showSource: false}`
Screen Shot 2022-05-24 at 18 29 46


gorkem:
The build is failing. We have recently merged a change to enable more checks. Can you please fix them and update the PR.

Jomik:
Couldn't we use `hover` instead of `hoverSettings` to keep in line with `format`?
```ts
{
  format = { enable = true },
  hover = { enable = true, showTitle = false },
}
```
You can "just" set the type to be `boolean | CustomHoverSettings` and then handle boolean as before.

PR #719, opened 2022-05-27
Fixed auto complete even parameters not in the specific order

msivasubramaniaan:
### What does this PR do?
This PR checks the enum values that matches started characters of node value on auto suggestions

https://github.com/redhat-developer/vscode-yaml/issues/705

### Is it tested? How?

Yes. With existing UT.
image

### DIFF
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -798,7 +798,7 @@ function validate(
       const val = getNodeValue(node);
       let enumValueMatch = false;
       for (const e of schema.enum) {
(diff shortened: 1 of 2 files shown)


gorkem (changes requested):
Looking good but do our current tests really cover the case? Wouldn't they already fail if they did?

msivasubramaniaan:
@gorkem The existing test cases were covered with proper order of schema and not with the irregular order. I will add the testcase for this case as well.

evidolob (approved):

gorkem (approved):
LGTM

PR #724, opened 2022-06-02
fix: completion invoke when next line has an indented comment

rickcowan:
### What does this PR do?
fix: completion invoke when next line has an indented comment

The change for this was to evaluate `range[1]` (value-end) rather than `range[2]` (node-end, which may contain trivia such as comments) when trying to find the closest node.

```yaml
example:
  prop1: "test"
  # cursor here
    #comment
```

no ref

### Is it tested? How?
added unit tests

### DIFF
--- a/src/languageservice/parser/yaml-documents.ts
+++ b/src/languageservice/parser/yaml-documents.ts
@@ -137,7 +137,7 @@ export class SingleYAMLDocument extends JSONDocument {
       if (!range) {
         return;
       }
-      const diff = range[2] - offset;
+      const diff = range[1] - offset;
       if (maxOffset <= range[0] && diff <= 0 && Math.abs(diff) <= offsetDiff) {
         offsetDiff = Math.abs(diff);
         maxOffset = range[0];
--- a/test/autoCompletionFix.test.ts
+++ b/test/autoCompletionFix.test.ts
@@ -554,7 +554,7 @@ objB:
             },
           },
         },
-      }
+      },
     };
 
     it('completion should handle indented comment on new line', async () => {
--- a/test/autoCompletionFix.test.ts
+++ b/test/autoCompletionFix.test.ts
@@ -540,29 +540,43 @@ objB:
   });
   
   describe('should suggest property before indented comment', () => {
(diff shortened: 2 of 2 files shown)


test/autoCompletionFix.test.ts
```diff
@@ -538,4 +538,47 @@ objB:
...
+      const completion = await parseSetup(content, 2, 2);
+      expect(completion.items.length).equal(1);
+      expect(completion.items[0]).to.be.deep.equal(
+        createExpectedCompletion('prop2', 'prop2: ', 2, 2, 2, 2, CompletionItemKind.Property, InsertTextFormat.Snippet, {
+          documentation: ''
+        })
+      );
+    });
+
+    it('completion should handle comment at same indent level on new line', async () => {
```
rickcowan, line 572: This unit test is testing functionality that was already working before the fix. Simply added as a control check.

gorkem:
Looks like it failed prettier checks. `yarn build` should have caught them. Can you update the PR with the fixes.

src/languageservice/parser/yaml-documents.ts
```diff
@@ -137,7 +137,7 @@ export class SingleYAMLDocument extends JSONDocument {
       if (!range) {
         return;
       }
-      const diff = range[2] - offset;
+      const diff = range[1] - offset;
```
msivasubramaniaan, line 140: what could happened if next line hasn't comment?

rickcowan: Tested this manually and it works. Added another unit test for this situation.

gorkem:
@p-spacek do you mind reviewing this one?

gorkem (approved):
LGTM

gorkem:
Great catch, thank you for the contribution. I think this got a conflict on the test after another contribution was merged.

p-spacek:
> @p-spacek do you mind reviewing this one?

We were cooperating together with Rick on this issue. So from my point of view, it's ok.
@rickcowan can you please check the conflict and tests after the merge?

PR #725, opened 2022-06-03
fix: suggest prop of the object -based on not completed prop name

p-spacek:
### What does this PR do?
fix completion when property name is expected in nested object
```yaml
test2: 
  pr # 1 worked ok here
test1:
  nested: 
    pr # 2 didn't work, this PR fixes this
```

this PR fixes the # 2nd scenario and add UT for the # 1st example

note that I am not sure if there could be more issues in other places where the root is modifying (`currentDoc.internalDocument.set`) instead of modifying `parent`. 
It seems that this part of the code is not under the UTs, so if I try to comment all `currentDoc.internalDocument.set...` all tests will pass. So I am afraid and won't change others because not sure what yaml is behind.

no ref

### Is it tested? How?
add UT for both examples

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -370,7 +370,7 @@ export class YamlCompletion {
                           currentDoc.internalDocument = currentDoc.internalDocument;
                         }
                       } else {
-                        currentDoc.internalDocument.set(parent.key, map);
+                        parent.value = map;
                         // eslint-disable-next-line no-self-assign
                         currentDoc.internalDocument = currentDoc.internalDocument;
                       }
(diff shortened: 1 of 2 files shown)


src/languageservice/services/yamlCompletion.ts
```diff
@@ -370,7 +370,7 @@ export class YamlCompletion {
                           currentDoc.internalDocument = currentDoc.internalDocument;
                         }
                       } else {
-                        currentDoc.internalDocument.set(parent.key, map);
+                        parent.value = map;
```
msivasubramaniaan, line 373: overriding the `parent.value` seems to be not a good solution.

msivasubramaniaan (changes requested):

msivasubramaniaan (approved):
LGTM

PR #726, opened 2022-06-07
fix: update tests to use virtual caret

grant-d:
### What does this PR do?
This is a follow-up to #722
* The previous PR introduced the concept of a _virtual caret_. However, I only updated a few tests in order to keep the noise low so the reviewer could easily see what was going on.
* This PR updates all relevant/remaining tests to use the new functionality. Note that some tests are not updated, for example when they specify an index that is out of range (see review comments)

### Is it tested? How?
Unit tests continue to pass.

I double-checked all `position` values one-by-one in this PR, by running `test.only()` for each edited test, and ensuring the old `position` matched the new. This was an onerous task, but the benefit is that we can be sure the old and new units are testing exactly the same thing. Please see the review notes for more detail.

### DIFF
--- a/test/autoCompletion.test.ts
+++ b/test/autoCompletion.test.ts
@@ -45,7 +45,9 @@ describe('Auto Completion Tests', () => {
     if (typeof position === 'undefined') {
       position = content.indexOf('|:|');
       content = content.replace('|:|', '');
+      if (content.length) position--;
     }
+    // console.log('position:', position, '>' + content.substring(position) + '<');
 
     const testTextDocument = setupSchemaIDTextDocument(content);
     yamlSettings.documents = new TextDocumentTestManager();
@@ -192,7 +194,7 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'nam|:|e';
+        const content = 'name|:|';
         const completion = await parseSetup(content);
         assert.strictEqual(completion.items.length, 1);
         assert.deepStrictEqual(
@@ -353,8 +355,8 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'scripts:\n  sample';
-        const completion = parseSetup(content, 11);
+        const content = 'scripts:\n  s|:|ample';
+        const completion = parseSetup(content);
         completion
           .then(function (result) {
             assert.equal(result.items.length, 1);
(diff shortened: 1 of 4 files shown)


test/autoCompletion.test.ts
```diff
@@ -45,7 +45,9 @@ describe('Auto Completion Tests', () => {
     if (typeof position === 'undefined') {
       position = content.indexOf('|:|');
       content = content.replace('|:|', '');
+      if (content.length) position--;
```
grant-d, line 48: I made a calculation error in the initial PR. With the exacting checks I made (per the description of this PR), I corrected it here.

test/autoCompletion.test.ts
```diff
@@ -558,8 +560,8 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'scripts: |:|';
-        const completion = parseSetup(content);
+        const content = 'scripts: ';
+        const completion = parseSetup(content, content.length);
```
grant-d, line 564: I have no way of being true to the original test here, since `content[content.length]` is (intentionally it would seem) out of bounds.

I can either leave it as-is, or I can perhaps change the content to:
`scripts: |:|` which I believe would be the moral equivalent?

test/defaultSnippets.test.ts
```diff
@@ -28,7 +28,14 @@ describe('Default Snippet Tests', () => {
...
 
   describe('Snippet Tests', function () {
-    function parseSetup(content: string, position: number): Promise<CompletionList> {
+    function parseSetup(content: string, position?: number): Promise<CompletionList> {
+      if (typeof position === 'undefined') {
+        position = content.indexOf('|:|');
+        content = content.replace('|:|', '');
+        if (content.length) position--;
+      }
+      // console.log('position:', position, '>' + content.substring(position) + '<');
```
grant-d, line 37: This is how I checked that the `position` of the old and new tests was exactly the same. For example, if the original test called `parseSetup(content, 14)` then I made sure this log output `14` when running the new test.

test/hover.test.ts
```diff
@@ -174,7 +180,8 @@ describe('Hover Tests', () => {
         `Contains custom hooks used to trigger other automated tools\n\nSource: [${SCHEMA_ID}](file:///${SCHEMA_ID})`
       );
 
-      const secondHover = await parseSetup(content, 15);
+      const content2 = 'scripts:\n  posti|:|nstall: test';
+      const secondHover = await parseSetup(content2);
```
grant-d, line 184: Had to duplicate the content for this test, in order to 'move' the cursor.

PR #727, opened 2022-06-07
fix: update tests to use virtual caret

grant-d:
### What does this PR do?
This is a follow-up to #722
* The previous PR introduced the concept of a _virtual caret_. However, I only updated a few tests in order to keep the noise low so the reviewer could easily see what was going on.
* This PR updates all relevant/remaining tests to use the new functionality. Note that some tests are not updated, for example when they specify an index that is out of range (eg `parseSetup('abc', 10)`)

### Improvements
I improved the caret usage/algorithm to use "bookends" instead of absolute position. The previous method made it hard to specify some boundary cases.

- We now use caret "bookends" which surround the intended character, for example:
  - `'a|b|c'` is the same as `position = 1`, ie the caret is on the `'b'` char
  - `'ab| |'` is the same as `position = 2`, ie the caret is on the `' '` char
  - `'|\n|bc'` is the same as `position = 0`, but this one is interesting because the caret is on the `'\n'` char

### Algorithm
Here's how the `content` value is manipulated in order to derive the position given an embedded caret:
- Using an example: the original string: `content = '0123456789'`
- Imagine we want the caret on `'7'`, thus we specify: `'0123456|7|89'` instead
- Now calling `position = content.search(regex)` has the result `position === 7`
  - The regex looks for `|`, followed by `any one char including \n`, followed by `|`
- We then elide the caret "bookends", which has the result `content = '0123456789'`
- Now `content[position: 7] === '7'` which is what we originally intended with our caret placement

### Is it tested? How?
Unit tests continue to pass.

I double-checked all `position` values one-by-one in this PR, by running `test.only()` for each edited test, and ensuring the old `position` matched the new. This was an onerous task, but the benefit is that we can be sure the old and new units are testing exactly the same thing. Please see the review notes for more detail.

### DIFF
--- a/test/autoCompletion.test.ts
+++ b/test/autoCompletion.test.ts
@@ -42,10 +42,12 @@ describe('Auto Completion Tests', () => {
    * @returns A list of valid completions.
    */
   function parseSetup(content: string, position?: number): Promise<CompletionList> {
+    // console.log('original:', content.length, content, '>' + content.substring(position) + '<');
     if (typeof position === 'undefined') {
-      position = content.indexOf('|:|');
-      content = content.replace('|:|', '');
+      position = content.search(/\|[^]\|/); // | -> any char including newline -> |
+      content = content.substring(0, position) + content.substring(position + 1, position + 2) + content.substring(position + 3);
     }
+    // console.log('position:', position, content, '>' + content.substring(position) + '<');
 
     const testTextDocument = setupSchemaIDTextDocument(content);
     yamlSettings.documents = new TextDocumentTestManager();
@@ -72,8 +74,8 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = '|:|';
-        const completion = parseSetup(content);
+        const content = '';
+        const completion = parseSetup(content, 0);
         completion
           .then(function (result) {
             assert.equal(result.items.length, 1);
@@ -96,8 +98,8 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'na|:|';
-        const completion = parseSetup(content);
+        const content = 'na'; // len: 2
+        const completion = parseSetup(content, 2);
         completion
           .then(function (result) {
             assert.equal(result.items.length, 1);
@@ -121,7 +123,7 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'name';
+        const content = 'name'; // len: 4
         const completion = parseSetup(content, 10);
         completion
           .then(function (result) {
@@ -146,7 +148,7 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'name: ';
+        const content = 'name: '; // len: 6
         const completion = parseSetup(content, 12);
         completion
           .then(function (result) {
@@ -171,7 +173,7 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'name: ';
+        const content = 'name: '; // len: 6
         const completion = await parseSetup(content, 6);
         assert.strictEqual(completion.items.length, 1);
         assert.deepStrictEqual(
@@ -192,8 +194,8 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'nam|:|e';
-        const completion = await parseSetup(content);
+        const content = 'nam|e|'; // len: 4
+        const completion = await parseSetup(content); // 3
         assert.strictEqual(completion.items.length, 1);
         assert.deepStrictEqual(
           completion.items[0],
@@ -213,7 +215,7 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'name: ya';
+        const content = 'name: ya'; // len: 8
         const completion = parseSetup(content, 15);
         completion
           .then(function (result) {
@@ -237,7 +239,7 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'yaml: ';
+        const content = 'yaml: '; // len: 6
         const completion = parseSetup(content, 11);
         completion
           .then(function (result) {
@@ -267,7 +269,7 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'yaml: fal';
+        const content = 'yaml: fal'; // len: 9
         const completion = parseSetup(content, 11);
         completion
           .then(function (result) {
@@ -298,7 +300,7 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'timeout: ';
+        const content = 'timeout: '; // len: 9
         const completion = parseSetup(content, 9);
         completion
           .then(function (result) {
@@ -323,7 +325,7 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'timeout: 6';
+        const content = 'timeout: 6'; // len: 10
         const completion = parseSetup(content, 10);
         completion
           .then(function (result) {
@@ -353,8 +355,8 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'scripts:\n  sample';
-        const completion = parseSetup(content, 11);
+        const content = 'scripts:\n  |s|ample'; // len: 17, pos: 11
+        const completion = parseSetup(content);
         completion
           .then(function (result) {
             assert.equal(result.items.length, 1);
@@ -383,8 +385,8 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'scripts:\n  sam';
-        const completion = parseSetup(content, 11);
+        const content = 'scripts:\n  |s|am'; // len: 14, pos: 11
+        const completion = parseSetup(content);
         completion
           .then(function (result) {
             assert.equal(result.items.length, 1);
@@ -479,8 +481,8 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'scripts:\n  sample: test\n  myOther';
-        const completion = parseSetup(content, 31);
+        const content = 'scripts:\n  sample: test\n  myOth|e|r'; // len: 33, pos: 31
+        const completion = parseSetup(content);
         completion
           .then(function (result) {
             assert.equal(result.items.length, 1);
@@ -504,7 +506,7 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'timeout:';
+        const content = 'timeout:'; // len: 8
         const completion = parseSetup(content, 9);
         completion
           .then(function (result) {
@@ -532,7 +534,7 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'scripts:\n  sample:';
+        const content = 'scripts:\n  sample:'; // len: 18
         const completion = parseSetup(content, 21);
         completion
           .then(function (result) {
@@ -558,8 +560,8 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'scripts: |:|';
-        const completion = parseSetup(content);
+        const content = 'scripts: ';
+        const completion = parseSetup(content, content.length);
         completion
           .then(function (result) {
             assert.equal(result.items.length, 1);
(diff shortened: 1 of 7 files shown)


test/autoCompletion.test.ts
```diff
@@ -42,10 +42,12 @@ describe('Auto Completion Tests', () => {
    * @returns A list of valid completions.
    */
   function parseSetup(content: string, position?: number): Promise<CompletionList> {
+    // console.log('original:', content.length, content, '>' + content.substring(position) + '<');
     if (typeof position === 'undefined') {
-      position = content.indexOf('|:|');
-      content = content.replace('|:|', '');
+      position = content.search(/\|[^]\|/); // | -> any char including newline -> |
+      content = content.substring(0, position) + content.substring(position + 1, position + 2) + content.substring(position + 3);
```
grant-d, line 48: See the main description for details on this algorithm

test/autoCompletion.test.ts
```diff
@@ -96,8 +98,8 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'na|:|';
-        const completion = parseSetup(content);
+        const content = 'na'; // len: 2
+        const completion = parseSetup(content, 2);
```
grant-d, line 102: I have added a comment to every `content` string stating its `length`
This is so that the maintainer has a quick understanding of the `position` value in the next call.
In this case, `position === 2` which is beyond the `content.length`
So we cannot use the new algorithm, so we leave the old one in place.

test/autoCompletion.test.ts
```diff
@@ -479,8 +481,8 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = 'scripts:\n  sample: test\n  myOther';
-        const completion = parseSetup(content, 31);
+        const content = 'scripts:\n  sample: test\n  myOth|e|r'; // len: 33, pos: 31
+        const completion = parseSetup(content);
```
grant-d, line 485: Where I have changed to the new semantic, I add a comment with the original string length, and the caret position.
Once again, this is just for the maintainer's benefit, so they can quickly understand the test's intent

test/autoCompletion.test.ts
```diff
@@ -72,8 +74,8 @@ describe('Auto Completion Tests', () => {
             },
           },
         });
-        const content = '|:|';
-        const completion = parseSetup(content);
+        const content = '';
+        const completion = parseSetup(content, 0);
```
grant-d, line 78: Reverted this earlier change - a caret cannot be placed in a zero-length string, so use the old algorithm

test/multipleDocuments.test.ts
```diff
@@ -54,7 +54,7 @@ describe('Multiple Documents Validation Tests', () => {
       return validationHandler.validateTextDocument(testTextDocument);
     }
 
-    function hoverSetup(content: string, position): Promise<Hover> {
+    function hoverSetup(content: string, position: number): Promise<Hover> {
```
grant-d, line 57: Being a good citizen

test/autoCompletionFix.test.ts
```diff
@@ -50,6 +50,28 @@ describe('Auto Completion Fix Tests', () => {
     });
   }
 
+  /**
+   * Generates a completion list for the given document and caret (cursor) position.
+   * @param content The content of the document.
+   * Alternatively, `position` can be omitted if the caret is located in the content using `|` bookends.
```
gorkem, line 56: Looks like a copy/paste left-over. This function is carets only, this probably needs to be modified to match.

test/autoCompletionFix.test.ts
```diff
@@ -50,6 +50,28 @@ describe('Auto Completion Fix Tests', () => {
...
+   * For example, `content = 'ab|c|d'` places the caret over the `'c'`, at `position = 2`
+   * @returns A list of valid completions.
+   */
+  function parseCaret(content: string): Promise<CompletionList> {
+    // console.log(`was: len: ${content.length}, content: "${content}"`);
+    const position = content.search(/\|[^]\|/); // | -> any char including newline -> |
+    content = content.substring(0, position) + content.substring(position + 1, position + 2) + content.substring(position + 3);
```
gorkem, line 63: How about throwing an error if the caret is not found?

test/defaultSnippets.test.ts
```diff
@@ -28,7 +28,23 @@ describe('Default Snippet Tests', () => {
...
+     * @param content The content of the document.
+     * @param position The position of the caret in the document.
+     * Alternatively, `position` can be omitted if the caret is located in the content using `|` bookends.
+     * For example, `content = 'ab|c|d'` places the caret over the `'c'`, at `position = 2`
+     * @returns A list of valid completions.
+     */
+    function parseSetup(content: string, position?: number): Promise<CompletionList> {
+      // console.log(`was: len: ${content.length}, content: "${content}", str: "${content.substring(position)}"`);
+      if (typeof position === 'undefined') {
+        position = content.search(/\|[^]\|/); // | -> any char including newline -> |
```
gorkem, line 42: How about throwing an error of caret is not found. Perhaps move this 2 lines to its own function on test\utils\testHelper.ts?

test/hover.test.ts
```diff
@@ -40,7 +40,22 @@ describe('Hover Tests', () => {
...
+   * @param position The position of the caret in the document.
+   * Alternatively, `position` can be omitted if the caret is located in the content using `|` bookends.
+   * For example, `content = 'ab|c|d'` places the caret over the `'c'`, at `position = 2`
+   * @returns An instance of `Hover`.
+   */
+  function parseSetup(content: string, position?: number): Promise<Hover> {
+    // console.log(`was: len: ${content.length}, content: "${content}", str: "${content.substring(position)}"`);
+    if (typeof position === 'undefined') {
+      position = content.search(/\|[^]\|/); // | -> any char including newline -> |
+      content = content.substring(0, position) + content.substring(position + 1, position + 2) + content.substring(position + 3);
```
gorkem, line 55: How about moving the above 2 lines to its own function in `test\utils\testHelper.ts`? Also as with others, what to do if caret is not found.

gorkem (changes requested):
This is a great PR. Just a couple of comments, one to move the repeated code to utils and covering the case when the caret is misused. 

I really appreciate the added comments.

test/autoCompletion.test.ts
```diff
@@ -43,12 +43,9 @@ describe('Auto Completion Tests', () => {
    * @returns A list of valid completions.
    */
   function parseSetup(content: string, position?: number): Promise<CompletionList> {
-    // console.log(`was: len: ${content.length}, content: "${content}", str: "${content.substring(position)}"`);
     if (typeof position === 'undefined') {
-      position = content.search(/\|[^]\|/); // | -> any char including newline -> |
-      content = content.substring(0, position) + content.substring(position + 1, position + 2) + content.substring(position + 3);
+      ({ content, position } = caretPosition(content));
     }
-    // console.log(`now: len: ${content.length}, content: "${content}", pos: ${position}, str: "${content.substring(position)}"`);
```
grant-d: Moved to utility function as requested

test/autoCompletionFix.test.ts
```diff
@@ -53,17 +53,14 @@ describe('Auto Completion Fix Tests', () => {
...
-   * Alternatively, `position` can be omitted if the caret is located in the content using `|` bookends.
+   * The caret is located in the content using `|` bookends.
    * For example, `content = 'ab|c|d'` places the caret over the `'c'`, at `position = 2`
    * @returns A list of valid completions.
    */
   function parseCaret(content: string): Promise<CompletionList> {
-    // console.log(`was: len: ${content.length}, content: "${content}"`);
-    const position = content.search(/\|[^]\|/); // | -> any char including newline -> |
-    content = content.substring(0, position) + content.substring(position + 1, position + 2) + content.substring(position + 3);
-    // console.log(`now: len: ${content.length}, content: "${content}", pos: ${position}, str: "${content.substring(position)}"`);
```
grant-d: Moved to utility function as requested

test/defaultSnippets.test.ts
```diff
@@ -37,13 +37,9 @@ describe('Default Snippet Tests', () => {
...
      */
     function parseSetup(content: string, position?: number): Promise<CompletionList> {
-      // console.log(`was: len: ${content.length}, content: "${content}", str: "${content.substring(position)}"`);
       if (typeof position === 'undefined') {
-        position = content.search(/\|[^]\|/); // | -> any char including newline -> |
-        content =
-          content.substring(0, position) + content.substring(position + 1, position + 2) + content.substring(position + 3);
+        ({ content, position } = caretPosition(content));
       }
-      // console.log(`now: len: ${content.length}, content: "${content}", pos: ${position}, str: "${content.substring(position)}"`);
```
grant-d: Moved to utility function as requested

test/hover.test.ts
```diff
@@ -49,12 +49,9 @@ describe('Hover Tests', () => {
    * @returns An instance of `Hover`.
    */
   function parseSetup(content: string, position?: number): Promise<Hover> {
-    // console.log(`was: len: ${content.length}, content: "${content}", str: "${content.substring(position)}"`);
     if (typeof position === 'undefined') {
-      position = content.search(/\|[^]\|/); // | -> any char including newline -> |
-      content = content.substring(0, position) + content.substring(position + 1, position + 2) + content.substring(position + 3);
+      ({ content, position } = caretPosition(content));
     }
-    // console.log(`now: len: ${content.length}, content: "${content}", pos: ${position}, str: "${content.substring(position)}"`);
```
grant-d: Moved to utility function as requested

test/autoCompletion.test.ts
```diff
@@ -38,14 +38,17 @@ describe('Auto Completion Tests', () => {
...
+   * Alternatively, `position` can be omitted if the caret is located in the content using `|` bookends.
+   * For example, `content = 'ab|c|d'` places the caret over the `'c'`, at `position = 2`
    * @returns A list of valid completions.
    */
   function parseSetup(content: string, position?: number): Promise<CompletionList> {
+    // console.log(`was: len: ${content.length}, content: "${content}", str: "${content.substring(position)}"`);
     if (typeof position === 'undefined') {
-      position = content.indexOf('|:|');
-      content = content.replace('|:|', '');
+      position = content.search(/\|[^]\|/); // | -> any char including newline -> |
```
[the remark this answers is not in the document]
grant-d: Good idea - done (see new utility function)

test/autoCompletionFix.test.ts
grant-d: Oops - fixed

test/defaultSnippets.test.ts
grant-d: Done

test/hover.test.ts
grant-d: Done

test/utils/testHelper.ts
```diff
@@ -93,3 +93,24 @@ export function setupLanguageService(languageSettings: LanguageSettings): TestLa
...
 }
+
+/**
+ * Derives the absolute `position` of the caret given `content` containing a virtual caret.
+ * @param content The content of the document.
+ * The caret is located in the content using `|` bookends.
+ * For example, `content = 'ab|c|d'` places the caret over the `'c'`, at `position = 2`
+ * @returns The absolute position of the caret.
+ */
+export function caretPosition(content: string): { position: number; content: string } {
```
grant-d, line 104: Good idea, thanks. New function in `utils`

test/utils/testHelper.ts
```diff
@@ -93,3 +93,24 @@ export function setupLanguageService(languageSettings: LanguageSettings): TestLa
...
+  // console.log(`was: len: ${content.length}, content: "${content}", str: "${content.substring(position)}"`);
+
+  // Find bookends `|.|` in content
+  const position = content.search(/\|[^]\|/); // | -> any char including newline -> |
+  if (position === -1) throw new Error('Error in test case: no caret found in content');
```
grant-d, line 109: New `Error` as suggested

test/autoCompletionFix.test.ts
grant-d: Done

gorkem (approved):
Thank you for your contribution. 
@msivasubramaniaan this should go in first as it will probably affect some of the other PRs.

grant-d:
Thanks @gorkem

msivasubramaniaan (approved):

Issue #709, opened 2022-04-29
Auto-complete on enum strings do not get string quoting


ssbarnea:
## Describe the bug

When autocompleting with an enum value that is a string that can is also a number it inserts the number instead of adding quotes to it.

Example https://github.com/ansible/schemas/blob/main/f/ansible-meta.json#L1051 -- with `"enum": ["12.1", "13.0", "13.1", "14.0", "all"],`. 

A correct implementation should insert `"12.1"` string instead of `12.1` value, which would a float (or integer). 

Not including the quotes is a bug, as schemas might not allow non-string values.

![](https://sbarnea.com/ss/Screen-Shot-2022-04-29-15-41-41.03.png)

## Expected Behavior

Add quotes when needed.

## Current Behavior

The unquoted string is inserted and this would be loaded as a different value.

This affects anything that looks like a **valid float or int in YAML**.

## Environment
- [x] Windows
- [x] Mac
- [x] Linux

PR #728, opened 2022-06-08
Fix auto complete enum string with quote on number

msivasubramaniaan:
### What does this PR do?

This PR adding double quotes to the number when enum type is string and it has number variable on it. 
Ex: "enum": ["12.1", "13.0", "13.1", "14.0", "all"]

https://github.com/redhat-developer/yaml-language-server/issues/709

### Is it tested? How?
Yes, added UT

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -29,7 +29,7 @@ import { YAMLSchemaService } from './yamlSchemaService';
 import { ResolvedSchema } from 'vscode-json-languageservice/lib/umd/services/jsonSchemaService';
 import { JSONSchema, JSONSchemaRef } from '../jsonSchema';
 import { stringifyObject, StringifySettings } from '../utils/json';
-import { convertErrorToTelemetryMsg, isDefined, isNumber, isString } from '../utils/objects';
+import { convertErrorToTelemetryMsg, isDefined, isString } from '../utils/objects';
 import * as nls from 'vscode-nls';
 import { setKubernetesParserOption } from '../parser/isKubernetes';
 import { asSchema } from '../parser/jsonParser07';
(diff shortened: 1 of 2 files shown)


gorkem (changes requested):
This is not working as I expected. I used the following schema and all the completions were inserted with quotes. Looks like the insertText logic needs to be improved. 

```Json
{
  "type": "object",
  "additionalProperties": {
    "anyOf": [
      {
        "type": "array",
        "items": {
          "enum": ["12.1", 33, "34", 78],
          "type":  ["string", "integer"]
        }
      }
    ]
  }
} 
```

msivasubramaniaan:
> This is not working as I expected. I used the following schema and all the completions were inserted with quotes. Looks like the insertText logic needs to be improved.
> 
> ```json
> {
>   "type": "object",
>   "additionalProperties": {
>     "anyOf": [
>       {
>         "type": "array",
>         "items": {
>           "enum": ["12.1", 33, "34", 78],
>           "type":  ["string", "integer"]
>         }
>       }
>     ]
>   }
> } 
> ```

changed logic when the value type is integer and updated UT accordingly

src/languageservice/services/yamlCompletion.ts
```diff
@@ -1209,8 +1209,11 @@ export class YamlCompletion {
         return this.getInsertTemplateForValue(value, indent, { index: 1 }, separatorAfter);
       }
     }
-    type = Array.isArray(type) ? type[0] : type;
-    if (type === 'string') {
+    const types = Array.isArray(type) ? type : type;
```
gorkem, line 1212: What does this line try to achieve. It looks like it is assigning the same value to `types` regardless whether `type` is array or not.

gorkem (changes requested):
Should we check if this is handling all possible schema types correctly. Possible types are `string, number, integer, object, array, boolean, null`.

src/languageservice/services/yamlCompletion.ts
msivasubramaniaan: Added all the test cases for **number, string, Boolean, null and object**. 

> Also this method called when arrays having ENUM. So I have a question here: Is arrays of ENUM having another array in YAML? if so please provide me a reference schema for that as well. I will update my UT based on that.

evidolob (approved):

PR #738, opened 2022-06-23
Check if value is string before calling method

miarodri:
### What does this PR do?
Check if an enum's values are strings before attempting to determine if they start with the value being validated.

[Draft 7 section 6.1.2 of the JSON Schema Specification](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-6.1.2) notes:

> Elements in the array might be of any value, including null.

As the parser assumes string values, if the value being validated does not match the first enum value specified, then the parser throws a type error (as seen in the error message provided by the `vscode-yaml` extension):

```txt
[2022-06-23 12:35:24.777] [exthost] [error] Error: Request textDocument/hover failed with message: i.startsWith is not a function
```

### Is it tested? How?
Built and tested with the `vscode-yaml` extension. Running the language server and attaching the extension reproduces the error when trying to get the hover text for a property that adheres to a schema with an enum with integer values.

For example, using the `vscode-yaml` extension for convenience:

test.schema.json
```json
{
    "$schema": "https://json-schema.org/draft-07/schema",
    "type": "object",
    "properties": {
        "test": {
            "enum": [
                null,
                1,
                2
            ]
        }
    }
}
```

test.yml
```yml
# yaml-language-server: $schema=test.schema.json
test: 2
```

Hovering over line 2 (`test: 2`) causes an error to appear in the Extension Host log. After the fix, no error is logged.

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -138,7 +138,8 @@ export class YamlCompletion {
     if (areOnlySpacesAfterPosition) {
       overwriteRange = Range.create(position, Position.create(position.line, lineContent.length));
       const isOnlyWhitespace = lineContent.trim().length === 0;
-      if (node && isScalar(node) && !isOnlyWhitespace) {
+      const isOnlyDash = lineContent.match(/^\s*(-)\s*$/);
+      if (node && isScalar(node) && !isOnlyWhitespace && !isOnlyDash) {
         // line contains part of a key with trailing spaces, adjust the overwrite range to include only the text
         const matches = lineContent.match(/^([\s-]*)[^:]+[ \t]+\n?$/);
         if (matches?.length) {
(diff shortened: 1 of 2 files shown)


src/languageservice/parser/jsonParser07.ts
```diff
@@ -797,7 +797,7 @@ function validate(
       const val = getNodeValue(node);
       let enumValueMatch = false;
       for (const e of schema.enum) {
-        if (equals(val, e) || (node.value && e.startsWith(val))) {
+        if (equals(val, e) || (node.value && (typeof e === 'string' || e instanceof String) && e.startsWith(val))) {
```
gorkem, line 800: Would it be better if we imported and used `isString` from `../utils/objects.ts`?

gorkem (changes requested):
One minor comment. Can we also convert the test case given on the PR description to a test case on `test/schemaValidation.test`.

miarodri:
As a note, I _am_ working on this. Turns out, I can't seem to get a failing test for this pre-fix. I'm investigating a little closer about why I see the bad behavior when using the YAML plugin in VSCode.

PR #740, opened 2022-06-29
Add JSON Schema validation

evidolob:
### What does this PR do?
It add JSON Schema validation with [ajv](https://ajv.js.org/) lib.
It allows to show errors in scheme instead of wrong errors during yaml validation, like https://github.com/redhat-developer/vscode-yaml/issues/788
Screenshot 2022-06-29 at 11 38 08

> Note: it also upgrades TS version, as latest `ajv` requires newer TS version

Fixes: https://github.com/redhat-developer/vscode-yaml/issues/788

### Is it tested? How?
with tests

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -29,6 +29,7 @@ import { JSONSchemaDescriptionExt } from '../../requestTypes';
 import { SchemaVersions } from '../yamlTypes';
 
 import Ajv, { DefinedError } from 'ajv';
+import { getSchemaTitle } from '../utils/schemaUtils';
 
 const localize = nls.loadMessageBundle();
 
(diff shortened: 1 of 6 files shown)


msivasubramaniaan (approved):
LGTM

src/languageservice/services/yamlSchemaService.ts
```diff
@@ -152,9 +161,19 @@ export class YAMLSchemaService extends JSONSchemaService {
...
+    let schema: JSONSchema = schemaToResolve.schema;
     const contextService = this.contextService;
 
+    if (!schema07Validator(schema)) {
+      const errs: string[] = [];
+      for (const err of schema07Validator.errors as DefinedError[]) {
+        errs.push(`${err.instancePath} : ${err.message}`);
+      }
+      resolveErrors.push(
+        `Schema '${schemaToResolve.schema.title ?? schemaToResolve.schema.url}' is not valid:\n${errs.join('\n')}`
```
gorkem, line 173: This is not sufficient to find a friendly name for the schema. When I tried this with a schema referenced locally with a modeline the schema is resolved as `undefined`. Can we use the same mechanics that is used on the Code lens for Jump to schema code lens.

gorkem (changes requested):

PR #732, opened 2022-06-09
fix: extra space after position

p-spacek:
### What does this PR do?
fix another situation with extra space after the current cursor.
My previous fixes covered only some of the possible situations (https://github.com/redhat-developer/yaml-language-server/pull/696, https://github.com/redhat-developer/yaml-language-server/pull/697)

there were still problems in this type of yaml (2nd position)

```yaml
array:
  - prop1: val1
     prop2: #__ (#cursor, _space)
```

```yaml
obj:
  prop1: val1
  prop2: #__ (#cursor, _space)
```

This fix will solve this situation (for object, arrays in one place) by finding the correct node inside `getNodeFromPosition`, instead of fixing it later and trying to find a better node (null node).

Another part of this PR is to overwrite range - remove extra space 

### What issues does this PR fix or reference?
no ref

### Is it tested? How?
modified and added unit tests

PR #742, opened 2022-06-30
fix: partial key with trailing spaces + PR#732 changes

p-spacek:
### What does this PR do?
Overwrite the range if the current line has trailing spaces.
this PR is based on PR #732 
  - that one fixed values with trailing spaces
but this one fixes properties with trailing spaces

note: it will be better to finish #732 and then this one, but it should be ok to just take this PR, because it includes all from 732.

```yaml
object:
  na#__ (#cursor, _space)
```
gives wrong code-completion range

no ref

### Is it tested? How?
added unit tests


evidolob (approved):
LGTM

msivasubramaniaan (approved):

PR #745, opened 2022-07-01
fix: proper indenting of snippet within an array

rickcowan:
### What does this PR do?
When choosing a snippet to auto-complete within an array, the indentation level was wrong on the lines below the insertion point.

```yaml
example:
  - #cursor here
```
would auto-complete to:
```yaml
example:
  - item1: "some item"
  item2: "another item"
```
rather than:
```yaml
example:
  - item1: "some item"
    item2: "another item"
```

no ref

### Is it tested? How?
Manual tests as well as new unit test.


gorkem (approved):
Approved. There is a typo on the commit message "identing" instead of "indenting". You can optionally fix that.

PR #746, opened 2022-07-06
Improve folding

evidolob:
Resolve [#763](https://github.com/redhat-developer/vscode-yaml/issues/763)

### Is it tested? How?
with tests


gorkem (changes requested):
The logic to find the comment should also look at the indentation.  For instance in the following example the comment 4 is folded with `dropins` node but it's indented in a way to indicate it is related to the next node. 

```YAML
systemd:
  units:
    - contents: ff
    ### comment 1
    - dropins:
        - name: dfffgg
        # comment 2
        # comment 3
# cxomment 4
boot_device:
```

gorkem:
Closing as there is no bandwidth to continue at this time

Issue #747, opened 2022-07-08
TypeError: i.startsWith is not a function


fbricon:
## Describe the bug
Telemetry data shows a high number (>50% of all errors) of "TypeError: i.startsWith is not a function" errors on `textDocument/hover` and completion requests apparently

## Expected Behavior
No error

## Current Behavior
Errors are thrown:
```
TypeError: i.startsWith is not a function
    at /Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.8.0/dist/languageserver.js:2:533672
    at T (/Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.8.0/dist/languageserver.js:2:534665)
    at /Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.8.0/dist/languageserver.js:2:523312
    at T (/Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.8.0/dist/languageserver.js:2:525873)
    at f.getMatchingSchemas (/Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.8.0/dist/languageserver.js:2:521654)
    at t.YamlCompletion.addPropertyCompletions (/Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.8.0/dist/languageserver.js:2:560153)
    at t.YamlCompletion.&lt;anonymous&gt; (/Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.8.0/dist/languageserver.js:2:558169)
    at Generator.next (&lt;anonymous&gt;)
    at s (/Users/_username_/.vscode/extensions/redhat.vscode-yaml-1.8.0/dist/languageserver.js:2:550676)
```

## Steps to Reproduce
No idea

fbricon:
This error started appearing with vscode-yaml 1.8.0

msivasubramaniaan:
This issue was resolved as part of the below ticket
https://github.com/redhat-developer/vscode-yaml/issues/790

PR #750, opened 2022-07-20
check string type or not for value as well

msivasubramaniaan:
### What does this PR do?
This PR resolves the `i.startsWith is not a function` issue in telemetry
https://github.com/redhat-developer/yaml-language-server/issues/747

### Is it tested? How?
Yes with existing UT and added UT for current scenario.


gorkem (approved):
LGTM

PR #757, opened 2022-07-29
Fix enum validation

msivasubramaniaan:
### What does this PR do?
Added `doComplete `flag and sent it as `true` while doing the schema validation. 

https://github.com/redhat-developer/vscode-yaml/issues/803
https://github.com/redhat-developer/vscode-yaml/issues/705

### Is it tested? How?
Yes with existing UT and added the UT for current scenario as well


src/languageservice/parser/jsonParser07.ts
```diff
@@ -580,12 +580,18 @@ export class JSONDocument {
     return null;
   }
 
-  public getMatchingSchemas(schema: JSONSchema, focusOffset = -1, exclude: ASTNode = null): IApplicableSchema[] {
+  public getMatchingSchemas(
```
gorkem, line 583: Can you add documentation for this function. It is very hard to understand what `doComplete` intends to do?

gorkem (changes requested):
It is not very easy to understand what the new flag means. Can you add documentation or think of better naming?

gorkem (approved):
LGTM

PR #758, opened 2022-08-01
fix: completion nested array item to flat structure

p-spacek:
### What does this PR do?
 - fix array nested structure
    - the previous implementation didn't go deeper into nested schemas (nested anyof)
 - unify array item value completion
    - there were three almost identical implementation 

schema:
```json
"test_array_nested_anyOf": {
      "items": {
        "anyOf": [
          {
            "anyOf": [
              {
                "$ref": "#/definitions/obj1"
              },
              {
                "$ref": "#/definitions/obj2"
              }
            ]
          },
          {
            "$ref": "#/definitions/obj3"
          }
        ]
      },
      "type": "array"
    }
```

#### Previous
```yaml
test_array_nested_anyOf:
  # invoke here
```
gives only two items:
 1st for `anyOf`
 2nd for `obj3`

image

#### After
The nested array item structure is flattened
image

note that without `- obj1: null` in screenshots YLS doesn't know that we are in an array and won't give any results

no ref

### Is it tested? How?
add UT and modify existing

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -35,7 +35,7 @@ import { setKubernetesParserOption } from '../parser/isKubernetes';
 import { asSchema } from '../parser/jsonParser07';
 import { indexOf, isInComment, isMapContainsEmptyPair } from '../utils/astUtils';
 import { isModeline } from './modelineUtil';
-import { getSchemaTypeName } from '../utils/schemaUtils';
+import { getSchemaTypeName, isAnyOfAllOfOneOfType, isPrimitiveType } from '../utils/schemaUtils';
 import { YamlNode } from '../jsonASTTypes';
 
 const localize = nls.loadMessageBundle();
@@ -58,6 +58,7 @@ interface CompletionsCollector {
   error(message: string): void;
   log(message: string): void;
   getNumberOfProposals(): number;
+  result: CompletionList;
 }
 
 interface InsertText {
@@ -277,6 +278,7 @@ export class YamlCompletion {
       getNumberOfProposals: () => {
         return result.items.length;
       },
+      result,
     };
 
     if (this.customTags.length > 0) {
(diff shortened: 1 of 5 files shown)


src/languageservice/services/yamlCompletion.ts
```diff
@@ -58,6 +58,7 @@ interface CompletionsCollector {
   error(message: string): void;
   log(message: string): void;
   getNumberOfProposals(): number;
+  result: CompletionList;
```
p-spacek, line 61: exposing the result help a lot during debugging

src/languageservice/services/yamlCompletion.ts
```diff
@@ -1251,25 +1237,32 @@ export class YamlCompletion {
...
     isArray?: boolean
   ): void {
     if (typeof schema === 'object') {
       this.addEnumValueCompletions(schema, separatorAfter, collector, isArray);
       this.addDefaultValueCompletions(schema, separatorAfter, collector);
       this.collectTypes(schema, types);
+
+      if (isArray && completionType === 'value' && !isAnyOfAllOfOneOfType(schema)) {
+        // add array only for final types (no anyOf, allOf, oneOf)
+        this.addArrayItemValueCompletion(schema, separatorAfter, collector);
```
p-spacek, line 1250: moving array item value completion into this `addSchemaValueCompletions` has two advantages:
 - we can remove 3 duplicity implementation
 - we can use recursion to `flat` the schema structure

test/fixtures/testArrayCompletionSchema.json
```diff
@@ -1,85 +1,110 @@
...
-        "properties": {
-          "obj1": {
-            "type": "object"
-          }
-        },
-        "required": [
-          "obj1"
-        ],
-        "type": "object"
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
p-spacek, line 2: the changes aren't very readable inside the github diff tool:
 - I just added a new test schema `test_array_nested_anyOf`
 - add `obj3` into the `definitions`
 - autoformat file (there were two extra spaces)

src/languageservice/utils/schemaUtils.ts
```diff
@@ -55,3 +55,11 @@ export function getSchemaTitle(schema: JSONSchema, url: string): string {
 
   return baseName;
 }
+
+export function isPrimitiveType(schema: JSONSchema): boolean {
```
gorkem, line 59: JSON Schema does not really define primitive types. It has a concept of basic types but it is not exactly what is meant in here. Can you add some more documentation to define what this function returns since JSON schema lacks the concept. Alternately we may try to figure out a better name.

gorkem (changes requested):
This looks good. I think it is good to go after documenting the new util function.

src/languageservice/utils/schemaUtils.ts
p-spacek: Do you think that this implementation is better?
```ts
export function isNotReferenceType(schema: JSONSchema): boolean {
  return schema.type !== 'object' && schema.type !== 'array' && !isAnyOfAllOfOneOfType(schema);
}
```

The purpose of this change is to exclude some of the previous fix (see screenshot) and not to apply this fix when the 'array complation' is handled in different part of code
image

or if this shared functionality doesn't make much sense (I began to think so) I can just remove exported function and only modify the previous implementation:
```ts
if (nodeParent && isSeq(nodeParent) && schema.schema.type !== 'object' && !isAnyOfAllOfOneOfType(schema.schema)) {
```

p-spacek: @gorkem , any updates here, please?

gorkem: Sorry, I was on leave and missed the update. the new function and name look better.  Shared functionality makes sense, I just wanted it to be clear for everybody to use.

gorkem (approved):

Issue #684, opened 2022-03-22
autocompletion problem when value is null inside anyOf object


p-spacek:
## Describe the bug
when schema contains anyOf objects with null type, other types are ignored on value completion.

```json
{
  "anyOf": [
    {
      "properties": {
        "prop": {
          "const": "const value"
        }
      }
    },
    {
      "properties": {
        "prop": {
          "type": "null"
        }
      }
    }
  ]
}

```

yaml
```yaml
prop: #cursor here
```

## Expected Behavior
autocompletion should suggest `null` and `const value`

## Current Behavior
autocompletion suggest only `null`

<img width="119" alt="image" src="https://user-images.githubusercontent.com/38421337/159473407-e60faf46-1783-4ea7-8899-c778f1910970.png">

note that problem is probably in `doc.getMatchingSchemas(schema.schema);` where `prop: ` is equivalent to `prop: null` so the only correct schema is the one with null type.

PR #759, opened 2022-08-02
Fix auto complete when null value inside any of

msivasubramaniaan:
### What does this PR do?
This PR should suggest all the matched values from anyOf

https://github.com/redhat-developer/yaml-language-server/issues/684

### Is it tested? How?
Yes. Added UT for current scenario.


gorkem (approved):
LGTM

PR #763, opened 2022-08-06
Add deprecated hint

yassun4dev:
### What does this PR do?
Add [derecated key](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.3) hint.

Nothing.

### Is it tested? How?
Add deprecated test case.


src/languageservice/parser/jsonParser07.ts
```diff
@@ -878,6 +882,30 @@ function validate(
...
+        location = {
+          offset: node.parent.offset,
+          length: node.parent.length,
+        };
+      } else {
+        location = { offset: node.offset, length: node.length };
+      }
+      validationResult.problems.push({
+        location,
+        severity: DiagnosticSeverity.Hint,
```
gorkem, line 898: Should this be a warning?

gorkem (changes requested):

gorkem:
@yassun4dev do you plan to work on updating this ?

src/languageservice/parser/jsonParser07.ts
ssbarnea: Usually I would say the same deprecation as warning, but if they are doing hint for json language server, i would try to keep it the same, to avoid diverging.

ReubenFrankel:
Also interested in this.

ReubenFrankel:
I am also interested in this. :+1:

yassun4dev:
@gorkem @ReubenFrankel

Sorry, I have recently realized that I have stopped this PR.

Change Severity from Hint to Warning.

The question with changing Severity is that the internal process forces Warning to be converted to Error (See the added test case).

I could not figure out why it was treated as an Error.

PR #775, opened 2022-09-02
feat: add diagnostic data with missing properties

p-spacek:
### What does this PR do?
When problem `'Property {0} is not allowed.'` is found during diagnostic it will extend `diagnostic.data` by possible props that could be inside the object.
 - `diagnostic.data = Object.keys(schema.properties) subtract seenProperties`

then it can be used inside `CodeAction` to suggest proper properties instead of the wrong one.

no ref

### Is it tested? How?
added tests

### DIFF
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -80,6 +80,7 @@ export interface IProblem {
   problemType?: ProblemType;
   problemArgs?: string[];
   schemaUri?: string[];
+  data?: Record<string, unknown>;
 }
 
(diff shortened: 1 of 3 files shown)


gorkem:
Are you also planning to contribute `CodeAction` changes?

p-spacek:
> Are you also planning to contribute `CodeAction` changes?

I didn't plan it (I needed some customized functionality in client, so I didn't need it in YLS), but I can add it (but not sure when).

gorkem (approved):

msivasubramaniaan (approved):

PR #782, opened 2022-09-20
Update TypeScript target from es5 to es2020

remcohaszing:
### What does this PR do?

es2020 is compatible with Node.js 14 and all modern browsers. Notably this no longer transpiles async/await syntax.

N/A

### Is it tested? How?

```sh
npm test
```


gorkem:
I do not see any issue with moving to a newer nodejs version. 12 is EOL anyway.

gorkem:
@remcohaszing I think this will pass now if you rebase

remcohaszing:
Thanks! Rebased

ssbarnea (approved):

Issue #769, opened 2022-08-23
Can't suggest object correctly with certain schema definition.


chuyangguangshin:
## Describe the bug
Can't suggest object correctly with certain schema definition.

Schema used in the demo:
```
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "type":"object",
  "oneOf": [
    {
      "additionalProperties": false,
      "properties": {
        "A" : {
          "type":"string"
        }
      },
      "required": [
        "A"
      ]
    },
    {
      "additionalProperties": false,
      "properties": {
        "B" : {
          "type":"string"
        }
      },
      "required": [
        "B"
      ]
    }
  ]
}
```

## Expected Behavior
It should suggest A and B while the yaml is empty as version yaml-language-server 0.23.0 does.
![image](https://user-images.githubusercontent.com/6897710/186119066-5dea959a-8df2-4273-bfa5-7fd98119c675.png)

## Current Behavior
It only suggest A when  yaml-language-server 1.9.1 installed
![image](https://user-images.githubusercontent.com/6897710/186119466-370041d8-3b94-45d6-bb3c-f2fb633e3151.png)

Not only in empty file. It will not suggest all possible objects at some child node position with the schema provided .

## Steps to Reproduce
1. Config yaml to verify with the schema provided.
2. Trig suggestion on a empty yaml.

## Environment
- [ ] Windows

PR #785, opened 2022-09-26
added child label value to parent label if parent label undefined

msivasubramaniaan:

### What does this PR do?
This PR should suggest all matching combination values when auto completion

https://github.com/redhat-developer/yaml-language-server/issues/769

### Is it tested? How?
Yes added UT


gorkem (approved):
LGTM

PR #786, opened 2022-09-26
suggest valid options from oneOf on code completion

msivasubramaniaan:

### What does this PR do?
This PR suggest the valid option on code completion from oneOf

https://github.com/redhat-developer/vscode-yaml/issues/829

### Is it tested? How?
Yes added UT


gorkem (changes requested):
There is one failing test for me locally. I have kicked the CI and I see the same test failure.

msivasubramaniaan:
> There is one failing test for me locally. I have kicked the CI and I see the same test failure.

Done

gorkem (approved):

src/languageservice/services/yamlCompletion.ts
```diff
@@ -524,6 +524,24 @@ export class YamlCompletion {
 
     this.finalizeParentCompletion(result);
 
+    const uniqueItems = result.items.filter(
```
p-spacek, line 527: Hello, I have a small note here, not sure if I should create an issue for that...

but this piece of code (and test modifications below) is a little bit hiding another problem that some internal methods are called multiple times from different places with the same result.

for example, some schema.snippets are calculated multiple times

A user can't see any problems (results are merged), but it's not correct.

If you decide to create a task for that, I can try to solve it.

Thanks for your great open-source project!!

PR #801, opened 2022-11-10
added quotes for null key during completion

msivasubramaniaan:

### What does this PR do?
This PR added quotes if the key is 'null'

https://github.com/redhat-developer/vscode-yaml/issues/842

### Is it tested? How?
Yes added UT.


gorkem (approved):
/lgtm

PR #802, opened 2022-11-11
added ipv4 and ipv6 format validation

msivasubramaniaan:

### What does this PR do?
Added IPv4 and IPv6 format validation on parser

https://github.com/redhat-developer/vscode-yaml/issues/832

### Is it tested? How?
Yes added UT

### DIFF
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -29,6 +29,7 @@ import { safeCreateUnicodeRegExp } from '../utils/strings';
 import { FilePatternAssociation } from '../services/yamlSchemaService';
 
 const localize = nls.loadMessageBundle();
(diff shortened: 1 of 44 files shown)


src/languageservice/parser/jsonParser07.ts
```diff
@@ -1066,6 +1066,21 @@ function validate(
             }
           }
           break;
+        case 'ipv4':
```
gorkem, line 1069: What is the reason why `ipv4` and `ipv6` are not added to the `formats` and handled the same way above? I think it is better to unify the implementation path which also fixes how the message should be handled.

gorkem (changes requested):
Commented also on `jsonParser07.ts`. The implemenation should follow the same path as other patterns. The current implemenation does not handle schema provided error messages which would be fixed by using the same implementation path. 

We also need issues filed for fixing other known formats.

msivasubramaniaan:
I will raise new PR

PR #828, opened 2023-01-09
feat: add possibility to define multiple validation patterns

p-spacek:
### What does this PR do?
add the possibility to define multiple validation patterns
The previous implementation supported only single pattern and single patternErrorMessage. It was too hard to define various validations.
With this implementation, it's possible to define multiple patterns (conditions) with more specific messages

no ref

### Is it tested? How?
unit test

### DIFF
--- a/src/languageservice/jsonSchema.ts
+++ b/src/languageservice/jsonSchema.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { CompletionItemKind } from 'vscode-json-languageservice';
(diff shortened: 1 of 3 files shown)


gorkem:
Does the jsonSchema specification have patterns as arrays with the additional fields that you have added here? I can not seem to find a reference to it.

p-spacek:
> Does the jsonSchema specification have patterns as arrays with the additional fields that you have added here? I can not seem to find a reference to it.

Unfortunately not, it's a custom implementation, it's not based on jsonSchema standard.
Is it a problem?
It can be quite useful to be able to define 'validation conditions'.

msivasubramaniaan:
@p-spacek please update the branch

PR #831, opened 2023-01-17
refactor(language service): Clean up the language service interface

jpinkney-aws:
### What does this PR do?
Some of the functionality currently exposed by the language service doesn't make sense when running the language service as a library rather than through the language server. This PR seeks to clean up some of those cases by marking connection, telemetry, and yamlSettings as optional

None

### Is it tested? How?

Tested by running `npm run test` and then ensuring that VSCode-YAML works with these changes

### DIFF
--- a/src/languageserver/handlers/schemaSelectionHandlers.ts
+++ b/src/languageserver/handlers/schemaSelectionHandlers.ts
@@ -14,13 +14,13 @@ import { JSONSchemaDescription, JSONSchemaDescriptionExt, SchemaSelectionRequest
 export class JSONSchemaSelection {
   constructor(
     private readonly schemaService: YAMLSchemaService,
-    private readonly yamlSettings: SettingsState,
-    private readonly connection: Connection
+    private readonly yamlSettings?: SettingsState,
+    private readonly connection?: Connection
   ) {
-    this.connection.onRequest(SchemaSelectionRequests.getSchema, (fileUri) => {
+    this.connection?.onRequest(SchemaSelectionRequests.getSchema, (fileUri) => {
       return this.getSchemas(fileUri);
     });
-    this.connection.onRequest(SchemaSelectionRequests.getAllSchemas, (fileUri) => {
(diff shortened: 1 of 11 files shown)


gorkem:
We have really not put an effort to support language service as a library, it is not something we regularly check for. Are there any tests that can be added so that this does not get broken in the future.

nkomonen-amazon:
Hey @gorkem, I'm going to be more involved with API changes to the YAML service.

> Are there any tests that can be added so that this does not get broken in the future.

IMO this isn't really related to anything we can test against. It is more related to the overall design which would have needed to been addressed in the initial design of the function.

The root change of this PR is [here](https://github.com/redhat-developer/yaml-language-server/pull/831/files#diff-c359e47ac27381149a26a7112d00be0af15543e40c5610ee9d1d7c2126a4cf2fR180-R183), which is making the non-essential arguments of the core YAML Language Service optional. They are not essential to core language service functionality and most users of the API will not need the functionality they provide.

Without this PR it forces complex creation of unnecessary objects to get access to the language service.

---

I think there is some work we can do to improve this API and would be interested in contributing to it. I've seen lots of use of the language service api, so I think this is something that would be beneficial to the community.

gorkem:
This is awesome. It is a reunion with my oldterns :)
 
I think I understand what the PR is trying to achieve. I think there should be a test that tries to initialize the language service as you described and check if it works as expected. Otherwise, we do not regularly test the project in such usage.

nkomonen-amazon:
Its a small world in the IDE space :) I never thought I'd have a reunion on a Github issue lol.

> I think there should be a test that tries to initialize the language service

I see now, that makes sense. I'll add that in to this PR soon. Thanks :)

gorkem:
@nkomonen-amazon the new test is failing for me locally and at CI too.

nkomonen-amazon:
@gorkem I've updated with a new commit that fixes the issue

gorkem:
@msivasubramaniaan Can you check this one?

PR #832, opened 2023-01-17
fix: completion value with dash

p-spacek:
### What does this PR do?
fix the completion item range when yaml contains a word ending with a dash symbol.

https://user-images.githubusercontent.com/38421337/212872479-136c433b-9588-4c7c-a3d9-747a64af719d.mov

I removed the piece of code that caused this problem (all existing tests passed after removing it).
To be honest I am not sure why this code was there. I tried to think about but I wasn't able to find a scenario when is needed to subtract one character if the current position contains `-`

note that this example didn't use removed piece of code
```yaml
array:
 -
```

no ref

### Is it tested? How?
added test


msivasubramaniaan:
@p-spacek Can you please update the base branch?

msivasubramaniaan (approved):
LGTM

PR #833, opened 2023-01-17
feat: exclude deprecated schema from autocompletion

p-spacek:
### What does this PR do?
exclude deprecated schema from autocompletion
useful when you want to exclude some subschemas from autocompletion, but you need it to be there for validation purposes. So you can't just remove it.

```json
"anyOf":[
  {
     "$ref": "schema1",
     "deprecatedMessage": "Deprecated: use schema2 instead"
   },
  {
     "$ref": "schema2",
   }
]
```

no ref

### Is it tested? How?
added tests

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -697,6 +697,10 @@ export class YamlCompletion {
       });
     }
     for (const schema of matchingSchemas) {
+      if (schema.schema.deprecationMessage) {
+        continue;
+      }
+
       if (
         ((schema.node.internalNode === node && !matchOriginal) ||
           (schema.node.internalNode === originalNode && !hasColon) ||
--- a/test/autoCompletionFix.test.ts
+++ b/test/autoCompletionFix.test.ts
@@ -1206,4 +1206,38 @@ test1:
     expect(completion.items[0].insertText).to.be.equal('${1:property}: ');
     expect(completion.items[0].documentation).to.be.equal('Property Description');
   });
+
+  describe('Deprecated schema', () => {
+    it('should not autocomplete deprecated schema', async () => {
+      const schema: JSONSchema = {
+        properties: {
+          prop1: { type: 'string' },
+        },
+        deprecationMessage: 'Deprecated',
+      };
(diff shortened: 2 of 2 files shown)


gorkem:
This should be setting to exclude them. Otherwise, LSPs do have a way to mark deprecated code completions as deprecated and we should actually use that mechanism instead of excluding them directly.

p-spacek:
> This should be setting to exclude them. Otherwise, LSPs do have a way to mark deprecated code completions as deprecated and we should actually use that mechanism instead of excluding them directly.

Hi @gorkem , Sorry, I am not sure if I understand correctly your thoughts. Can you explain them, please?
The previous implementation allows only to exclude properties, not the whole schema.

gorkem:
Deprecated APIs are common in programming languages. Typically language tooling deals with it, not by removing them without notice but by marking them. Deprecated properties can be marked as deprecated by the `yaml-language-server` using the deprecated hint. #763 is an incomplete implementation in that direction. We can also take it further and introduce a setting that will allow users to select that they do not want deprecated properties to be presented as code completion.

p-spacek:
Oh, I understand now. Sorry for the late reply.
So I think that I need `doNotSuggest` schema property to exclude something from code-completion

**Note1: current implementation excludes properties with `deprecationMessage`**
https://github.com/redhat-developer/yaml-language-server/blob/main/src/languageservice/services/yamlCompletion.ts#L719
`if (typeof propertySchema === 'object' && !propertySchema.deprecationMessage && !propertySchema['doNotSuggest']) {`

So I will probably create another fix PR that will finish the `doNotSuggest` implementation
 - `doNotSuggest` is implemented only for properties, but it should work also for schemas (value and property completion)

 do you agree, @gorkem ?
 
I won't modify the `deprecation` functionality at all for now...
 - maybe later as you suggested:
   - cross it out (typescript has the same)
   - some user configuration to define if a completion item should be excluded completely or crossed out

gorkem:
A complete solution would be to enable the deprecated hint (the cross out) and add a setting (not a schema property) to not suggest the deprecated properties. I do not think it is the schema author's decision whether to use a deprecated property or not. Schema author already has done its responsibility by marking it deprecated.

PR #834, opened 2023-01-19
fix: value completion with trailing spaces

p-spacek:
### What does this PR do?
fix range position with unfinished value completion with trailing spaces

#### before fix

https://user-images.githubusercontent.com/38421337/213417164-99b35c7f-0193-4ec8-8512-7bc5d90f0d2b.mov

#### after fix

https://user-images.githubusercontent.com/38421337/213417241-cdb0081e-764b-4236-9b95-254124dfb1ea.mov

no ref

### Is it tested? How?
added similar unit tests to 'partial key' tests


gorkem (approved):
LGTM

PR #839, opened 2023-01-24
Map key order

gorkem:
### What does this PR do?
Introduces a new setting `yaml.keyOrder` that enforces the alphabetical key order for mappings when set to true. 
In addition to validation, it also provides a code action to fix the order for the map

https://github.com/redhat-developer/vscode-yaml/issues/95

### Is it tested? How?
Adds tests for validation, new settings and code action

### DIFF
--- a/src/yamlSettings.ts
+++ b/src/yamlSettings.ts
@@ -29,6 +29,7 @@ export interface Settings {
       flowMapping: 'allow' | 'forbid';
       flowSequence: 'allow' | 'forbid';
     };
+    keyOrdering: boolean;
     maxItemsComputed: number;
     yamlVersion: YamlVersion;
   };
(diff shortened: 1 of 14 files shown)


msivasubramaniaan (approved):
LGTM

src/yamlSettings.ts
```diff
@@ -86,6 +87,7 @@ export class SettingsState {
     flowMapping: 'allow' | 'forbid';
     flowSequence: 'allow' | 'forbid';
   };
+  keyOrdering = true;
```
polyzen, line 90: Seems this should be `false`.

snaggen:
When I open a yaml file now in neovim (where I have installed yamlls via Mason and default config) I get a lot of "Wrong order of key" in my docker-compose.yaml and .gitlab-ci.yml files. I have tried to disable it by setting                         keyOrdering = false but with no improvement. 
Also, when looking at https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/yamlls.lua  it sems that they are not enabling this in their default config.

Issue #778, opened 2022-09-17
Newer/older Kubernetes schemas break autocompletion/diagnostics


someone-stole-my-name:
## Summary

While debugging https://github.com/someone-stole-my-name/yaml-companion.nvim/issues/12 we've seen some bizarre behavior. One would [expect](https://github.com/redhat-developer/yaml-language-server/pull/429/files) `yaml-language-server` to work with Kubernetes schemas higher or lower than the one [hardcoded](https://github.com/redhat-developer/yaml-language-server/blob/main/src/languageservice/utils/schemaUrls.ts#L8) if the user explicitly adds them, but that's not currently the case. Bumping even the patch version seems to break diagnostics and autocompletion behaves differently despite `1.22.4` and `1.22.5` being the same, e.g.

```json
{
    "yaml.schemas": {
        "https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/v1.22.5-standalone-strict/all.json": "/*.yaml"
    }
}
```

With `1.22.5` the following document doesn't display the expected `Property foo is not allowed`. 

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
foo: bar
```

And autocompletion is not working correctly either. Anything under `spec:` is not autocompleted as it does in `1.22.4`:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
spec:
```

### Related:
- https://github.com/redhat-developer/vscode-yaml/issues/253#issuecomment-564455764
- https://github.com/redhat-developer/vscode-yaml/issues/822
- https://github.com/someone-stole-my-name/yaml-companion.nvim/issues/12

devodev:
Bump ?

smthbrothrjdev:
I am having the same issue. Has anyone found a more permanent work-around besides the one listed [here](https://github.com/someone-stole-my-name/yaml-companion.nvim/issues/12) ?

mosheavni:
same issue here. can it be fixed?

Issue #307, opened 2020-08-20
Remove built in kubernetes support


JPinkney:
It's probably time that we start to deprecate the built-in Kubernetes support from the language server. Although it is useful, it really shouldn't be bundled with the extension and it creates issues that have to be specifically solved inside of the language server that don't really make sense in the context of a pure YAML language server.

E.g. https://github.com/redhat-developer/yaml-language-server/issues/220, https://github.com/redhat-developer/yaml-language-server/issues/211, https://github.com/redhat-developer/yaml-language-server/issues/132, https://github.com/redhat-developer/vscode-yaml/issues/153, https://github.com/redhat-developer/vscode-yaml/issues/246.

These are the things that should be done:
- [ ] Remove kubernetes references from README
- [ ] Remove kubernetes schema automapping (kubernetes -> https://raw.githubusercontent.com/instrumenta/kubernetes-json-schema/master/v1.17.0-standalone-strict/all.json)
- [ ] Remove kubernetes references in server.ts
- [ ] Rename all references of isKubernetes in https://github.com/redhat-developer/yaml-language-server/blob/master/src/languageservice/parser/jsonParser07.ts. _These need to be renamed and not removed because that part of the code is still useful. It allows you to use an alternative schema matching mechanism that may work better. This issue shows heavily in cases like the kubernetes schema where there are a lot of "oneOf" present in the json schema_
- [ ] Update Changelog with breaking change

ibotty:
It would be great to document how to migrate to a post-kubernetes yaml-language-server without losing kubernetes integration.

FranciscoKurpiel:
I hate to be negative, but Kubernetes integration is *the* reason why I'm using this language server.

trallnag:
@FranciscoKurpiel, can you share the kubernetes schema url? I cannot seem to find it

FranciscoKurpiel:
That's a snippet of my `coc-settings.json`. It is probably awful in a thousand ways, but it works. Kinda. I mean, who am I to know how to configure neovim?

Note that the URL is for k8s 1.18, but it should work with newer versions after making obvious changes to the url.

```json
{
  "languageserver": {
    "yaml": {
      "filetypes": ["yaml", "yml"],
      "module": "/home/xyko/.config/nvim/plugged/coc-yaml/node_modules/yaml-language-server/out/server/src/server.js"
    }
  },

  "yaml.schemas": {
[...]

Issue #211, opened 2020-01-03
Set specific Kubernetes Schema Version


joshrosso:
It appears the Kubernetes schema is hardcoded at https://github.com/redhat-developer/yaml-language-server/blob/master/src/server.ts#L32.

Is it possible to set a specific version? For example, if I wanted to use https://raw.githubusercontent.com/instrumenta/kubernetes-json-schema/master/v1.16.0-standalone-strict/all.json?

I tried

```json
yaml.schemas: {
        "https://raw.githubusercontent.com/instrumenta/kubernetes-json-schema/master/v1.16.0-standalone-strict/all.json": "/*.yaml"
}

```

but no luck.

apupier:
> It appears the Kubernetes schema is hardcoded at https://github.com/redhat-developer/yaml-language-server/blob/master/src/server.ts#L32.

it has been moved to https://github.com/redhat-developer/yaml-language-server/blob/7346049ac3c6dca6d3b27852c491f6851e3e41ab/src/languageservice/utils/schemaUrls.ts#L1

Using the yaml.schemas should be something working. I think it is a matter of priorities in schema resolution. hope it would be handled with https://github.com/redhat-developer/yaml-language-server/issues/250

Another solution would be to be able to specify the yaml schema in the yaml data file directly https://github.com/redhat-developer/yaml-language-server/issues/204

> I wonder if an [optional] environment variable could be helpful?

I'm not a big fan of environment variable as it can have collateral effect but it seems a quick win to add this possibility. (and given the number of +1 on this issue ti can allow to please a lot of people)

@JPinkney From technical point of view, is it a matter of modifying only https://github.com/redhat-developer/yaml-language-server/blob/7346049ac3c6dca6d3b27852c491f6851e3e41ab/src/server.ts#L115-L116 ?

joshrosso:
@JPinkney I think that's the right move.

It feels weird that Kubernetes is just kind of embedded and "works" (assuming the version is good :p) when you turn it on.

JPinkney:
@joshrosso Yeah, this language server originally started (3~ years ago) as a kubernetes language server and once we found out that the support for kubernetes worked well we expanded it to all of yaml but kept that schema embedded. That's why its been there for so long

joshrosso:
@JPinkney 100% makes sense!

PR #841, opened 2023-01-31
Add Custom Kubernetes Schema Support

tricktron:
### What does this PR do?

- It allows to set a custom Kubernetes schema in `yaml.schemas`. E.g:
```json
{
    "yaml.schemas": {
       "https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/v1.26.1-standalone-strict/all.json": "/**/*.yml",
    }
}
```
- Would allow to remove the hardcoded Kubernetes schema completely from `src`. I would still keep it for the tests.

### How does it work?

If the schema uri contains the substring `kubernetes`, then the kubernetes validation path is taken.

- https://github.com/redhat-developer/yaml-language-server/issues/778
- https://github.com/redhat-developer/yaml-language-server/issues/211
- https://github.com/redhat-developer/yaml-language-server/issues/307

### Is it tested? How?

Yes I added a test.

### Next Steps If This Goes Into the Right Direction

- Remove the hardcoded Kubernetes schema
- Refactor Kubernetes related tests to its own block
- Document it in the readme and possibly changelog files.


fourpastmidnight:
Why not allow one to specify, within the settings JSON under the `kubernetes` property, either a simple string value (currently a glob) or a map specifying the version(s) of schemas to be used? 

The default configuration would be as it is today, and would simply use the latest released version of the schema. But, with the proposal above, the configuration could look like (note the syntax is using formal JSON syntax to be compatible with coc.nvim, but this could equally apply to VS Code, too, by not using quotes around the property names):

```json
{
    "yaml.settings": {
        "kubernetes": {
            "versions": [
                "1.22.3",
                "1.22.4",
                "1.22.5"
            ],
            "filetype": "*.ya?ml"
        }
    }
}
```

gorkem:
I am not thrilled with the idea that a segment in the URL changes the meaning of the setting. It is prone to users unintentionally defining Kubernetes schema.  I would rather have an intentional setting to manage Kubernetes schemas. Perhaps a combination of what is proposed on #824 and what @fourpastmidnight is proposing. 

OTOH, I am a bit torn about Kubernetes support. Kubernetes is the only built-in schema that we have on yaml-language-server and this is due to the reasons how I had bootstrapped the project initially. Once we implemented a generic YAML LSP, the built-in Kubernetes support became an outlier.

fourpastmidnight:
I thought it was strange that this lsp server explicitly supported Kubernetes while providing a generic mechanism for other YAML schemas. I wasn't sure if there was any appetite for removing this support, or plans for adding explicit support for other "popular"types.

I also think it would be better to remove explicit Kubernetes support as well in light of no plans for explicitly supporting other schemas in lieu of the generic mechanism. I only made the suggestion above based on the current implementation.

tricktron:
@gorkem
> I am not thrilled with the idea that a segment in the URL changes the meaning of the setting. It is prone to users unintentionally defining Kubernetes schema. I would rather have an intentional setting to manage Kubernetes schemas.

I thought about this as well. I went with this approach because it is a non-breaking API change, e.g. the clients do not have to change anything. I was unable to come up with an intentional non-breaking change.

Maybe we can open a new pull request for the intentional kubernetes change in a second step?

If we document the magic `kubernetes` regex in the readme for now, e.g. in a separate Kubernetes support section, then I think this is good enough because allowing to specify the exact Kubernetes version is a great enhancement.

tricktron:
@gorkem @fourpastmidnight Any updates?

tricktron:
@gorkem @fourpastmidnight @msivasubramaniaan I completely reworked this PR on the basis of https://github.com/redhat-developer/yaml-language-server/pull/824. What do you think?

JannoTjarks:
Hi, what is the status of this pr?
I would like to have this feature. We are running Kubernetes in Version 1.27. There happened some changes since the hard coded version 👀

tricktron:
> Hi, what is the status of this pr? I would like to have this feature. We are running Kubernetes in Version 1.27. There happened some changes since the hard coded version 👀

I'll rebase and fix the code scanning results so that it is ready for merge again.

tricktron:
Ok. I rebased it. @gorkem @fourpastmidnight @msivasubramaniaan Could you run the build again and maybe give some feedback😃.

tricktron:
Any update @gorkem @fourpastmidnight @msivasubramaniaan?

mosheavni:
any updates?

tricktron:
@msivasubramaniaan I have seen that you recently merged https://github.com/redhat-developer/vscode-yaml/pull/912 which depends on this pull request. Could you  merge this as well?

tricktron:
@msivasubramaniaan I rebased again because I suspect that the merge of main through the GUI somehow triggered the `duplicate items issue` in the upload sarif stage.

If the same `duplicate items issue` happens again, then it might be worth to look at this similar issue: https://github.com/microsoft/sarif-js-sdk/issues/86.

PhiliNBG:
Any updates?

mvillafuertem:
any updates?

eternalphane:
kindly ping

PR #843, opened 2023-02-13
added ipv4 and ipv6 validation

msivasubramaniaan:

The existing [PR](https://github.com/redhat-developer/yaml-language-server/pull/802) has some issues due to my local reabase, Hence created a new one

### What does this PR do?
Added IPv4 and IPv6 format validation on parser

https://github.com/redhat-developer/vscode-yaml/issues/832

### Is it tested? How?
Yes added UT

### DIFF
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -62,7 +62,7 @@ export const formats = {
     pattern: /^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}$/,
   },
   ipv6: {
-    errorMessage: localize('ipv4FormatWarning', 'String does not match IPv6 format.'),
+    errorMessage: localize('ipv6FormatWarning', 'String does not match IPv6 format.'),
     pattern: /^([0-9a-f]|:){1,4}(:([0-9a-f]{0,4})*){1,7}$/i,
   },
 };
@@ -1066,7 +1066,7 @@ function validate(
         case 'ipv6':
           {
             const format = formats[schema.format];
-            if (!node.value || !format.pattern.exec(node.value)) {
+            if (!node.value || !format.pattern.test(node.value)) {
               validationResult.problems.push({
                 location: { offset: node.offset, length: node.length },
                 severity: DiagnosticSeverity.Warning,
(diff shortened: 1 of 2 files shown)


rgrunber:
When I have something like :

```json
{
    "type": "array",
    "items": {
      "type": "string",
      "format": "ipv6"
    }
}
```

```yaml
[
"2001:0db8:85a3:0000:0000:8a2e:0370:7334",
"2001:0db8:85a3:0000:0000:8a2e:0370:7334"
]
```

and associate the schema with the file, one of the entires doesn't validate.

msivasubramaniaan:
> ```yaml
> 2001:0db8:85a3:0000:0000:8a2e:0370:7334
> ```

@rgrunber 

YAML don't show any error on the mentioned string and it validates as expected. Please refer the screenshot
image

msivasubramaniaan:
> When I have something like :
> 
> ```json
> {
>     "type": "array",
>     "items": {
>       "type": "string",
>       "format": "ipv6"
>     }
> }
> ```
> 
> ```yaml
> [
> "2001:0db8:85a3:0000:0000:8a2e:0370:7334",
> "2001:0db8:85a3:0000:0000:8a2e:0370:7334"
> ]
> ```
> 
> and associate the schema with the file, one of the entires doesn't validate.
> 
> **Update:** This is definitely a bug. In fact the root cause is the regular expression used for the ipv4/ipv6 patterns. You've included a flag for your particular pattern that will results in some unpredictable behaviour.

@rgrunber 
Agree. That happened due to multi-line flag. Updated the regex pattern and added all the scenarios on the test case for better coverage.

src/languageservice/parser/jsonParser07.ts
```diff
@@ -1054,6 +1062,8 @@ function validate(
         case 'date':
         case 'time':
         case 'email':
+        case 'ipv4':
+        case 'ipv6':
```
rgrunber, line 1066: A few lines below here where the `exec` call is made, would it be better to change to `test` ? We don't really use any of the objects from `exec`.

src/languageservice/parser/jsonParser07.ts
```diff
@@ -57,6 +57,14 @@ export const formats = {
     errorMessage: localize('emailFormatWarning', 'String is not an e-mail address.'),
     pattern: /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/,
   },
+  ipv4: {
+    errorMessage: localize('ipv4FormatWarning', 'String does not match IPv4 format.'),
+    pattern: /^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}$/,
+  },
+  ipv6: {
+    errorMessage: localize('ipv4FormatWarning', 'String does not match IPv6 format.'),
```
rgrunber, line 65: Should this be "ipv6FormatWarning" ?

rgrunber (approved):
Overall looks good, so once the last comments are addressed, I thin this is good to squash + merge.

Issue #837, opened 2023-01-24
Hover is always converting the contents to markdown


jpinkney-aws:
## Describe the bug

When you hover over a element that has a schema applied the hover results get converted directly into markdown regardless of whether or not you are using the `description` json schema field or the `markdownDescription` json schema field. This actually breaks non-VSCode editors like Visual Studio, which can't/don't handle the markdown from hover gracefully.

It's also worth noting that we don't convert our completion description results to markdown. They are shown as plaintext.

## Expected Behavior

Hovers coming from `description` json schema fields should not re-render everything as markdown. We should only convert the incoming description as markdown when the user uses the `markdownDescription` field in the schema.

## Current Behavior

Every hover request gets rendered as markdown.

Related code: https://github.com/redhat-developer/yaml-language-server/blob/main/src/languageservice/services/yamlHover.ts#L89 and https://github.com/redhat-developer/yaml-language-server/blob/main/src/languageservice/services/yamlHover.ts#L117

## Steps to Reproduce

1. Create file with schema attached that has a description that should be rendered as plaintext
2. Hover over an element
3. Observe that the hover element gets converted to markdown 

## Environment
- [ ] Windows
- [X] Mac
- [ ] Linux
- [ ] other (please specify)

gorkem:
Are you planning to send a PR for this ?

rgrunber:
Looks like at https://github.com/redhat-developer/yaml-language-server/blob/f8885262ad219e28121d4571bd6fc2a07fec984c/src/languageservice/services/yamlHover.ts#L120 , https://github.com/redhat-developer/yaml-language-server/blob/f8885262ad219e28121d4571bd6fc2a07fec984c/src/languageservice/services/yamlHover.ts#L141

which just treat the regular description as markdown. Seems like it shouldn't be that tricky to handle these cases. We already uses [`MarkupContent`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContent) in the hover, which supports both.

PR #844, opened 2023-02-13
replace multiple space into single space on markdown text in hover

msivasubramaniaan:

### What does this PR do?
This PR shows the description without escape sequence char even description has indentation on hover

https://github.com/redhat-developer/vscode-yaml/issues/886

### Is it tested? How?
Yes. Added UT

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,7 +10,8 @@
       "sourceMaps": true,
       "outFiles": ["${workspaceRoot}/out/server/**/*.ts"],
       "protocol": "inspector",
-      "trace": true
+      "trace": true,
+      "preLaunchTask": "watch typescript",
     },
     // A launch configuration that compiles the server and runs mocha unit tests
     {
--- a/src/languageservice/services/yamlHover.ts
+++ b/src/languageservice/services/yamlHover.ts
@@ -204,8 +204,9 @@ function toMarkdown(plain: string): string;
 function toMarkdown(plain: string | undefined): string | undefined;
 function toMarkdown(plain: string | undefined): string | undefined {
   if (plain) {
-    const res = plain.replace(/([^\n\r])(\r?\n)([^\n\r])/gm, '$1\n\n$3'); // single new lines to \n\n (Markdown paragraph)
-    return res.replace(/[\\`*_{}[\]()#+\-.!]/g, '\\$&'); // escape markdown syntax tokens: http://daringfireball.net/projects/markdown/syntax#backslash
+    let res = plain.replace(/([^\n\r])(\r?\n)([^\n\r])/gm, '$1\n\n$3'); // single new lines to \n\n (Markdown paragraph)
+    res = res.replace(/[\\`*_{}[\]()#+\-.!]/g, '\\$&'); // escape markdown syntax tokens: http://daringfireball.net/projects/markdown/syntax#backslash
+    return res.replace(/([^\S\n\r]{4,})/g, ' '); //replace multiple space to single space
   }
   return undefined;
 }
--- a/src/languageservice/services/yamlHover.ts
+++ b/src/languageservice/services/yamlHover.ts
@@ -206,17 +206,20 @@ function toMarkdown(plain: string | undefined): string | undefined {
   if (plain) {
     let res = plain.replace(/([^\n\r])(\r?\n)([^\n\r])/gm, '$1\n\n$3'); // single new lines to \n\n (Markdown paragraph)
     res = res.replace(/[\\`*_{}[\]()#+\-.!]/g, '\\$&'); // escape markdown syntax tokens: http://daringfireball.net/projects/markdown/syntax#backslash
-    return res.split('\n\n').length > 0 ? removeMarkdownEscapes(res) : res;
+    const splitedStrings = res.split('\n\n');
+    return splitedStrings.length > 0 ? removeMarkdownEscapes(splitedStrings) : res;
   }
   return undefined;
 }
 
 //logic used from https://github.com/microsoft/vscode/issues/16936
-function removeMarkdownEscapes(value: string): string {
+function removeMarkdownEscapes(values: string[]): string {
   let markdownString = '';
-  value.split('\n\n').forEach((splitVal) => {
+  values.forEach((splitVal, index) => {
     markdownString += splitVal.match(/([^\S\n\r]{2,})/g) ? splitVal.replace(/\\([\\`*_{}[\]()#+\-.!])/g, '$1') : splitVal;
-    markdownString += '\n\n';
+    if (index < values.length - 1) {
+      markdownString += '\n\n';
+    }
   });
   return markdownString;
 }
--- a/src/languageservice/services/yamlHover.ts
+++ b/src/languageservice/services/yamlHover.ts
@@ -5,7 +5,7 @@
  *--------------------------------------------------------------------------------------------*/
 'use strict';
 
-import { Hover, MarkupContent, Position, Range } from 'vscode-languageserver-types';
+import { Hover, MarkupContent, MarkupKind, Position, Range } from 'vscode-languageserver-types';
 import { matchOffsetToDocument } from '../utils/arrUtils';
 import { LanguageSettings } from '../yamlLanguageService';
 import { YAMLSchemaService } from './yamlSchemaService';
@@ -23,6 +23,7 @@ import { ASTNode } from 'vscode-json-languageservice';
 
 export class YAMLHover {
   private shouldHover: boolean;
+  private indentation: string;
   private schemaService: YAMLSchemaService;
 
   constructor(schemaService: YAMLSchemaService, private readonly telemetry: Telemetry) {
@@ -33,6 +34,7 @@ export class YAMLHover {
   configure(languageSettings: LanguageSettings): void {
     if (languageSettings) {
       this.shouldHover = languageSettings.hover;
+      this.indentation = languageSettings.indentation;
     }
   }
 
@@ -86,9 +88,10 @@ export class YAMLHover {
     );
 
     const createHover = (contents: string): Hover => {
+      const regex = new RegExp(this.indentation, 'g');
       const markupContent: MarkupContent = {
-        kind: 'markdown',
-        value: contents,
+        kind: MarkupKind.Markdown,
+        value: contents.replace(regex, '&emsp;'),
       };
       const result: Hover = {
         contents: markupContent,
@@ -204,24 +207,10 @@ function toMarkdown(plain: string): string;
 function toMarkdown(plain: string | undefined): string | undefined;
 function toMarkdown(plain: string | undefined): string | undefined {
   if (plain) {
-    let res = plain.replace(/([^\n\r])(\r?\n)([^\n\r])/gm, '$1\n\n$3'); // single new lines to \n\n (Markdown paragraph)
-    res = res.replace(/[\\`*_{}[\]()#+\-.!]/g, '\\$&'); // escape markdown syntax tokens: http://daringfireball.net/projects/markdown/syntax#backslash
-    const splitedStrings = res.split('\n\n');
-    return splitedStrings.length > 0 ? removeMarkdownEscapes(splitedStrings) : res;
+    const res = plain.replace(/([^\n\r])(\r?\n)([^\n\r])/gm, '$1\n\n$3'); // single new lines to \n\n (Markdown paragraph)
+    return res.replace(/[\\`*_{}[\]()#+\-.!]/g, '\\$&'); // escape markdown syntax tokens: http://daringfireball.net/projects/markdown/syntax#backslash
   }
   return undefined;
-}
-
-//logic used from https://github.com/microsoft/vscode/issues/16936
-function removeMarkdownEscapes(values: string[]): string {
-  let markdownString = '';
-  values.forEach((splitVal, index) => {
-    markdownString += splitVal.match(/([^\S\n\r]{2,})/g) ? splitVal.replace(/\\([\\`*_{}[\]()#+\-.!])/g, '$1') : splitVal;
-    if (index < values.length - 1) {
-      markdownString += '\n\n';
-    }
-  });
-  return markdownString;
 }
 
 // copied from https://github.com/microsoft/vscode-json-languageservice/blob/2ea5ad3d2ffbbe40dea11cfe764a502becf113ce/src/services/jsonHover.ts#L122
--- a/src/languageservice/services/yamlHover.ts
+++ b/src/languageservice/services/yamlHover.ts
@@ -206,9 +206,19 @@ function toMarkdown(plain: string | undefined): string | undefined {
   if (plain) {
     let res = plain.replace(/([^\n\r])(\r?\n)([^\n\r])/gm, '$1\n\n$3'); // single new lines to \n\n (Markdown paragraph)
     res = res.replace(/[\\`*_{}[\]()#+\-.!]/g, '\\$&'); // escape markdown syntax tokens: http://daringfireball.net/projects/markdown/syntax#backslash
-    return res.replace(/([^\S\n\r]{4,})/g, ' '); //replace multiple space to single space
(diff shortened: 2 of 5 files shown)


fabiodouek:
Hi, thanks a lot for looking into this issue.
I had a quick look into this PR. It seems that the adopted solution is to replace multiple spaces by a single space, which in fact is a workaround to the issue. 
However, in my view spaces should be preserved. In my specific case, the json schema has python code snippets and json snippets. If the spaces are not preserved, the code/json will be unreadable. 
Could we preserve the spaces and address the backslash issue?

Thanks, 
Fabio.

msivasubramaniaan:
> Hi, thanks a lot for looking into this issue. I had a quick look into this PR. It seems that the adopted solution is to replace multiple spaces by a single space, which in fact is a workaround to the issue. However, in my view, spaces should be preserved. In my specific case, in the json schema, the description attribute has python code snippets and json snippets. If the spaces are not preserved, the code/json will be unreadable. Could we preserve the spaces and address the backslash issue?
> 
> Thanks, Fabio.

Thanks for you response. I will try to preserve the space from string

fabiodouek:
> > Hi, thanks a lot for looking into this issue. I had a quick look into this PR. It seems that the adopted solution is to replace multiple spaces by a single space, which in fact is a workaround to the issue. However, in my view, spaces should be preserved. In my specific case, in the json schema, the description attribute has python code snippets and json snippets. If the spaces are not preserved, the code/json will be unreadable. Could we preserve the spaces and address the backslash issue?
> > Thanks, Fabio.
> 
> Thanks for your response. I will try to preserve the space from string

Thanks a lot. Please let me know how it goes, Hopefully the spaces can be preserved, otherwise it would have a big impact for our users.

msivasubramaniaan:
> > > Hi, thanks a lot for looking into this issue. I had a quick look into this PR. It seems that the adopted solution is to replace multiple spaces by a single space, which in fact is a workaround to the issue. However, in my view, spaces should be preserved. In my specific case, in the json schema, the description attribute has python code snippets and json snippets. If the spaces are not preserved, the code/json will be unreadable. Could we preserve the spaces and address the backslash issue?
> > > Thanks, Fabio.
> > 
> > 
> > Thanks for your response. I will try to preserve the space from string
> 
> Thanks a lot. Please let me know how it goes, Hopefully the spaces can be preserved, otherwise it would have a big impact for our users.

Hello @fabiodouek,

As I am preserving the space from string. Please refer the attachment 
image

fabiodouek:
> msivasubramaniaan

Hi @msivasubramaniaan , that's great, thanks a lot for that.
No pressure, but do you know roughly when this fix will be released? 

Thanks,
Fabio.

msivasubramaniaan:
Hello @fabiodouek 

You can expect the release 1st week of March. The pre-release by end of next week.

rgrunber:
I don't think this is the right approach. The language server is behaving exactly per the spec. The problem is with the client not being able to handle the response.

The language server returns a `{ contents: MarkupContent, range: Range }` which is fine. The VS Code client expects a `contents: Array<MarkdownString | MarkedString>, range: Range` . 

https://github.com/microsoft/vscode-languageserver-node/blob/release/client/7.0.0/client/src/common/protocolConverter.ts#L361 is probably what gets called to convert our hover LS response into a VS Code hover so I think we just need to figure out what's going wrong there.

FWIW, I see that vscode-java properly deals with the hover and doesn't unescape the underscores.

.vscode/launch.json
```diff
@@ -10,7 +10,8 @@
       "sourceMaps": true,
       "outFiles": ["${workspaceRoot}/out/server/**/*.ts"],
       "protocol": "inspector",
-      "trace": true
+      "trace": true,
+      "preLaunchTask": "watch typescript",
```
msivasubramaniaan, line 14: Added pre launch check

src/languageservice/services/yamlHover.ts
```diff
@@ -86,9 +88,10 @@ export class YAMLHover {
     );
 
     const createHover = (contents: string): Hover => {
+      const regex = new RegExp(this.indentation, 'g');
```
msivasubramaniaan, line 91: @rgrunber 
Tabs and whitespace have a special meaning in [Markdown.](https://www.markdownguide.org/hacks/#indent-tab) So converted spaces into `non braking space` as a workaround here.

rgrunber (approved):
I think this is pretty reasonable, so feel free to merge when ready. If in the future we end up having to do more replacement (beyond just the indentation string), It might be good to create a separate helper method.

polyzen:
Seems this broke presentation in Neovim:
```
&emsp;#&emsp;#&emsp;#&emsp;#&emsp; &emsp;g&emsp;i&emsp;t&emsp;l&emsp;a&emsp;b&emsp;\&emsp;-&emsp;c&emsp;i&emsp; &emsp;|&emsp;|&emsp; &emsp;g&emsp;i&emsp;t&emsp;l&emsp;a&emsp;b&emsp;\&emsp;-&emsp;c&emsp;i&emsp;
&emsp;
&emsp; &emsp;
&emsp;
&emsp;S&emsp;o&emsp;u&emsp;r&emsp;c&emsp;e&emsp;:&emsp; &emsp;[&emsp;c&emsp;i&emsp;.&emsp;j&emsp;s&emsp;o&emsp;n&emsp;]&emsp;(&emsp;h&emsp;t&emsp;t&emsp;p&emsp;s&emsp;:&emsp;/&emsp;/&emsp;g&emsp;i&emsp;t&emsp;l&emsp;a&emsp;b&emsp;.&emsp;c&emsp;o&emsp;m&emsp;/&emsp;g&emsp;i&emsp;t&emsp;l&emsp;a&emsp;b&emsp;-&emsp;o&emsp;r&emsp;g&emsp;/&emsp;g&emsp;i&emsp;t&emsp;l&emsp;a&emsp;b&emsp;/&emsp;-&emsp;/&emsp;r&emsp;a&emsp;w&emsp;/&emsp;m&emsp;a&emsp;s&emsp;t&emsp;e&emsp;r&emsp;/&emsp;a&emsp;p&emsp;p&emsp;/&emsp;a&emsp;s&emsp;s&emsp;e&emsp;t&emsp;s&emsp;/&emsp;j&emsp;a&emsp;v&emsp;a&emsp;s&emsp;c&emsp;r&emsp;i&emsp;p&emsp;t&emsp;s&emsp;/&emsp;e&emsp;d&emsp;i&emsp;t&emsp;o&emsp;r&emsp;/&emsp;s&emsp;c&emsp;h&emsp;e&emsp;m&emsp;a&emsp;/&emsp;c&emsp;i&emsp;.&emsp;j&emsp;s&emsp;o&emsp;n&emsp;)&emsp;
```

rgrunber:
@polyzen , I'm guessing Neovim doesn't handle Markdown coming from the hovers ? @msivasubramaniaan , could https://github.com/redhat-developer/yaml-language-server/issues/837 solve this issue ?

polyzen:
https://neovim.io/doc/user/lsp.html#lsp-util
`convert_input_to_markdown_lines({input}, {contents})`

rgrunber:
It mentions : _... Converts any of MarkedString | MarkedString[] | MarkupContent into a list of lines containing valid markdown._

However, if you look at [`MarkupContent](`https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContentInnerDefinition), it's `kind` field determines whether the string value is markdown or plain text. In our case, markdown is always sent back.

https://github.com/redhat-developer/yaml-language-server/blob/c657e66ce36a1481b56b363bcc97e886161ab98a/src/languageservice/services/yamlHover.ts#L93 .

The other thing to check is if we respect https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#hoverClientCapabilities , which allows the clients to set their preferred format through `contentFormat`.

polyzen:
I figure it would accept `MarkupKing.Markdown`, but didn't get that far when looking into it.

polyzen:
If `value: contents,` is used as before, it renders again:
2023-03-08T23:07:30-05:00

msivasubramaniaan:
> @polyzen , I'm guessing Neovim doesn't handle Markdown coming from the hovers ? @msivasubramaniaan , could #837 solve this issue ?

Yes @rgrunber I hope #837 will resolve.
@polyzen Can you please share the init.vim for better understanding of plugin which you have used and configuration on the hover things.

Kasama:
I'm facing the same issue as polyzen

> I'm guessing Neovim doesn't handle Markdown coming from the hovers

Actually neovim handles [markdown from hovers](https://github.com/neovim/neovim/blob/f01f18cdf4ffc3ce035db5fde2f45493eebb7fd9/runtime/lua/vim/lsp/handlers.lua#L364) and [parses it a bit](https://github.com/neovim/neovim/blob/f01f18cdf4ffc3ce035db5fde2f45493eebb7fd9/runtime/lua/vim/lsp/util.lua#L1259) for better stylization. The thing is that it doesn't actually handle HTML escape sequences, so the `&emsp;` sequences are left as-is.

Consider the following content returned from a `textDocument/hover` request:
```json
{
  "contents": {
    "kind": "markdown",
    "value": "&emsp;#&emsp;#&emsp;#&emsp;#&emsp; &emsp;g&emsp;i&emsp;t&emsp;l&emsp;a&emsp;b&emsp;\\&emsp;-&emsp;c&emsp;i&emsp;\n&emsp;\n&emsp;T&emsp;h&emsp;e&emsp; &emsp;n&emsp;a&emsp;m&emsp;e&emsp; &emsp;o&emsp;f&emsp; &emsp;o&emsp;n&emsp;e&emsp; &emsp;o&emsp;r&emsp; &emsp;m&emsp;o&emsp;r&emsp;e&emsp; &emsp;j&emsp;o&emsp;b&emsp;s&emsp; &emsp;t&emsp;o&emsp; &emsp;i&emsp;n&emsp;h&emsp;e&emsp;r&emsp;i&emsp;t&emsp; &emsp;c&emsp;o&emsp;n&emsp;f&emsp;i&emsp;g&emsp;u&emsp;r&emsp;a&emsp;t&emsp;i&emsp;o&emsp;n&emsp; &emsp;f&emsp;r&emsp;o&emsp;m&emsp;\\&emsp;.&emsp;\n&emsp;\n&emsp;S&emsp;o&emsp;u&emsp;r&emsp;c&emsp;e&emsp;:&emsp; &emsp;[&emsp;c&emsp;i&emsp;.&emsp;j&emsp;s&emsp;o&emsp;n&emsp;]&emsp;(&emsp;h&emsp;t&emsp;t&emsp;p&emsp;s&emsp;:&emsp;/&emsp;/&emsp;g&emsp;i&emsp;t&emsp;l&emsp;a&emsp;b&emsp;.&emsp;c&emsp;o&emsp;m&emsp;/&emsp;g&emsp;i&emsp;t&emsp;l&emsp;a&emsp;b&emsp;-&emsp;o&emsp;r&emsp;g&emsp;/&emsp;g&emsp;i&emsp;t&emsp;l&emsp;a&emsp;b&emsp;/&emsp;-&emsp;/&emsp;r&emsp;a&emsp;w&emsp;/&emsp;m&emsp;a&emsp;s&emsp;t&emsp;e&emsp;r&emsp;/&emsp;a&emsp;p&emsp;p&emsp;/&emsp;a&emsp;s&emsp;s&emsp;e&emsp;t&emsp;s&emsp;/&emsp;j&emsp;a&emsp;v&emsp;a&emsp;s&emsp;c&emsp;r&emsp;i&emsp;p&emsp;t&emsp;s&emsp;/&emsp;e&emsp;d&emsp;i&emsp;t&emsp;o&emsp;r&emsp;/&emsp;s&emsp;c&emsp;h&emsp;e&emsp;m&emsp;a&emsp;/&emsp;c&emsp;i&emsp;.&emsp;j&emsp;s&emsp;o&emsp;n&emsp;)&emsp;"
  },
  "range": {
    "end": {
      "character": 9,
      "line": 2377
    },
    "start": {
      "character": 2,
      "line": 2377
    }
  }
}
```
The return itself is weird, containing `&emsp;` sequences between every character.

In neovim we can treat this by adding a wrapper function to handle the `textDocument/hover` responses

```lua
local function hover_wrapper(err, request, ctx, config)
  local bufnr, winnr = vim.lsp.handlers.hover(err, request, ctx, config)
  local contents = vim.api.nvim_buf_get_lines(bufnr, 0, -1, false)
  contents = vim.tbl_map(
    function(line)
      return string.gsub(line, "&emsp;", " ")
    end,
    contents
  )
  contents = vim.lsp.util.trim_empty_lines(contents)
  vim.api.nvim_buf_set_option(bufnr, 'modifiable', true)
  vim.api.nvim_buf_set_lines(bufnr, 0, -1, false, contents)
  vim.api.nvim_buf_set_option(bufnr, 'modifiable', false)
  vim.api.nvim_win_set_height(winnr, #contents)
end

vim.lsp.handlers["textDocument/hover"] = hover_wrapper
```

Even so, by replacing `"&emsp;"` with a space `" "`, the content is still pretty weird, with spaces around every character.

image

while replacing it with nothing `""` yields the expected result

image

---

I'm not sure I understand the reasoning behind [replacing indentations with `&emsp;`](https://github.com/redhat-developer/yaml-language-server/blame/42b7a8e3f465e72d55cb5ca95c7e1c8ffa8ca770/src/languageservice/services/yamlHover.ts#L94) or why it's causing it to be inserted between every character, but I believe this is a bug with this implementation.

Aside from that, I also think that neovim can benefit from properly handling HTML escape sequences, I'll create an issue there as well about this.

Kasama:
Oh yeah, I almost forgot. To reproduce this, here's a minimal `init.nvim`

```lua
local settings = {
  yaml = {
    schemas = {
      ["https://gitlab.com/gitlab-org/gitlab/-/raw/master/app/assets/javascripts/editor/schema/ci.json"] = ".gitlab-ci.y*l",
    },
    hover = true,
  }
}

vim.api.nvim_create_autocmd('FileType', {
  pattern = "yaml",
  callback = function(args)
    vim.lsp.start({
      name = 'yamlls',
      cmd = { 'yaml-language-server', '--stdio' },
      root_dir = vim.fn.fnamemodify('./.gitlab-ci.yaml', ':p:h'),
      settings = settings
    })
  end
})

vim.keymap.set('n', '<space>', vim.lsp.buf.hover, {})
```

- run `cd $(mktemp -d)`, and copy the contents above to a file called `init.nvim`.

- create a file called `.gitlab-ci.yaml` with the following contents
```yaml
stages:
  - hello
```

- run neovim (`v0.8.3` in my case) with `nvim --noplugin -u init.nvim .gitlab-ci.yaml`.

- give it a couple of seconds to spin up `yaml-language-server` and press `<space>` with the cursor on top of the `stages` word.

image

- you can then apply the patch I showed in my last message to reproduce the "correct" behavior. (change the `gsub` call to replace `"&emsp;"` with `""`)

EdgeJ:
One small tweak to @Kasama's fix that will account for instances where there is no LSP information for a hover (e.g. a yaml file without an accompanying schema):
```
 local function hover_wrapper(err, request, ctx, config)
  local bufnr, winnr = vim.lsp.handlers.hover(err, request, ctx, config)
  if bufnr then
    local contents = vim.api.nvim_buf_get_lines(bufnr, 0, -1, false)
    contents = vim.tbl_map(
      function(line)
        return string.gsub(line, "&emsp;", "")
      end,
      contents
    )
    contents = vim.lsp.util.trim_empty_lines(contents)
    vim.api.nvim_buf_set_option(bufnr, 'modifiable', true)
    vim.api.nvim_buf_set_lines(bufnr, 0, -1, false, contents)
    vim.api.nvim_buf_set_option(bufnr, 'modifiable', false)
    vim.api.nvim_win_set_height(winnr, #contents)
  end
end

vim.lsp.handlers["textDocument/hover"] = hover_wrapper
```

Prevents issues like:
```
No information available
Error executing vim.schedule lua callback: ../.config/nvim/lsp.lua:82 Expected lua number
stack traceback:
    [C]: in function 'nvim_buf_get_lines'
    .../.config/nvim/lsp.lua:82: in function 'handler'
```

PR #846, opened 2023-02-16
Fix anchor diagnostics

kraftwerk28:
### What does this PR do?
Fixes broken warnings on unused anchors inside an array:
image

### Is it tested? How?
Test case included in PR.

### DIFF
--- a/src/languageservice/services/validation/unused-anchors.ts
+++ b/src/languageservice/services/validation/unused-anchors.ts
@@ -33,7 +33,7 @@ export class UnusedAnchorsValidator implements AdditionalValidator {
 
     for (const anchor of anchors) {
       if (!usedAnchors.has(anchor)) {
-        const aToken = this.getAnchorNode(anchorParent.get(anchor));
+        const aToken = this.getAnchorNode(anchorParent.get(anchor), anchor);
         if (aToken) {
           const range = Range.create(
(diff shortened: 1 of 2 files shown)


test/schemaValidation.test.ts
```diff
@@ -1966,5 +1966,17 @@ obj:
       expect(result[0].message).to.eq('Matches multiple schemas when only one must validate.');
       expect(telemetry.messages).to.be.empty;
     });
+
+    it('Unused anchors in an array', async () => {
```
gorkem, line 1970: This test should be in https://github.com/redhat-developer/yaml-language-server/blob/ba7452ee94105636f7c517394f378d09a71e2f6b/test/yamlValidation.test.ts#L58

gorkem (changes requested):
Thank you for the contribution. If you can move the test to the correct place, it is good to go for me.

test/schemaValidation.test.ts
kraftwerk28: Sure, I've updated an existing test case to have multiple [un]used anchors handling in an array.

gorkem (approved):
LGTM

gorkem:
@msivasubramaniaan Can you review and merge this one if it is all right?

msivasubramaniaan (approved):
LGTM

PR #847, opened 2023-02-17
Pin 3rd-party actions to SHA1

fbricon:
Hi!

Following the [GH Action Security Hardening](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) guide we should use the commit SHA instead of the `branch` or `tag` for any third-party untrusted action.

This PR was submitted by a script.


msivasubramaniaan (approved):
LGTM

PR #852, opened 2023-02-27
Add API to update AST tree from internal document

gorkem:
### What does this PR do?

Adds a new API to update the internal AST from the internal document without requiring re-assigning the internal document.

Fixes eslint issues about self-assigning 

### Is it tested? How?
Existing test set


msivasubramaniaan (approved):
LGTM

Issue #845, opened 2023-02-14
Type errors inform wrong type when parent schema has keyword "title"


nieomylnieja:
## Describe the bug
It seems that using the `title` keyword propagates the element name to all it's properties which also have a `type` of `object`.
I verified the behaviour against another JSON Schema implementation: https://www.jsonschemavalidator.net/.
Given the following schema:

```yaml
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "$id": "file:///home/mh/schema/test.schema.json",
  "type": "object",
  "title": "THE OBJECT",
  "properties": {
    "foo": {
      "type": "object"
    },
    "bar": {
      "type": "string"
    }
  }
}
```

It produced the following error:
![2023-02-14_17-23](https://user-images.githubusercontent.com/48822818/218798417-d38b0bcb-327f-401d-9921-951877185a6c.png)
Which makes absolute sense, since the type of the property `foo` is `object`.
However that's not the case with yaml-language-server, which instead treats all `type` errors for `object` properties of a schema with set `title` keyword as If they had the same type as the parent schema, i.e. `THE OBJECT`.
![2023-02-14_17-24](https://user-images.githubusercontent.com/48822818/218799244-2944715d-1c44-4508-9547-66f8a8c11a72.png)
It only happens when the property is of type `object`, If we now change `bar` to an invalid type, let's say `object` it will produce the right error, telling us it should be a string:
![2023-02-14_17-34](https://user-images.githubusercontent.com/48822818/218799837-a992cb3f-1fad-418c-af44-e4dc3eebaab3.png)

## Expected Behavior
Displayed error should be referring to the property `type`, not the parent schema `type`.

## Current Behavior
Displayed error informs about parent schema `type`, instead of the property `type` mismatch.

## Steps to Reproduce
Copy the schema provided in the description and play around with it as described.

## Environment
- [ ] Windows
- [ ] Mac
- [x] Linux
- [ ] other (please specify)

nieomylnieja:
I did some preliminary code digging and it looks like `validate` function in `jsonParser07.ts` walks over it's properties propagating the parent schema, when `_validateNode` is called, it does this:

```ts
//get more specific name than just object
const schemaType = schema.type === 'object' ? getSchemaTypeName(schema) : schema.type;
```

`getSchemaTypeName` accepts the parent schema and eventually returns it's title as the `schemaType` which is used to construct the error:

```ts
if (schema.title) {
  return schema.title;
}
```

[...]

PR #853, opened 2023-02-27
provide preference on schema type than its title

msivasubramaniaan:
### What does this PR do?

This PR shows schema type when mismatch error than its title

https://github.com/redhat-developer/yaml-language-server/issues/845

### Is it tested? How?
Yes. Updated existing UT


rgrunber (approved):
Works for me. @msivasubramaniaan . closestTitle should probably a fallback in the event we can't get the type.

PR #854, opened 2023-03-02
845 type errors inform wrong type when parent schema has keyword title

msivasubramaniaan:
### What does this PR do?
This PR shows schema type when mismatch error along with it title
image

https://github.com/redhat-developer/yaml-language-server/issues/845

### Is it tested? How?
Yes. Added UT


rgrunber (approved):
Makes sense. The extra bit of information is nice to have.

rgrunber (approved):
Makes sense to me. Thanks for catching the case of schema.type being undefined.

PR #863, opened 2023-03-26
fix(indentation): hover behavior when indentation not set

Kasama:
### What does this PR do?

The previous behavior for hover replaces indentations with the `&emsp;` entity, because Markdown treats whitespace at the beginning of a line in a special way. However, when the indentation setting wasn't set, the regex that does that replace was set to empty (`""`) and so would include an `&emsp;` entity between every character.

This went under the radar because the previous PR #844 also changed the tests setup to include a `.withIndentation` call, which made sure all tests were passing.

In this PR I'm reverting that to the default behavior for most tests, but overriding the default only on the test introduced in #844.

The issue is referenced on a comment of PR #844. [here](https://github.com/redhat-developer/yaml-language-server/pull/844#issuecomment-1480559728)

### Is it tested? How?

I have tested this fix with a neovim setup as described in the original problem report in [the PR comment](https://github.com/redhat-developer/yaml-language-server/pull/844#issuecomment-1480591337).

At the risk of simply repeating myself, below is the full set of reproduction steps:

<details>
  <summary>Open for steps</summary>
  
1. Download neovim (I used v0.8.3)
2. Create a new folder `cd $(mktemp)`
3. create a file named `init.lua` with the following contents:

```lua
local settings = {
  yaml = {
    schemas = {
      ["https://gitlab.com/gitlab-org/gitlab/-/raw/master/app/assets/javascripts/editor/schema/ci.json"] = ".gitlab-ci.y*l",
    },
    hover = true,
  }
}

vim.api.nvim_create_autocmd('FileType', {
  pattern = "yaml",
  callback = function(args)
    vim.lsp.start({
      name = 'yamlls',
      cmd = { 'yaml-language-server', '--stdio' }, -- replace with local installation as necessary
      root_dir = vim.fn.fnamemodify('./.gitlab-ci.yaml', ':p:h'),
      settings = settings
    })
  end
})

vim.keymap.set('n', '<space>', vim.lsp.buf.hover, {})
```

5. create a file called `.gitlab-ci.yaml` with the following contents
```yaml
stages:
  - hello
```

6. run neovim with `nvim --noplugin -u init.lua .gitlab-ci.yaml`.

7. give it a couple of seconds to spin up `yaml-language-server` and press `<space>` with the cursor on top of the `stages` word.

image

8. With this PR patch, the result is the following
image

</details>

### DIFF
--- a/test/hover.test.ts
+++ b/test/hover.test.ts
@@ -20,13 +20,10 @@ describe('Hover Tests', () => {
   let telemetry: TestTelemetry;
 
   before(() => {
-    languageSettingsSetup = new ServiceSetup()
-      .withHover()
-      .withIndentation('  ')
-      .withSchemaFileMatch({
-        uri: 'http://google.com',
-        fileMatch: ['bad-schema.yaml'],
-      });
+    languageSettingsSetup = new ServiceSetup().withHover().withSchemaFileMatch({
+      uri: 'http://google.com',
+      fileMatch: ['bad-schema.yaml'],
+    });
     const {
       languageService: langService,
       languageHandler: langHandler,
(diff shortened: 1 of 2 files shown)


polyzen (dismissed):

polyzen:
Works for me. Backporting for the Arch Linux package ❤️

gorkem:
can you rebase the PR?

Kasama:
It seems the actual issue has been fixed at [#831](https://github.com/redhat-developer/yaml-language-server/pull/831/files#diff-560588b83f4b671b671ab7660e678cb1f6be1de7c34378d5d9eeb9d3947c0e42R91), but I think the test changes are still relevant.

WYT?

I have rebased it, anyway

gorkem:
CI failed with a prettier error. I did not notice that #831 fixed this on the side but did not provide any tests. This is a good complement PR, and I would like to merge it if you can update it for the prettier fix

gorkem (approved):
lgtm

Kasama:
Nice, would be great to have a release soon, so we can have this fix rolled out!

Thanks for the help!

PR #866, opened 2023-04-01
Run sarif seperately

gorkem:
### What does this PR do?
Runs SARIF generation separate from the build. Also uploads only once on matrix build


msivasubramaniaan (approved):
Lgtm

PR #882, opened 2023-05-17
fix: array item with existing property

p-spacek:
### What does this PR do?
fix autocompletion when the position is on the same line with the hyphen followed by other props

there are two problems:
1) intellisense doesn't work correctly for the first item when there is another prop in the array item
Screenshot 2023-05-17 at 09 37 56
Screenshot 2023-05-17 at 09 37 24

https://github.com/redhat-developer/yaml-language-server/assets/38421337/97997ad7-06dc-44f1-89ff-4581859f7ee5

2) the compensation of the indentation is not correct for the first item after the hyphen

https://github.com/redhat-developer/yaml-language-server/assets/38421337/46f4a5fe-a450-45b6-bc6e-77f5e982d8a4

after fix of the 1st problem, the 2nd problem looks like this 

https://github.com/redhat-developer/yaml-language-server/assets/38421337/2f7b6430-b4ac-49de-93e3-98281e9f7911

after fix:

https://github.com/redhat-developer/yaml-language-server/assets/38421337/f04e5ea0-6eda-4586-9464-d4fa41aaef5f

no reference

### Is it tested? How?
add unit tests


src/languageservice/services/yamlCompletion.ts
```diff
@@ -144,11 +144,11 @@ export class YamlCompletion {
 
     this.arrayPrefixIndentation = '';
     let overwriteRange: Range = null;
+    const isOnlyHyphen = lineContent.match(/^\s*(-)\s*$/);
```
gorkem, line 147: I am not sure if doing this with RegEx is a good idea. For instance this fails on a line like `  -   # a comment `. Can we use the AST to check if the rest of the line does not have any semantically significant tokens?

gorkem (changes requested):

src/languageservice/services/yamlCompletion.ts
p-spacek: it was there from the previous implementation, I just reused it (moved it up). I didn't want to change it.

But agree, 
I went a little bit deeper and there is another problem.

The real problem is in `getNodeFromPosition`
This method doesn't give proper results for arrays and maybe some others. That is the reason why there has to be custom code to 'fix' proper node based on special conditions 'empty lines, arrays, ...' (lines 343 to 515 in yamlCompletion)

for example, this yaml gives these results:
foo:
  - # foo object is returned (should be foo[0])
    # foo object is returned (should be foo[0])
    item1: aaaf
    # foo[0] object is returned (OK)
  - # foo object is returned (should be foo[1])
    # foo[!!0!!] object is returned (should be foo[1])
    item2: bbb
    # foo[1] object is returned (OK)

So this PR doesn't make sense anymore (it's possible to reuse tests)

The question is whether your team has the capacity to fix it.
If not I will try it in the future, but don't have much time for it :-/

p-spacek:
@gorkem can you please consider this PR

it will also fix another issue (see updated description 3) )

I think that `getNodeFromPosition` doesn't work correctly with arrays (check the previous comment) so it's not possible to detect an array by AST (as you suggested),

but this PR doesn't try to reimplement it, this PR fixes a few issues that are caused by a combination of:
 - how the node is selected (`getNodeFromPosition`)
 - how is the indent array calculated

PR #888, opened 2023-06-02
fix: nested anyof const

p-spacek:
### What does this PR do?
fix validation of nested `anyOf` consts
schema:
```json
{
  "type": "object",
  "properties": {
    "options": {
      "anyOf": [
        {
          "type": "object",
          "properties": {
            "form": {
              "type": "string"
            },
            "provider": {
              "type": "string",
              "const": "test1"
            }
          },
          "required": [
            "form",
            "provider"
          ]
        },
        {
          "type": "object",
          "properties": {
            "form": {
              "type": "string"
            },
            "provider": {
              "anyOf": [
                {
                  "type": "string",
                  "const": "testX"
                }
              ]
            }
          },
          "required": [
            "form",
            "provider"
          ]
        }
      ]
    }
  }
}
```

will produce this error:
image

https://github.com/redhat-developer/yaml-language-server/assets/38421337/8524d0ea-5adf-453a-bbe4-e42a3464c4d8

note that 'missing form property' is necessary to trigger the bug (there has to be some problem in both subSchemas)
order of the object in `anyOf` is also important

no existing issue

### Is it tested? How?
unit test

### DIFF
--- a/test/schemaValidation.test.ts
+++ b/test/schemaValidation.test.ts
@@ -2016,7 +2016,7 @@ obj:
         },
       },
     };
-    languageService.addSchema(SCHEMA_ID, schema);
+    schemaProvider.addSchema(SCHEMA_ID, schema);
     const content = `options:\n  provider: testX`;
     const result = await parseSetup(content);
     assert.deepEqual(
--- a/test/schemaValidation.test.ts
+++ b/test/schemaValidation.test.ts
@@ -1974,4 +1974,54 @@ obj:
       expect(telemetry.messages).to.be.empty;
     });
   });
+  it('Nested AnyOf const should correctly evaluate and merge problems', async () => {
+    // note that 'missing form property' is necessary to trigger the bug (there has to be some problem in both subSchemas)
+    // order of the object in `anyOf` is also important
+    const schema: JSONSchema = {
+      type: 'object',
+      properties: {
+        options: {
+          anyOf: [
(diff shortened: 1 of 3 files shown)


test/schemaValidation.test.ts
```diff
@@ -1967,4 +1967,54 @@ obj:
...
+                  ],
+                },
+              },
+              required: ['form', 'provider'],
+            },
+          ],
+        },
+      },
+    };
+    languageService.addSchema(SCHEMA_ID, schema);
```
gorkem, line 2012: Can you rebase this commit to head and use `schemaProvider.addSchema` it is the new mechanism for feeding schemas for tests and since we have removed the earlier hack in the parser this will probably fail.

gorkem (changes requested):
This looks good. If you can rebase and adjust to the  testing changes main we are good to go.

test/schemaValidation.test.ts
p-spacek: thanks, @gorkem done,

I also pushed a small change for const matching.
there was a small inconsistency between `enum` and `const` match.
Enum allows having a partial match, but const had to be exactly the same.

It's a little bit related to this PR, so sorry for putting it together...

gorkem (approved):
lgtm

msivasubramaniaan (commented):
LGTM

PR #900, opened 2023-07-10
feat: quickFix for enum, const, property

p-spacek:
### What does this PR do?
add quick fix code-action for enums, consts, properties

Screenshot 2023-07-10 at 13 56 00

Screenshot 2023-07-10 at 13 55 16

no ref

### Is it tested? How?
tests

### DIFF
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -888,6 +888,7 @@ function validate(
             ),
           source: getSchemaSource(schema, originalSchema),
           schemaUri: getSchemaUri(schema, originalSchema),
+          data: { values: schema.enum },
         });
       }
     }
@@ -907,6 +908,7 @@ function validate(
           source: getSchemaSource(schema, originalSchema),
           schemaUri: getSchemaUri(schema, originalSchema),
           problemArgs: [JSON.stringify(schema.const)],
+          data: { values: [schema.const] },
         });
         validationResult.enumValueMatch = false;
       } else {
@@ -1385,6 +1387,7 @@ function validate(
                 length: propertyNode.keyNode.length,
               },
               severity: DiagnosticSeverity.Warning,
+              code: ErrorCode.PropertyExpected,
               message: schema.errorMessage || localize('DisallowedExtraPropWarning', MSG_PROPERTY_NOT_ALLOWED, propertyName),
               source: getSchemaSource(schema, originalSchema),
               schemaUri: getSchemaUri(schema, originalSchema),
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -400,7 +400,7 @@ export class ValidationResult {
               (problemType !== ProblemType.missingRequiredPropWarning || isArrayEqual(p.problemArgs, bestResult.problemArgs)) // missingProp is merged only with same problemArg
           );
           if (mergingResult) {
-            if (mergingResult.problemArgs.length) {
+            if (mergingResult.problemArgs?.length) {
               mergingResult.problemArgs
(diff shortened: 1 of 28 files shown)


src/languageservice/parser/jsonParser07.ts
```diff
@@ -1385,6 +1387,7 @@ function validate(
                 length: propertyNode.keyNode.length,
               },
               severity: DiagnosticSeverity.Warning,
+              code: ErrorCode.PropertyExpected,
```
p-spacek, line 1390: not sure if it wouldn't be better to unify `data.properties` to `data.values` for the incorrect property

p-spacek:
Hello @msivasubramaniaan, can I ask you for a review, please?

src/languageservice/services/yamlCodeActions.ts
```diff
@@ -312,9 +316,52 @@ export class YamlCodeActions {
...
+    if (typeof diagnostic.data !== 'object') {
+      return;
+    }
+    if (
+      diagnostic.code === ErrorCode.EnumValueMismatch &&
+      'values' in diagnostic.data &&
+      Array.isArray((diagnostic.data as YamlDiagnosticData).values)
+    ) {
+      return (diagnostic.data as YamlDiagnosticData).values;
+    } else if (
```
gorkem, line 334: Is this else really needed?

src/languageservice/services/yamlCodeActions.ts
```diff
@@ -312,9 +316,52 @@ export class YamlCodeActions {
...
+    const results: CodeAction[] = [];
+    for (const diagnostic of diagnostics) {
+      const values = this.getPossibleQuickFixValues(diagnostic);
+      if (!values?.length) {
+        continue;
+      }
+      for (const value of values) {
+        results.push(
+          CodeAction.create(
+            value,
```
gorkem, line 353: This returns only the value which is not very descriptive of the operation. perhaps we can replace it with something like 
```suggestion
            `Use value "${value}"`,
```

gorkem (changes requested):

gorkem (changes requested):
Only a couple of minor comments. It also needs to be rebased.

gorkem (changes requested):
Only a couple of minor comments. It also needs to be rebased.

src/languageservice/services/yamlCodeActions.ts
p-spacek: Do you mean quickFix for properties? or can you explain your point further?
 - I think that property quickFix can be useful
 - note that there are different property names in `diagnostic.data` object (`values` vs `properties`)

src/languageservice/services/yamlCodeActions.ts
p-spacek: I don't have preferences here, I can remember that I tried a few combinations but I ended up with only `{value}`.
I was inspired by the English spelling checker extension (Code Spell Checker extension)
image

small disadvantage of `Use value "${value}"` is that this PR adds quickfix also for properties of the object (not value), so we need to display the text based on the error type
 - but for sure it's not a problem
 
 so please make a decision

p-spacek:
@gorkem any updates here, please?
check updates and answers
Can you please rerun the jobs? It failed because of unavailable  jsonSchemaStore

gorkem (approved):

msivasubramaniaan:
@p-spacek Please upstream the branch

p-spacek:
> @p-spacek Please upstream the branch

@msivasubramaniaan I merged the main into this PR, but there is some schema change that caused the test to fail:

      -  "source": "yaml-schema: Composer Package"
      +  "source": "yaml-schema: Package"

PR #901, opened 2023-07-11
fix: suggest hyphen in array

p-spacek:
### What does this PR do?
`- (array item)` suggestion was missing for some specific situations, for example: enums.
it worked ok for 1st item after a colon, but not for 2nd

after fix:
image
    

no ref

### Is it tested? How?
modified existing
added one test for the enum


test/autoCompletion.test.ts
```diff
@@ -1488,6 +1488,39 @@ describe('Auto Completion Tests', () => {
...
+        });
+        const content = 'references:\n  - Test\n  |\n|';
+        const completion = parseSetup(content);
+        completion
+          .then(function (result) {
+            assert.deepEqual(
+              result.items.map((i) => ({ label: i.label, insertText: i.insertText })),
+              [
+                {
+                  insertText: 'Test',
```
p-spacek, line 1511: there should be '- Test' insert text, but it's another problem

src/languageservice/services/yamlCompletion.ts
```diff
@@ -910,13 +910,8 @@ export class YamlCompletion {
...
                   this.addSchemaValueCompletions(s.schema.items[index], separatorAfter, collector, types, 'value');
                 }
-              } else if (
-                typeof s.schema.items === 'object' &&
-                (s.schema.items.type === 'object' || isAnyOfAllOfOneOfType(s.schema.items))
-              ) {
-                this.addSchemaValueCompletions(s.schema.items, separatorAfter, collector, types, 'value', true);
               } else {
-                this.addSchemaValueCompletions(s.schema.items, separatorAfter, collector, types, 'value');
+                this.addSchemaValueCompletions(s.schema.items, separatorAfter, collector, types, 'value', true);
```
p-spacek, line 914: unify the code for both scenarios. If a schema contains `items` it's an array. No need to check if `type === 'object'`

p-spacek:
Hello @msivasubramaniaan, can I ask you for a review?

test/autoCompletion.test.ts
gorkem: Can you file that as an issue?

gorkem (dismissed):
LGTM. 

Needs to be rebased before I can merge and the failing CI (which looks like a GH problem) should clear with the rebase. 

Another level of improvement to this PR would be to filter the items that are already on the array to avoid duplicate entries on the array.  Approving this PR but please let me know if you are planning to add the improvement.

p-spacek:
@gorkem , I don't think that there should be such a filter/condition
because schema
```
{
          type: 'object',
          properties: {
            references: {
              type: 'array',
              items: {
                enum: ['Test', 'Test2'],
              },
            },
}
```
doesn't define that items of the array have to be unique. To be honest I am not sure how to define this (unique array items)...

This should be allowed based on the schema
```yaml
items:
 - Test1
 - Test1
 - Test2
```

or don't you agree? did I understand you incorrectly?

test/autoCompletion.test.ts
p-spacek: it's only a small detail I think, but logged the issue
https://github.com/redhat-developer/yaml-language-server/issues/930

p-spacek:
but not sure what's wrong with tests, 
```
 Multiple Documents Validation Tests
       Multiple Documents Validation
         Should find errors in both documents:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/petr/Documents/Jigx/git/rh-yaml-language-server/test/multipleDocuments.test.ts)
      at listOnTimeout (node:internal/timers:564:17)
      at processTimers (node:internal/timers:507:7)
```
it's failing also on main branch

p-spacek:
I found another related problem
So I converted this PR to a draft
I'll try to fix it at once

gorkem:
Correct, item uniqueness is determined by the [schema](https://json-schema.org/understanding-json-schema/reference/array#uniqueItems). The current implementation does not take into account the `uniqueItems` flag. An improvement to the current implementation would be checking this flag and filtering accordingly. In general, code assist should not suggest values leading to syntax errors, even though it is a trivial fix.

PR #909, opened 2023-08-05
Update to Prettier 3

remcohaszing:
### What does this PR do?

This updates Prettier to version 3.

- Prettier 3 is async, so related functions were made async.
- Code was reformatted using Prettier.
- Prettier now ships their own types, so `@types/prettier` was removed.
- Prettier was moved to `dependencies`. `optionalDependencies` is for dependencies that have a `install` script that may fail. This isn’t the case for Prettier.
- `eslint-plugin-prettier` was updated as well.

Alternatively the formatting functionality could be removed. IMO people should just use Prettier themselves for formatting YAML instead of relying on this `yaml-language-server`.

N/A

### Is it tested? How?

```sh
npm test
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,12 +29,10 @@
     "type": "git",
     "url": "https://github.com/redhat-developer/yaml-language-server.git"
   },
-  "optionalDependencies": {
-    "prettier": "2.8.7"
-  },
   "dependencies": {
     "ajv": "^8.11.0",
     "lodash": "4.17.21",
+    "prettier": "^3.0.0",
     "request-light": "^0.5.7",
     "vscode-json-languageservice": "4.1.8",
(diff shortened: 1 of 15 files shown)


artola:
@msivasubramaniaan This would be a nice enhancement to support Prettier v3, otherwise we are forced deliver 2 versions of Prettier within our app.

msivasubramaniaan (changes requested):
Seeing lots of lint changes. Kindly rebase and raise PR only with affected files

remcohaszing:
Done, but now CI fails. It’s hard to tell intent, because both ESLint are Prettier are not configured correctly for this repo.

msivasubramaniaan (approved):
LGTM

PR #911, opened 2023-08-10
fix: doNotSuggest for property and value completion

p-spacek:
### What does this PR do?
exclude doNotSuggest schema from autocompletion
`doNotSuggest` was already there but implemented only for the simple object properties.
This PR uses `doNotSuggest` also for other scenarios: anyOf, values

```json
"anyOf":[
  {
     "$ref": "schema1",
     "doNotSuggest": true
   },
  {
     "$ref": "schema2",
   }
]
```

related to the previous unfinished PR for the deprecated property:
https://github.com/redhat-developer/yaml-language-server/pull/833

### Is it tested? How?
adds unit tests


p-spacek:
Hello @msivasubramaniaan, can I ask you for a review, please?

gorkem:
We do inherit the `doNotSuggest` property from vscode json language service. As far as I can see on the JSON land it only applies to properties. Although I think this is an interesting usage of it.

@aeschli Is this something interesting for vscode json language service too?

p-spacek:
> Yes, if you want to make a PR against the json-language-service that would be great. Note that `doNotSuggest` is internal and we don't want to advertise this. But if you can use it internally, why not.

Hi, I created PR for this functionality https://github.com/microsoft/vscode-json-languageservice/pull/213
but I found a not-related issue probably, see the linked PR.

PR #917, opened 2023-09-06
Export Telemetry type

remcohaszing:
### What does this PR do?

Export the `Telemetry` type.

Alternatively it could be defined in `yamlLanguageService.ts`.

This is used by `monaco-yaml`.

### Is it tested? How?

No.


gorkem:
Can we alternately add `telemetry.ts` to `index.ts` ? This would also export `TelemetryEvent` which I feel maybe needed if a client really wants to send events.

remcohaszing:
I think you meant to move it into `yamlLanguageService`, of which everything is also re-exported in `index.ts`. Otherwise language service code would have to import language server code.

PR #918, opened 2023-09-06
Fix some small type issues

remcohaszing:
### What does this PR do?

Some parameters are optional, but were marked as required.

Language service methods were marked as bound methods, but they don’t have to be called as such.

The `Thenable` type from `json-rpc` was used. These were replaced with the builtin TypeScript type `PromiseLike`.

N/A

### Is it tested? How?

```sh
tsc
```


gorkem (approved):
lgtm

PR #921, opened 2023-09-08
Remove ide-yaml from known clients

remcohaszing:
### What does this PR do?

Atom has been sunset. The link is no longer available.  Clicking the link would take the user to a GitHub blog post about this instead.

N/A

### Is it tested? How?

Clicking the removed link.


gorkem (approved):
lgtm

PR #923, opened 2023-09-14
fix: schema loading performance

p-spacek:
### What does this PR do?
reduce the time to resolve schemas content
replace `Array` with `Set`
performance of checking if the `Set.has` item is better than `Array.indexOf`

no ref

### Is it tested? How?
manually with +-20 large schemas (connected between each other)
loading time from 3.5sec decreased to 1.5sec
`yarn test` is one second (20sec->19sec) faster on my side - but it's not really a reliable claim :)


gorkem (approved):
lgtm

gorkem:
rebased and pushed

PR #926, opened 2023-09-30
Drop lodash

zulus:
### What does this PR do?
Reduce number files in node_modules (problematic on windows)

Reduce time necessary to extract zipped node_modules on windows

### Is it tested? How?
I used modified version in wildwebdeveloper

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,8 +33,8 @@
     "prettier": "2.8.7"
   },
   "dependencies": {
+    "@types/lodash.clonedeep": "^4.5.7",
     "ajv": "^8.11.0",
-    "lodash": "4.17.21",
     "request-light": "^0.5.7",
     "vscode-json-languageservice": "4.1.8",
     "vscode-languageserver": "^7.0.0",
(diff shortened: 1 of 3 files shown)


gorkem (changes requested):
It looks like CI fails with these changes. Can you check?

package.json
```diff
@@ -33,6 +33,7 @@
     "prettier": "2.8.7"
   },
   "dependencies": {
+    "@types/lodash.clonedeep": "^4.5.7",
```
B2o5T, line 36: should be devDependency

src/languageservice/services/yamlCodeActions.ts
```diff
@@ -26,7 +26,7 @@ import { CST, isMap, isSeq, YAMLMap } from 'yaml';
 import { yamlDocumentsCache } from '../parser/yaml-documents';
 import { FlowStyleRewriter } from '../utils/flow-style-rewriter';
 import { ASTNode } from '../jsonASTTypes';
-import * as _ from 'lodash';
+import cloneDeep  from '@types/lodash.clonedeep';
```
B2o5T, line 29:
```suggestion
import cloneDeep  from 'lodash.clonedeep';
```

PR #929, opened 2023-10-07
Remove 'use strict'

remcohaszing:
### What does this PR do?

This removes all `'use strict'` directives. TypeScript adds this for us.

N/A

### Is it tested? How?

Inspect the output.


gorkem (approved):
lgtm

PR #934, opened 2023-11-08
feat: unify string insert text for array and property

p-spacek:
### What does this PR do?
unify string text inserting for array and property

when you choose string property from completion you get
`na` -> `name: `

but for array, it generated extra quotes:
`arr` -> 
```
array:
  - ""
```

These extra quotes caused, that it wasn't possible to invoke completion just after the code insert, but you had to delete `""` continue...

the behavior is just a little bit different than for property text inserting.

https://github.com/redhat-developer/yaml-language-server/assets/38421337/8d73fb4a-8b10-49d9-ac27-ac37362d89c4

Note that I didn't change it for `number` nor for `bool` because it seems to me a little bit different scenario

### Is it tested? How?
update current tests

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -1205,7 +1205,7 @@ export class YamlCompletion {
         insertText = `\${${insertIndex++}:0}`;
         break;
       case 'string':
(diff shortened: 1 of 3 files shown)


msivasubramaniaan (approved):
LGTM

PR #942, opened 2023-12-03
fix: show all enums on hover

### LINKED ISSUE #941: Show enum values on hover

I maintain a JSON Schema document for a YAML config file. That document has a number of ENUMs, and when my users hover over those fields they see the title and description from the schema, but not the allowed values.

The values are shown on autocomplete, but that's much less discoverable that having it displayed on hover w/ the other documentation.

I would like to see enum options shown underneath the description on hover - much like how `example` values are displayed:

```
ENUM TITLE

Some description...


Allowed values:

- `ENUM1`
- `ENUM2`
- `ENUM3`
```

twelvelabs:
### What does this PR do?

This PR updates the hover service to display all possible enum values, rather than just the currently selected one. Additionally, it will display the enums regardless of whether the schema contains explicit descriptions (though it will continue to display the descriptions if present - see included test case).

Screen Shot 2023-12-03 at 3 56 05 PM

<br />
<br />

Also: 

- Included a minor change to how examples are displayed (i.e. added markdown bullets) to make them visually consistent with the enums.
- Added a `ensureLineBreak` helper to make sure there is consistent spacing between each of the various hover sections.

Fixes #941

### Is it tested? How?

- Tested locally with https://github.com/twelvelabs/stamp/blob/main/docs/stamp.schema.json
- See test cases in `test/hover.test.ts`

### DIFF
--- a/src/languageservice/services/yamlHover.ts
+++ b/src/languageservice/services/yamlHover.ts
@@ -12,7 +12,7 @@ import { setKubernetesParserOption } from '../parser/isKubernetes';
 import { TextDocument } from 'vscode-languageserver-textdocument';
 import { yamlDocumentsCache } from '../parser/yaml-documents';
 import { SingleYAMLDocument } from '../parser/yamlParser07';
-import { getNodeValue, IApplicableSchema } from '../parser/jsonParser07';
+import { IApplicableSchema } from '../parser/jsonParser07';
 import { JSONSchema } from '../jsonSchema';
 import { URI } from 'vscode-uri';
 import * as path from 'path';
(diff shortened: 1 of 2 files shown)


twelvelabs:
cc @gorkem and @msivasubramaniaan for visibility.

msivasubramaniaan (approved):
LGTM

Thanks @twelvelabs for your contribution

prose-leviathan:
@msivasubramaniaan @twelvelabs This feature really broke our workflow, we use very large list of enums (and markdownEnumDescriptions), and now hover will not work as it tries to render the full list of enum instead of trying to show the markdownEnumDescription. I suggest to put this change behind a setting which can be disabled.

PR #947, opened 2023-12-27
fix: render examples as yaml on hover

twelvelabs:
### What does this PR do?

This MR improves the rendering of examples on hover by:

1. Rendering the examples as YAML rather than JSON.
2. Adding an info string to the [fenced code block](https://github.github.com/gfm/#fenced-code-blocks) so that the examples are properly syntax highlighted. According to the LSP spec, [markdown content should be expected to be GFM](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContent), so this shouldn't be an issue for clients.
3. Moving the indentation escaping logic to `toMarkdown`, so that we only escape non-markdown strings. Currently the escaping is done on the entire hover content, regardless of whether that content contains actual markdown. This change also fixes a bug where intentional indentation in `markdownDescription` is erroneously escaped.

Before:

Screenshot 2023-12-27 at 3 17 58 PM

After:

Screenshot 2023-12-27 at 3 12 24 PM

Note: I originally tried to keep the single header (rather than repeating it per-example), but the code fences don't appear to have a margin applied so the examples were all bunched together. Repeating the header looked much nicer.

- Fixes: redhat-developer/vscode-yaml#896
- Fixes: redhat-developer/vscode-yaml#959
- Fixes: redhat-developer/vscode-yaml#960
- Fixes: redhat-developer/vscode-yaml#961
- Fixes: redhat-developer/vscode-yaml#965

### Is it tested? How?

I updated an existing unit test. Also tested manually with the following (see screenshots):

- `schema.json`:

```json
{
    "$id": "schema.json",
    "$schema": "https://json-schema.org/draft-07/schema",
    "title": "Example Schema",
    "properties": {
        "example": {
            "title": "Example Property",
            "description": "This property has examples that should be rendered as YAML code blocks.",
            "examples": [
                "string value",
                {
                    "object_value": {
                        "enabled": true
                    }
                }
            ],
            "type": [
                "object",
                "string"
            ]
        }
    },
    "type": "object"
}
```

- `example.yaml`:

```yaml
# yaml-language-server: $schema=schema.json
example: "hello"
```

### DIFF
--- a/src/languageservice/services/yamlHover.ts
+++ b/src/languageservice/services/yamlHover.ts
@@ -19,6 +19,7 @@ import * as path from 'path';
 import { Telemetry } from '../telemetry';
 import { convertErrorToTelemetryMsg } from '../utils/objects';
 import { ASTNode } from 'vscode-json-languageservice';
+import { stringify as stringifyYAML } from 'yaml';
 
 export class YAMLHover {
   private shouldHover: boolean;
@@ -87,11 +88,6 @@ export class YAMLHover {
     );
 
     const createHover = (contents: string): Hover => {
-      if (this.indentation !== undefined) {
-        const indentationMatchRegex = new RegExp(` {${this.indentation.length}}`, 'g');
-        contents = contents.replace(indentationMatchRegex, '&emsp;');
-      }
-
       const markupContent: MarkupContent = {
         kind: MarkupKind.Markdown,
         value: contents,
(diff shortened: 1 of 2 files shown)


twelvelabs:
ping @msivasubramaniaan and @gorkem

msivasubramaniaan (approved):
LGTM

PR #955, opened 2024-03-21
Add null check for `customTags`

### LINKED ISSUE #807: Visual Studio Client Integration: Server throws cannot read properties of undefined (reading 'length')
## Describe the bug
When integrating the Yaml Language Server with Visual Studio I faced an error during the 'textDocument/completion' call to the server. 
The error is outmsg = "{\"jsonrpc\":\"2.0\",\"id\":9,\"error\":{\"code\":-32603,\"message\":\"Request textDocument/completion failed with message: Cannot read properties of undefined (reading 'length')\"}}"
The version of yaml language server is 1.10.0

I tried to attach the yaml language server nodejs process and found that the error is thrown at line 283 of this file https://github.com/redhat-developer/yaml-language-server/blob/1.10.0/src/languageservice/services/yamlCompletion.ts

I fixed this error by changing the condition to 'if (**this.customTags** && this.customTags.length > 0) {'
The schema is fine and gets validated with the same language server. I am happy to provide if required. 

## Expected Behavior
Even if customTags are not defined the completion should work successfully.

## Current Behavior
It throws an error "Request textDocument/completion failed with message: Cannot read properties of undefined (reading 'length')\"}}"

## Steps to Reproduce
1. Setup a new VS client 
2. Don't initialize the language server and start it with defaults.
3. Run the autocompletion on a schema file.

## Environment
- [ yes] Windows
- [ ] Mac
- [ ] Linux
- [ ] other (please specify)

datho7561:
### What does this PR do?
Add null check for `customTags` setting, so if the setting is not set completion will not fail.

Fixes #807

### Is it tested? How?
I added an integration test.

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -293,7 +293,7 @@ export class YamlCompletion {
       proposed,
     };
 
-    if (this.customTags.length > 0) {
+    if (this.customTags && this.customTags.length > 0) {
       this.getCustomTagValueCompletions(collector);
     }
 
--- a/test/autoCompletion.test.ts
+++ b/test/autoCompletion.test.ts
@@ -3126,5 +3126,10 @@ describe('Auto Completion Tests', () => {
         expect(result.items.map((i) => i.label)).to.have.members(['fruit', 'vegetable']);
       });
     });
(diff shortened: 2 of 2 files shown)


gorkem (approved):

msivasubramaniaan (approved):
LGTM

PR #963, opened 2024-05-02
fix: default value with special chars with anyOf

p-spacek:
### What does this PR do?
fixes a small specific issue when schema contains default value with special chars next to anyOf.

This schema
```json
    {
          type: 'object',
          properties: {
            car: {
              type: 'object',
              required: ['engine'],
              properties: {
                engine: {
                  anyOf: [
                    {
                      type: 'object',
                    },
                    {
                      type: 'string',
                    },
                  ],
                  default: 'type$1234',
                },
              },
            },
          }
```
doesn't escape default value

note that an object with anyOf (engine) needs to be inside another object (car), then getInserTextForObject is called with schema { car: {engine: anyOf}}

no ref

### Is it tested? How?
unit-test


src/languageservice/services/yamlCompletion.ts
```diff
@@ -1113,9 +1113,9 @@ export class YamlCompletion {
           case 'number':
           case 'integer':
           case 'anyOf': {
-            let value = propertySchema.default || propertySchema.const;
-            if (value) {
-              if (type === 'string') {
+            let value = propertySchema.default === undefined ? propertySchema.const : propertySchema.default;
```
p-spacek, line 1116: fixes condition to allow null,0,emptyString as a default or const
the same logic is correctly used in other places

src/languageservice/services/yamlCompletion.ts
```diff
@@ -1242,10 +1242,6 @@ export class YamlCompletion {
     return this.getInsertTextForValue(value, separatorAfter, type);
   }
 
-  private getInsertTextForPlainText(text: string): string {
```
p-spacek, line 1245: moved outside of the class to be able to use it from global functions

msivasubramaniaan:
@p-spacek please rebase the repo

PR #966, opened 2024-05-07
fix: not schema error message

p-spacek:
### What does this PR do?
adds the possibility to customize errorMessage for `not` schema
using the same approach `schema.errorMessage || localize(...` as elsewhere

so it's possible to write:
```js
   {
      not: {
        properties: {
          prop1: {
            type: 'number',
          },
        },
        errorMessage: 'prop1 must not be a number',
      }
    }
```

### Is it tested? How?
manually, the same approach is already used elsewhere


msivasubramaniaan:
@p-spacek please rebase the repo

PR #967, opened 2024-05-07
Added language setting option to skip runtime schema compilation

aderende:
## What does this PR do?
It adds `skipSchemaValidation` settings to enable/disable JSON schema validation.

By default, `skipSchemaValidation` is `false`, but if the integrator always calls configure with `skipSchemaValidation = true`, then the code `schema07Validator = ajv.compile(jsonSchema07)` will never get called. This allows the browser to avoid adding the `unsafe-eval` policy directive.

## What issues does this PR fix or reference?
The ajv.compile(...) function requires the `unsafe-eval` policy directive for the `script-src` policy. https://github.com/eclipsesource/jsonforms/issues/1498

The `skipSchemaValidation` setting allows the application to avoid compiling the schema validator and prevents the need to use the `unsafe-eval` policy.

## Is it tested? How?
A test case with `skipSchemaValidation` enabled has been added.

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -5,7 +5,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { JSONSchema, JSONSchemaMap, JSONSchemaRef } from '../jsonSchema';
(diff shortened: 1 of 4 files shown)


msivasubramaniaan:
@aderende please fix the build issue

aderende:
@msivasubramaniaan fixed! Could you, please, take a look?

Issue #964, opened 2024-05-02
Use schema from `"$schema":` for current JSON file


LnLcFlx:
I am using with `nvim-lspconfig` for editing a `JSONC` file like [this](https://github.com/fastfetch-cli/fastfetch/wiki/Configuration#json-based-configuration).
When I add a modeline like `# yaml-language-server: $schema=https://github.com/fastfetch-cli/fastfetch/raw/dev/doc/json_schema.json` I get the desired LSP functionalities.
But when I only use `"$schema": "https://github.com/fastfetch-cli/fastfetch/raw/dev/doc/json_schema.json"` instead of the modeline, like suggested in the link above, it does not work.

What am I doing wrong?

bollwyvl:
Yes, also recently encountered this. 

Having to suggest "only-works-in-this-tool" magic comments feels fairly antithetical to encouraging use of schema conventions.

For reference, other JSON-adjancent tools such as `taplo` support `$schema`.

bollwyvl:
Opened #970.

cromefire:
> For reference, other JSON-adjancent tools such as `taplo` support `$schema`.

Also JetBrains' IDEs would be which are probably one of the other often used IDEs that are used with YAML files.

Only caveat though is that some tools GitLab CI don't like any extra key on their YAML and fail, but #959 would resolve that as well then.

shin19991207:
Closed as completed in https://github.com/redhat-developer/yaml-language-server/pull/970

Issue #437, opened 2021-04-10
enable schema self identification using inline $schema


sooncj:
JSON schema allows a JSON file to self identify its schema using the $schema key. This should also be possible with YAML.

without first saving to a file, $schema should enable validation and auto-completion
```
$schema: https://raw.githubusercontent.com/compose-spec/compose-spec/master/schema/compose-spec.json
version: '3'
services:
  hello:
    image: hello-world
```

Further, this should enable use of local schema files.
For example, using the product catalog example from https://json-schema.org/learn/getting-started-step-by-step.html
If I were to save the final product catalog JSON schema locally to /usr/shared/productCatalog.json
I can then implement a product catalog with validation and auto-completion:
```
$schema: /usr/shared/productCatalog.json
productId: 123
productName: "self identify YAML schema"
}
```

evidolob:
Duplicate https://github.com/redhat-developer/yaml-language-server/pull/354

jeremyfiel:
this is not supported behavior from JSON Schema Specification.  it just so happens that quite a few implementations (json language server) and monaco editor which uses the same under the hood, allow this behavior.

PR #970, opened 2024-05-17
add basic support for declaring schema with inline `$schema`

bollwyvl:
### What does this PR do?

- adds basic support for `$schema`
  - this has _second_ priority
    -  `# yaml-language-server $schema=`
    - this feature
    - customSchemaProvider

- #964

### Is it tested? How?

- `test/fixtures/sample-dollar-schema.json` added
- `test/schema.test.ts`
  - the modeline test is updated to include a `$schema` (which it ignores)
  - a new test is added which _only_ provides a `$schema` (which beats settings)


bollwyvl:
Looking more through older PRs and issues:
- https://github.com/redhat-developer/yaml-language-server/issues/437
- https://github.com/redhat-developer/yaml-language-server/pull/354

it appears some of the relevant argument there was in somehow ignoring `$schema` in `additionalProperties`... this PR doesn't propose anything like that, and the `# yls` modeline or by configuration would indeed be the only way in such a case.

bollwyvl:
I see 1.15.0 went out :tada:... what can I do to help this get into the review pipeline, and hopefully make it into the next release?

bollwyvl:
Updated to latest `main`. I would love to know what changes might be needed: just about need to throw a birthday party for this PR :birthday: !

sergiimk:
Would really like to draw attention to this PR.

Support for raw `$schema` property would be very valuable for tools that use schema ID to differentiate between multiple config types, e.g.:

```yaml
---
$schema: https://example.org/A
a: foo
---
$schema: https://example.org/B
b: bar
```
In this case `$schema` instance property **IS** part of the schema as the tooling needs it to understand the resource type.

The case where target schema does not expect `$schema` and has `additionalProperties: false` is already covered with syntax:
```yaml
# $schema: https://example.org/A
a: foo
```
so I don't think concern about `additionalProperties` that blocked #354 is valid.

Currently we have to write this:
```yaml
# $schema: https://example.org/A
$schema: https://example.org/A
a: foo
```
and it's a big deal for tools that work with user-authored YAML files where such duplication degrades the UX.

datho7561:
I think it's worth having this change. When it was first submitted, yaml-language-server was on the backburner and not receiving much attention, but we've been trying to catch up on maintenance over the past year. This PR needs a rebase before it can be merged.

bollwyvl:
@datho7561 ready for review

datho7561 (approved):
Seems good from my testing. Thanks, Nicholas!

bollwyvl:
Thanks, all.

PR #978, opened 2024-06-07
Fix code scanning issues

remcohaszing:
### What does this PR do?

This resolves 30 occurrences of:

```ts
require(path.join(__dirname), './relative-path'))
```

By replacing them with the equivalent:

```
require('./relative-path')
```

30 code scanning reports on my fork, that I’m sure are also reported on the original repository, though I can’t see them.

### Is it tested? How?

```sh
npm test
```


test/autoCompletion.test.ts
```diff
@@ -786,7 +786,7 @@ describe('Auto Completion Tests', () => {
       });
 
       it('Insert required attributes at correct level', (done) => {
-        const schema = require(path.join(__dirname, './fixtures/testRequiredProperties.json'));
+        const schema = require('./fixtures/testRequiredProperties.json');
```
remcohaszing, line 789: Apparently this trades one code scanning alert for another, though the new one is not as bad as the old one.

I do agree with the new failed checks. The proper solution would be to write this at the top of the file:

```ts
import schema = require('./fixtures/testRequiredProperties.json');
```

PR #981, opened 2024-07-07
Send real errors to telemetry

remcohaszing:
### What does this PR do?

Every call to `telemetry.sendError` contained the same boilerplate code to create an object with the property `error`, which contains a string representation of the real error. This change moves this conversion into the `Telemetry` implementation. This reduces the need for boilerplate code in other places.

This change also affects `monaco-yaml`. Because the telemetry now resolves the real error, the browser can log the error to the console with a better stack trace.

### Is it tested? How?

```sh
npm test
```


remcohaszing:
Bonus tip: if you replace `convertErrorToTelemetryMsg` with Node.js’ builtin function [`util.inspect()`](https://nodejs.org/api/util.html#utilinspectobject-options), you get better error representations, including error causes.

Issue #933, opened 2023-10-27
Remove formatting provider


remcohaszing:
### Is your enhancement related to a problem? Please describe.

`yaml-language-server` supports formatting YAML files. Under the hood this delegates the formatting to Pretttier. However, this integration is very limited. It doesn’t support Prettier options or a Prettier configuration file.

Additionally, the dependency on Prettier is huge. This isn’t great when `yaml-language-server` is used as part of a web bundle.

### Describe the solution you would like

Remove the formatting provider entirely. Maybe recommend Prettier and their official VS Code extension instead in the readme.

### Describe alternatives you have considered

Keep the formatting provider and close this issue.

gorkem:
I am not against the idea. There is a lot of users who are not happy with the prettier based formatter. My concern is the silent crowd who are happy with it. If we can devise a plan that will allow them to replace the function they depend on easily that would be ideal.

znd4:
I really like using prettier for yaml formatting, but it needs to respect `.editorconfig` / `.prettierrc`.

This can be set up with `prettierd` for most usecases, but it would be preferable for `yaml-language-server` to support formatting.

A way to "make everyone happy" would be to break out prettier support as a plugin, so that users can choose which formatting provider `yaml-language-server` invokes (or to choose none and have a minimal `yaml-language-server` installation)

That being said, it seems like a decent chunk of work, and I'm currently disabling `yaml-language-server`'s formatting, so I wouldn't be upset about a decision to remove it completely

remcohaszing:
I also like using Prettier, but it’s a generic formatter that supports a bunch of languages. For them YAML is just one of them. Better integrations already exist.

I think it’s the responsibility of an editor to allow the user to pick a formatter, not that of a language server. I.e. VS Code supports this using the per language `editor.defaultFormatter` setting. I don’t think it’s a good idea to add a plugin system to `yaml-language-server`, because I see the editor as the _plugin system_ and `yaml-language-server` as the _plugin_.

remcohaszing:
I replaced Prettier an alternative formatter in #983. I think that caters to the silent crowd who just want to have some formatting options, but probably don’t care about the implementation used.

Princesseuh:
+1 to this issue, using `yaml-language-server` as a dependency is annoying because the old Prettier version it uses sometimes override other versions of Prettier installed. This is causing issue for us here: https://github.com/withastro/language-tools/issues/627#issuecomment-2307507501

[...]

PR #983, opened 2024-07-24
Use yaml package for formatting

remcohaszing:
### What does this PR do?

This replaces Prettier as the LSP formatting solution with the `yaml` package. The `yaml` package can format YAML just fine, and we already have it as a dependency. Prettier is a big dependency.

If people want to use Prettier, other Prettier integrations are probably better suited for them. For example, the YAML language server doesn’t respect Prettier configuration files.

This is a proof of concept. I tried to map existing options to the new implementation. A more ideal solution might be to change the formatting options.

Refs #933

### Is it tested? How?

```sh
npm test
```

I discovered https://github.com/eemeli/yaml/issues/562 while working on this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,9 +29,6 @@
     "type": "git",
     "url": "https://github.com/redhat-developer/yaml-language-server.git"
   },
-  "optionalDependencies": {
-    "prettier": "2.8.7"
-  },
   "dependencies": {
     "ajv": "^8.11.0",
     "lodash": "4.17.21",
@@ -42,7 +39,7 @@
     "vscode-languageserver-types": "^3.16.0",
     "vscode-nls": "^5.0.0",
     "vscode-uri": "^3.0.2",
-    "yaml": "2.2.2"
+    "yaml": "2.4.5"
   },
   "devDependencies": {
     "@microsoft/eslint-formatter-sarif": "3.0.0",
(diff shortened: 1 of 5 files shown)


datho7561:
I took another look at this; this seems like a feasible alternative to prettier. The main drawback of this approach is that we still don't have direct control of the formatting logic. There's quite a few bugs about weird formatting behaviour and feature requests for new formatting settings. I guess we could always submit PRs upstream to eemeli/yaml, and I'm guessing the maintainer of the parser is more receptive to adding new formatter options than prettier.

There's one test failure after the rebase: "Formatting handles trailing commas (enabled)". Prettier added this option so I added it to the extension a while back, but eemeli/yaml doesn't support this yet. In order to preserve this functionality, we'd need to file a bug upstream, and probably submit the patch to implement it.

datho7561:
> There's one test failure after the rebase: "Formatting handles trailing commas (enabled)". Prettier added this option so I added it to the extension a while back, but eemeli/yaml doesn't support this yet.

My PR to support it got merged: https://github.com/eemeli/yaml/pull/670

datho7561:
@remcohaszing I updated the yaml parser, so it should be possible to implement the trailing comma option now. Do you have time to rebase this PR? If not, I can look into creating a new version of this PR.

remcohaszing:
Sorry, I don’t have the time currently. I would like to get to this eventually, just not right now.

Don’t let me block you. I’m just happy to have proposed the idea. You are more than welcome to work on this in your own branch.

Issue #780, opened 2022-09-19
YAML Language Server checks against latest meta-schema instead of the defined one


ngregory-rbi:
## Expected Behavior

The YAML Language Server respects the `$schema` property and checks against the meta-schema defined by it.

## Current Behavior
The YAML Language Server ignores the `$schema` property and instead checks against the latest meta-schema.

## Steps to Reproduce

1. Open VSCode
2. Create a new file called `openapi.yml`
3. Switch the schema to version 3.0 (based on the added feature for #639)
4. Type anything
5. Observe the error:
```
Schema 'schema.json' is not valid:
/definitions/Schema/properties/multipleOf/exclusiveMinimum : must be number
```

The Open API v3.0 schema uses the `$schema` property like so:
```json
  "$schema": "http://json-schema.org/draft-04/schema#",
```

In Draft 4, `exclusiveMinimum` [MUST be a boolean](https://datatracker.ietf.org/doc/html/draft-fge-json-schema-validation-00#section-5.1.3.1).

However, in the latest draft (currently 2020-12), `exclusiveMinimum` [MUST be a number](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-01#section-6.2.5). It appears this is the meta-schema the language server is checking against, despite the schema specifically specifying Draft 4.

## Environment
- [ ] Windows
- [x] Mac
- [ ] Linux
- [ ] other (please specify)

okybr:
check this out: https://github.com/okybr/yaml-language-server/blob/0c344c441be72b7fffc7b2d98a81ef9113aced76/README.md

related issue:
- https://github.com/redhat-developer/yaml-language-server/issues/752
- https://github.com/OAI/OpenAPI-Specification/issues/2969

KillDozerX2:
Is anybody picking this up?

okybr:
you could just merge this commit in order to fix it: https://github.com/okybr/yaml-language-server/commit/455d452f66c700a3101f87b3970f882c3967e986

KillDozerX2:
Okay, I might have something up my sleeve for this.

amimas:
Any update available on this? I'm looking into creating some schema and it'd be great if I could use the `2020-12` schema.

okybr:
@gorkem, this is not an enhancement, it's a bug!

@KillDozerX2, have you found time to work on this?

KillDozerX2:
> @gorkem, this is not an enhancement, it's a bug!
[...]

Issue #752, opened 2022-07-21
OpenAPI 3.0.0 specification is treated as invalid


cpkio:
## Describe the bug

Schema: `https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/schemas/v3.0/schema.yaml` is treated as invalid.

OpenAPI issue [#2969](https://github.com/OAI/OpenAPI-Specification/issues/2969).

## Expected Behavior

Should be considered valid.

## Current Behavior

Result: `Schema 'schema.yaml' is not valid:  /definitions/Schema/properties/multipleOf/exclusiveMinimum : must be number`

## Steps to Reproduce

1. Set `# yaml-language-server: $schema=https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/schemas/v3.0/schema.yaml` in Neovim modeline in OpenAPI specification file.
2. Set `openapi: 3.0.0` in this OpenAPI specification file.

Schema 3.1.0 works fine.

## Environment
- [x] Windows

okybr:
you can use this patch in order to use draft-04 which is needed for openapi-3.0-schema: https://github.com/okybr/yaml-language-server/commit/455d452f66c700a3101f87b3970f882c3967e986

also check out this readme: https://github.com/okybr/yaml-language-server/blob/0c344c441be72b7fffc7b2d98a81ef9113aced76/README.md

Davidonium:
After encountering this same issue and digging through why this happens, I believe that the issue is that `ajv`, the library used for json schema validation does not support the draft-04 version because there's no security support for it. 
See:
- The line having ajv v8 as a dep https://github.com/redhat-developer/yaml-language-server/blob/main/package.json#L33
- The docs that say there's no support in ajv and prompt the user to use v6 https://ajv.js.org/guide/schema-language.html#draft-04

Davidonium:
I gave a stab supporting this int the PR that was linked. Let me know what you all think!

PR #1006, opened 2024-12-26
feat: add support for draft-04 (2019 and 2020 included) json schemas while supporting draft-07

Davidonium:
### What does this PR do?
Adds support for draft-04, draft-2019-09 and draft-2020-12 json schemas. I first started adding support for draft-04 but some of the changes actually make the code validate the schemas more strictly so I felt the need to add the rest of the drafts. I am not entirely sure of the implications of adding the other drafts which are more complex so I would need some help there testing.

The use of avj-draft-04 is documented here: https://ajv.js.org/json-schema.html#draft-04 so it should not be deprecated or out of support.

According to https://ajv.js.org/api.html#ajv-validateschema-schema-object-boolean using that method should be preferred but I am not sure of the consequences of it. 
Citing to avoid having to click there:
> Validates schema. This method should be used to validate schemas rather than validate due to the inconsistency of uri format in JSON Schema standard.

- https://github.com/redhat-developer/yaml-language-server/issues/780
- https://github.com/redhat-developer/yaml-language-server/issues/752

### Is it tested? How?

I added a unit test but I would like to test it in my IDE. I have no idea how to do that so help is more than welcome.

### DIFF
--- a/test/fixtures/sample-openapiv3.0.0-schema.json
+++ b/test/fixtures/sample-openapiv3.0.0-schema.json
@@ -1,0 +1,1662 @@
+{
+  "id": "https://spec.openapis.org/oas/3.0/schema/2021-09-28",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "The description of OpenAPI v3.0.x documents, as defined by https://spec.openapis.org/oas/v3.0.3",
+  "type": "object",
+  "required": [
+    "openapi",
+    "info",
+    "paths"
+  ],
+  "properties": {
+    "openapi": {
+      "type": "string",
+      "pattern": "^4\\.0\\.\\d(-.+)?$"
+    },
+    "info": {
+      "$ref": "#/definitions/Info"
+    },
+    "externalDocs": {
+      "$ref": "#/definitions/ExternalDocumentation"
+    },
+    "servers": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/Server"
+      }
+    },
+    "security": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/SecurityRequirement"
+      }
+    },
+    "tags": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/Tag"
+      },
+      "uniqueItems": true
+    },
+    "paths": {
+      "$ref": "#/definitions/Paths"
+    },
+    "components": {
+      "$ref": "#/definitions/Components"
+    }
+  },
+  "patternProperties": {
+    "^x-": {
+    }
+  },
+  "additionalProperties": false,
+  "definitions": {
+    "Reference": {
+      "type": "object",
+      "required": [
+        "$ref"
+      ],
+      "patternProperties": {
+        "^\\$ref$": {
+          "type": "string",
+          "format": "uri-reference"
+        }
+      }
+    },
+    "Info": {
+      "type": "object",
+      "required": [
+        "title",
+        "version"
+      ],
+      "properties": {
+        "title": {
+          "type": "string"
+        },
+        "description": {
+          "type": "string"
+        },
+        "termsOfService": {
+          "type": "string",
+          "format": "uri-reference"
+        },
+        "contact": {
+          "$ref": "#/definitions/Contact"
+        },
+        "license": {
+          "$ref": "#/definitions/License"
+        },
+        "version": {
+          "type": "string"
+        }
+      },
+      "patternProperties": {
+        "^x-": {
+        }
+      },
+      "additionalProperties": false
+    },
+    "Contact": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string"
+        },
+        "url": {
+          "type": "string",
+          "format": "uri-reference"
+        },
+        "email": {
+          "type": "string",
+          "format": "email"
+        }
+      },
+      "patternProperties": {
+        "^x-": {
+        }
+      },
+      "additionalProperties": false
+    },
+    "License": {
+      "type": "object",
+      "required": [
+        "name"
+      ],
+      "properties": {
+        "name": {
+          "type": "string"
+        },
+        "url": {
+          "type": "string",
+          "format": "uri-reference"
+        }
+      },
+      "patternProperties": {
+        "^x-": {
+        }
+      },
+      "additionalProperties": false
+    },
+    "Server": {
+      "type": "object",
+      "required": [
+        "url"
+      ],
+      "properties": {
+        "url": {
+          "type": "string"
+        },
+        "description": {
+          "type": "string"
+        },
+        "variables": {
+          "type": "object",
+          "additionalProperties": {
+            "$ref": "#/definitions/ServerVariable"
+          }
+        }
+      },
+      "patternProperties": {
+        "^x-": {
+        }
+      },
+      "additionalProperties": false
+    },
+    "ServerVariable": {
+      "type": "object",
+      "required": [
+        "default"
+      ],
+      "properties": {
+        "enum": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "default": {
+          "type": "string"
+        },
+        "description": {
+          "type": "string"
+        }
+      },
+      "patternProperties": {
+        "^x-": {
+        }
+      },
+      "additionalProperties": false
+    },
+    "Components": {
+      "type": "object",
+      "properties": {
+        "schemas": {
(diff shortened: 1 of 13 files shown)


test/yamlLanguageService.test.ts
```diff
@@ -39,7 +39,7 @@ describe('getLanguageService()', () => {
       // Setup object that resolves schema content
       schemaContentMap[schemaUri] = `
         {
-            "$schema": "https://json-schema.org/draft/2020-12/schema",
+            "$schema": "http://json-schema.org/draft-07/schema",
```
Davidonium, line 42: This is the part that scares me about the change, this now errors out because the $schema field actually gets checked by tha `validateSchema` method.

test/yamlSchemaService.test.ts
```diff
@@ -67,6 +67,50 @@ describe('YAML Schema Service', () => {
...
+      properties:
+        ping:
+          type: string
+          example: pong`;
+
+      const yamlDock = parse(content);
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const openapiV3Schema = require(path.join(__dirname, './fixtures/sample-openapiv3.0.0-schema.json'));
+
+      requestServiceMock = sandbox.fake.resolves(JSON.stringify(openapiV3Schema));
```
Davidonium, line 101: Felt this is a bit redundant but not sure what's the best way to do it. require parses the json but the requestService response must be a string so I stringify it. Later it gets parsed again into an object by the code that loads the schema.

test/yamlSchemaService.test.ts
```diff
@@ -67,6 +67,50 @@ describe('YAML Schema Service', () => {
       expect(schema.schema.type).eqls('array');
     });
 
+    it('should handle schemas that use draft-04', async () => {
+      const content = `openapi: "3.0.0"
```
Davidonium, line 71: I specifically wanted to test an openapi v3.0.0 schema because that's what is not working in my IDE.

test/yamlSchemaService.test.ts
```diff
@@ -67,6 +67,50 @@ describe('YAML Schema Service', () => {
...
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const openapiV3Schema = require(path.join(__dirname, './fixtures/sample-openapiv3.0.0-schema.json'));
+
+      requestServiceMock = sandbox.fake.resolves(JSON.stringify(openapiV3Schema));
+      const service = new SchemaService.YAMLSchemaService(requestServiceMock);
+      service.registerCustomSchemaProvider(() => {
+        return new Promise<string | string[]>((resolve) => {
+          resolve('http://fakeschema.faketld');
+        });
+      });
```
Davidonium, line 107: I came up with this to enforce the retrieval of the schema. Not sure if this is the best, feel free to suggest better ways.

test/fixtures/sample-openapiv3.0.0-schema.json
```diff
@@ -0,0 +1,1662 @@
+{
```
Davidonium, line 1: Maybe having the entire openapi v3.0.0 spec is a bit too much. Let me know if you want me to just use a simple one that uses draft-04.
I used this because it's the use case I wanted to cover (and I think supporting openapi v3.0.0 is very important as there are tools that are unable to upgrade to 3.1).

test/yamlLanguageService.test.ts
Davidonium: I went ahead and added support for the rest of the drafts from this and left the test as it is.

test/yamlSchemaService.test.ts
Davidonium: I fixed this by loading the file using `readFile` instead of require.

HampusHauffman:
Bump 🥇

Davidonium:
Is there anything I can do to help you all review this one?

msivasubramaniaan (changes requested):
Please fix lint issues

Davidonium:
The issue can't be resolved because the bot lint [github-advanced-security](https://github.com/apps/github-advanced-security) 
is ignoring the rule ignore just above. Running `yarn lint-ci` is fine locally but that linter is misconfigured.
I tried using es imports and addin json resolution but the `umd` build then fails because it's not supported.

What can be done to resolve that?

Davidonium:
I removed support for draft-06 just to avoid the import. I guess if anyone needs support for that will need to deal with the linting issue.

msivasubramaniaan (approved):
LGTM

Davidonium:
Thanks for the review!

lantw44:
It looks like it is broken again in 1.17.0.

Davidonium:
This PR was reverted because some users complained about the validation failing when the `$schema` field starts with `https`.
Validation failing is techincally correct, as `https://json-schema.org/draft-07/schema` is a correct URL but not a correct schema identifier. The correct schema identifier for `draft-07` is `http://json-schema.org/draft-07/schema`.

Since this seems to be a source of confusion to users, I guess that a solution could be reached by either not validating the schemas or providing a better error message to convey the above.

Reference on the schema identifier actually being right in this PR:
- https://github.com/ajv-validator/ajv/issues/1104
- https://github.com/ajv-validator/ajv/issues/472
- https://stackoverflow.com/questions/69133771/ajv-no-schema-with-key-or-ref-https-json-schema-org-draft-07-schema

Issues that cause the PR to be reverted:
- https://github.com/redhat-developer/vscode-yaml/issues/1100
- https://github.com/redhat-developer/vscode-yaml/issues/1096
- https://github.com/redhat-developer/yaml-language-server/issues/1023

Relequestual:
> This PR was reverted because some users complained about the validation failing when the `$schema` field starts with `https`. Validation failing is techincally correct, as `https://json-schema.org/draft-07/schema` is a correct URL but not a correct schema identifier. The correct schema identifier for `draft-07` is `http://json-schema.org/draft-07/schema`.

This is the correct behaviour. Can you help me understand the decision to revert this PR please?

Relequestual:
You know, newer modern tooling which does validation using JSON Schema, often will throw an error in such a situation, and not try to download any schems based on the `$schema` value. Unless they are the meta-schemas, they usually have to be pre-loaded.

karenetheridge:
The canonical metaschema URI is actually `http://json-schema.org/draft-07/schema#` (with the `#`).

Davidonium:
> This is the correct behaviour. Can you help me understand the decision to revert this PR please?

From what I understood, it was reverted from some issues raised in the repo.

The revert happened here:
- https://github.com/redhat-developer/yaml-language-server/pull/1026

And it was attempted here:
- https://github.com/redhat-developer/yaml-language-server/pull/1025

If you follow through both PRs, these are some of the raised issues:
- https://github.com/redhat-developer/vscode-yaml/issues/1100
- https://github.com/redhat-developer/yaml-language-server/issues/1023
- https://github.com/redhat-developer/vscode-yaml/issues/1096

Just for clarification, I didn't take part in the revert, just explaining what I saw from the outside. 

From what I understand, this is just lack of knowledge (which is understandable) on how the actual schema ids and references work and that there's a difference between where they can be downloaded and how they are referenced.

> The canonical metaschema URI is actually `http://json-schema.org/draft-07/schema#` (with the `#`).
That's right, I fixed my previous comment.

damioune123:
Hi, this PR seemed to support  for draft 04, 2019 and 2020 but was eventually revert due to issue with draft 04 (if I understand correctly). If there were no particular issues with 2019 and 2020, cant we add them back ? 
It seems like a lot of users would love to have support for draft 07 > features.

If it's just about adding back validators for 2019 and 2020, I dont mind creating the PR myself, but I d like to know first what was the main issue with them (if any).

datho7561:
I wanted to focus on getting draft-04 working, because that's what a lot of openapi documents were using (we had many bugs filed specifically about openapi documents).

The process for supporting 2019 and 2020 should be similar to what I did for draft-04; I think that ajv has validators for them. The implementation used in the older PR that got reverted was causing problems.

If you're interested in implementing the support, I can review it.

jeremyfiel:
ajv doesn't fully support draft-2019 or draft-2020-12 because they don't support $dynamic* keywords.  Using ajv to validate OpenAPI definitions is very brittle because OpenAPI schemas are a superset and subset of draft-04. Not all draft-04 features are supported and some keywords have slight behavioral differences.  https://github.com/ajv-validator/ajv/issues/1573

there's another implementation of all JSON Schema drafts, including OpenAPI versions at www.github.com/hyperjump-io/json-schema

PR #1022, opened 2025-02-13
upstream es lint target

msivasubramaniaan:
### What does this PR do?
upstream [ES lint target](https://www.typescriptlang.org/tsconfig/#target) 

NA

### Is it tested? How?
NA


datho7561:
`@types/node` version 22.14.0 isn't out yet, I think you'll need to specify `22.13.2` and rerun `yarn` to update the lockfile (yarn.lock)

PR #1024, opened 2025-02-19
Import Prettier from prettier/standalone

remcohaszing:
### What does this PR do?

The import from `prettier` includes support for all languages supported by Prettier. The `prettier/standalone` import only includes the core. The `prettier` export also points to `prettier/standalone` for the `browser` export condition.

This means that with this change `yaml-language-server` has to load less code, leading to faster startup time. And depending on configuration, it may lead to smaller bundle sizes.

This PR is an improvement, but I would strongly prefer #983.

### Is it tested? How?

`npm test`


msivasubramaniaan (approved):
LGTM

Issue #997, opened 2024-11-11
Build image for ARM64


Stef16Robbe:
### Is your enhancement related to a problem? Please describe.
I tried running the docker image on my MacOS (`15.1 24B83 arm64`) but I got this error that an ARM manifest is not available

```
$ docker pull quay.io/redhat-developer/yaml-language-server
Using default tag: latest
latest: Pulling from redhat-developer/yaml-language-server
no matching manifest for linux/arm64/v8 in the manifest list entries
```

### Describe the solution you would like
I want the Containerfile to be built for the project and uploaded in the registry

### Describe alternatives you have considered
I am now building it myself for arm64

PR #1027, opened 2025-03-03
supported docker arm64 image

msivasubramaniaan:
### What does this PR do?
Support linux/arm64 docker image

https://github.com/redhat-developer/yaml-language-server/issues/997

### Is it tested? How?
NA


datho7561 (approved):
From what I can tell, this should work. On my computer locally, docker lets me build contains for arm64 even though my computer is amd64.

Issue #1036, opened 2025-03-12
Enum strings `YES` / `NO` are converted to boolean incorrectly


jeremyfiel:
## Describe the bug

In the following schema, the enum has two *string* values, `YES` and `NO`.  These values are converted into a boolean because these two values are equivalent replacements in prior versions of YAML. I want to use them as string values in my data instance. 

```yaml
$schema: http://json-schema.org/draft-04/schema#
type: object
additionalProperties: false
required:
  - test
properties:
  test:
    type: string
    enum:
      - 'YES'
      - 'NO'
      - JEREMY
```

```yaml
# yaml-language-server: $schema=test.schema.yaml
test: YES
```

![Image](https://github.com/user-attachments/assets/846ea7f7-8447-44a6-a4e2-1b283d89baa3)

![Image](https://github.com/user-attachments/assets/d1e212a9-d80a-44aa-b7be-ebf6c4179226)

## Expected Behavior

The string values should be preserved and not converted to boolean per the [1.2 Specification.](https://yaml.org/spec/1.2.0/)

> The primary objective of this revision is to bring YAML into compliance with JSON as an official subset. YAML 1.2 is compatible with 1.1 for most practical applications - this is a minor revision. An expected source of incompatibility with prior versions of YAML, especially the syck implementation, is the change in implicit typing rules. We have removed unique implicit typing rules and have updated these rules to align them with JSON's productions. In this version of YAML, boolean values may be serialized as “true” or “false”;

## Current Behavior

The string values are converted to boolean, incorrectly.

## Steps to Reproduce

1. create the schema
2. create the test file and add the modeline to assign the schema
3.  use auto-suggestion to display enum values
4. select `YES` or `NO`
5. observe value is converted to boolean

## Environment
- [x] Windows
- [ ] Mac
- [ ] Linux
- [ ] other (please specify)

kevinkinyuasonos:
Hi has there been any news on when this would be released to the YAML redhat extension in vscode?

PR #1038, opened 2025-03-12
Fix ENUM yes no values

msivasubramaniaan:
### What does this PR do?
This PR suggests ENUM values based on type. Especially on the YAML spec 1.1 Boolean values like `on,off,true,false,yes,no`

https://github.com/redhat-developer/yaml-language-server/issues/1036

### Is it tested? How?
Yes. Added test cases

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -1674,7 +1674,7 @@ function convertToStringValue(param: unknown): string {
   let value: string;
   if (typeof param === 'string') {
     //support YAML spec 1.1 boolean values
-    value = ['on', 'off', 'true', 'false', 'yes', 'no'].includes(param.toLowerCase()) ? `'${param}'` : param;
+    value = ['on', 'off', 'true', 'false', 'yes', 'no'].includes(param.toLowerCase()) ? `"${param}"` : param;
   } else {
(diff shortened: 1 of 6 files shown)


src/languageservice/services/yamlCompletion.ts
```diff
@@ -1673,7 +1673,8 @@ const isNumberExp = /^\d+$/;
 function convertToStringValue(param: unknown): string {
   let value: string;
   if (typeof param === 'string') {
-    value = param;
+    //support YAML spec 1.1 boolean values
+    value = ['on', 'off', 'true', 'false', 'yes', 'no'].includes(param.toLowerCase()) ? `"${param}"` : param;
```
datho7561, line 1677: It seems there is a configuration `yaml.format.singleQuote` that dictates if `"` or `'`. I think that should be respected.

datho7561 (changes requested):
Something that might be worth addressing, but otherwise works well from my testing.

datho7561 (approved):
Works great! Thanks, Muthu!

jeremyfiel:
Thanks for a quick turn around on this.

kevinkinyuasonos:
I'd like to follow up on this PR and see when the update will be available for the YAML extension in VS Code?

PR #1040, opened 2025-03-20
fix for auto completion with escape sequence chars

msivasubramaniaan:
### What does this PR do?
This PR suggest characters with escape chars if needed

https://github.com/redhat-developer/vscode-yaml/issues/1097

### Is it tested? How?
Yes added test case for the fix

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -243,12 +243,29 @@ export class YamlCompletion {
           }
         }
 
-        // trim $1 from end of completion
-        if (completionItem.insertText.endsWith('$1') && !isForParentCompletion) {
-          completionItem.insertText = completionItem.insertText.substr(0, completionItem.insertText.length - 2);
-        }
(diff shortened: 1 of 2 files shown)


datho7561:
This doesn't fix issue for me. Make the following document, and open completion at the `|`:

```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/RedCMD/TmLanguage-Syntax-Highlighter/main/vscode.tmLanguage.schema.json
|
```

Accept the completion "New Grammar".

The generated content is still broken.

datho7561:
The bug report was about the completing the content using the "New Grammar" completion. In order to address the bug, the "New Grammar" completion should work as expected, with the quotes properly escaped when necessary.

msivasubramaniaan:
Yes but we can do the workaround only. Complete fix not be possible. I've tried it out fully.

datho7561 (approved):
Okay. We should keep the original issue open since this doesn't fix it, but I agree that this is worth merging because it addresses the issue you wrote a test case for.

PR #1051, opened 2025-04-22
fix: exclude not suggest properties from possible properties error

p-spacek:
### What does this PR do?
suggested properties are filtered out if the schema contains doNotSuggest or  deprecationMessage
this PR does the same for `possibleProperties` in validation error

no ref

### Is it tested? How?
adds unit test


msivasubramaniaan (approved):
LGTM

PR #1059, opened 2025-04-29
showed error on unresolved alias

msivasubramaniaan:
fixes [#1098](https://github.com/redhat-developer/vscode-yaml/issues/1098)

### What does this PR do?
This PR showed diagnostic error when there were no alias
image

https://github.com/redhat-developer/vscode-yaml/issues/1098

### Is it tested? How?
Yes added test case for the fix

### DIFF
--- a/src/languageservice/services/validation/unused-anchors.ts
+++ b/src/languageservice/services/validation/unused-anchors.ts
@@ -44,7 +44,12 @@ export class UnusedAnchorsValidator implements AdditionalValidator {
             document.positionAt(aToken.offset),
             document.positionAt(aToken.offset + aToken.source.length)
           );
-          const warningDiagnostic = Diagnostic.create(range, `Unused anchor "${aToken.source}"`, DiagnosticSeverity.Hint, 0);
+          const warningDiagnostic = Diagnostic.create(
+            range,
+            `Unused anchor "${aToken.source}"`,
+            DiagnosticSeverity.Information,
+            0
+          );
           warningDiagnostic.tags = [DiagnosticTag.Unnecessary];
           result.push(warningDiagnostic);
         }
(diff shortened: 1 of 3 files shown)


src/languageservice/services/validation/unused-anchors.ts
```diff
@@ -46,6 +51,17 @@ export class UnusedAnchorsValidator implements AdditionalValidator {
       }
     }
 
+    unIdentifiedAlias.forEach((node) => {
+      const nodeRange = node.range;
+      if (nodeRange) {
+        const startOffset = nodeRange[0];
+        const endOffset = nodeRange[1];
+        const range = Range.create(document.positionAt(startOffset), document.positionAt(endOffset));
+        const warningDiagnostic = Diagnostic.create(range, `Unresolved alias "${node}"`, DiagnosticSeverity.Hint, 0);
```
datho7561, line 60: Maybe it's worth setting a higher severity, such as `Info`? There's no underlining for `Hint` in VS Code, so it's hard to see.

datho7561 (changes requested):

src/languageservice/services/validation/unused-anchors.ts
datho7561, line 60: Also maybe worth adding "it will be treated as a string instead" into the error message

msivasubramaniaan:
image

datho7561:
Sorry, somehow this slipped through for me, I'll take another look

src/languageservice/services/validation/unused-anchors.ts
```diff
@@ -39,13 +44,29 @@ export class UnusedAnchorsValidator implements AdditionalValidator {
...
       }
     }
 
+    unIdentifiedAlias.forEach((node) => {
+      const nodeRange = node.range;
+      if (nodeRange) {
+        const startOffset = nodeRange[0];
+        const endOffset = nodeRange[1];
+        const range = Range.create(document.positionAt(startOffset), document.positionAt(endOffset));
+        const warningDiagnostic = Diagnostic.create(range, `Unresolved alias "${node}"`, DiagnosticSeverity.Information, 0);
```
datho7561, line 65: Do you think it's worth adding "it will be treated as a string instead" into the error message?

datho7561 (approved):
One more comment that I think you missed from last time, but it's fine to merge as-is.

src/languageservice/services/validation/unused-anchors.ts
msivasubramaniaan: No basically yaml treated as string. I feel no more additional messages

PR #1061, opened 2025-04-30
Fix corner case for 'flow sequence forbidden' quickfix

### LINKED ISSUE #1060: QuickFix for "flow sequence forbidden" leaves in trailing `]`
## Describe the bug

Given the following YAML:

```yaml
food: ['apple','bread','croutons','dandriff','escargot','fettucine','gastronomy handbook']
```

And the following in in `settings.json`:

```jsonc
{
// ...
  "yaml.style.flowSequence": "forbid"
// ...
}
```

An error is generated, because flow sequences aren't allowed. If you attempt to apply the quickfix code action to fix the problem, it results in invalid YAML, because the trailing `]` is left in.

Note that this only happens when the flow sequence is at the very end of the document, which messes up the error range as well.

## Expected Behavior
```yaml
food:
  - 'apple'
  - 'bread'
  - 'croutons'
  - 'dandriff'
  - 'escargot'
  - 'fettucine'
  - 'gastronomy handbook'
```

## Current Behavior
```yaml
food:
  - 'apple'
  - 'bread'
  - 'croutons'
  - 'dandriff'
  - 'escargot'
  - 'fettucine'
  - 'gastronomy handbook']
```

## Steps to Reproduce
1. Make the above
2. Use settings.json to forbid flow sequences
3. Apply the code action to turn the flow sequence into a normal YAML list

## Environment
- [ ] Windows
- [ ] Mac
- [x] Linux
- [ ] other (please specify)

datho7561:
### What does this PR do?
- Do not include any trailing spaces in the error range
  - This is a change in behaviour; it seems this behaviour was intentional based on the test cases, but I don't agree with it, since the error range should only cover the erroneous code.
- Fix a bug where `]` or `}` are not included in the error range if they're the last character in the document

Fixes #1060

### Is it tested? How?
Updated the unit tests to reflect that trailing whitespaces are not included in the range. The corner case of a flow sequence at the end of the document was already covered in the suite, but the expected range of the diagnostic was wrong.

### DIFF
--- a/src/languageservice/services/validation/yaml-style.ts
+++ b/src/languageservice/services/validation/yaml-style.ts
@@ -42,6 +42,9 @@ export class YAMLStyleValidator implements AdditionalValidator {
   }
 
   private getRangeOf(document: TextDocument, node: FlowCollection): Range {
-    return Range.create(document.positionAt(node.start.offset), document.positionAt(node.end.pop().offset));
+    const endOffset = node.end[0].offset;
+    let endPosition = document.positionAt(endOffset);
+    endPosition = { character: endPosition.character + 1, line: endPosition.line };
+    return Range.create(document.positionAt(node.start.offset), endPosition);
   }
 }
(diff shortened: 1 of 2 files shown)


msivasubramaniaan (approved):
LGTM

PR #1062, opened 2025-05-01
Support statically registering format support

### LINKED ISSUE #1031: How to format?
:lua vim.lsp.buf.format() says `[LSP] Format request failed, no matching language servers.` but the lsp is installed and working. Removing a colon shows an error `Implicit keys need to be on a single line` so it is working. Why doesn't format work? It works for other lsps I have.

datho7561:
### What does this PR do?
Some clients, like Neovim, do not support dynamic registration of formatting and range formatting. In this case, to tell the client that formatting is supported, the response to `initialized`, which contains the `ServerCapabilities`, should mark formatting and range formatting as supported.

Fixes #1031

### Is it tested? How?
Tried it out in Neovim, it's working after this patch.

### DIFF
--- a/src/yamlServerInit.ts
+++ b/src/yamlServerInit.ts
@@ -78,8 +78,8 @@ export class YAMLServerInit {
     );
     this.yamlSettings.clientDynamicRegisterSupport = !!(
       this.yamlSettings.capabilities.textDocument &&
-      this.yamlSettings.capabilities.textDocument.rangeFormatting &&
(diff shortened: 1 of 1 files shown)


datho7561:
In this PR, I assume that iff (if and only if) the server supports dynamically registering range formatting it supports dynamically registering regular formatting. I think this is a safe assumption and I think that we're already making it for the dynamic code.

msivasubramaniaan (approved):
LGTM

PR #1063, opened 2025-05-05
Add option to remove trailing comma

datho7561:
### What does this PR do?
Adds the 'yaml.format.trailingComma' to the formatter.

This option is available in prettier so implementing it is trivial. Needs the 'estree' prettier plugin, which is responsible for formatting JSON. As a result, this might increase the size of the language server.

eg. in the following, when 'yaml.format.trailingComma` is set to `false`, the last comma should be removed:

```yaml
{
  key: 'value',
  food: 'raisins',
  airport: 'YYZ',
  lightened_bulb: 'illuminating',
}
```

See redhat-developer/vscode-yaml#1112

### Is it tested? How?
Unit tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "dependencies": {
     "ajv": "^8.11.0",
     "lodash": "4.17.21",
-    "prettier": "^3.0.0",
+    "prettier": "^3.5.0",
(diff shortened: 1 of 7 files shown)


msivasubramaniaan (approved):
LGTM

PR #1066, opened 2025-05-06
checked const value for the Boolean type

msivasubramaniaan:
### What does this PR do?
This PR do the validation for the const value of Boolean type

https://github.com/redhat-developer/vscode-yaml/issues/1116

### Is it tested? How?
Yes with existing test case

### DIFF
--- a/test/schemaValidation.test.ts
+++ b/test/schemaValidation.test.ts
@@ -2105,9 +2105,14 @@ obj:
         },
       },
     });
-    const content = `prop: false`;
-    const result = await parseSetup(content);
+
+    let content = `prop: false`;
+    let result = await parseSetup(content);
     expect(result.length).to.eq(1);
     expect(result[0].message).to.eq('Value must be true.');
+
+    content = `prop: true`;
+    result = await parseSetup(content);
+    expect(result.length).to.eq(0);
   });
 });
--- a/test/schemaValidation.test.ts
+++ b/test/schemaValidation.test.ts
@@ -2094,4 +2094,20 @@ obj:
     const result = await parseSetup(content);
     assert.equal(result.length, 0);
   });
+
+  it('value should match as per schema const on boolean', async () => {
+    schemaProvider.addSchema(SCHEMA_ID, {
+      type: 'object',
+      properties: {
+        prop: {
+          const: true,
+          type: 'boolean',
+        },
+      },
+    });
+    const content = `prop: false`;
+    const result = await parseSetup(content);
+    expect(result.length).to.eq(1);
+    expect(result[0].message).to.eq('Value must be true.');
+  });
 });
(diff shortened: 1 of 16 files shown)


datho7561 (changes requested):
The quickfix is still broken ( there is a quickfix entry with no label that does nothing when the value is wrong ).

eg. if you have

```yaml
prop: false
```
with the existing schema it will not suggest changing it to `true`

I couldn't find an existing test case that has a schema with `"const": true, "type": "boolean"`. It would be nice to have one so that we can detect any regressions.

datho7561:
For the quickfix, something like this could solve it. We should add a test case too, though.

```diff
diff --git a/src/languageservice/services/yamlCodeActions.ts b/src/languageservice/services/yamlCodeActions.ts
index da0c7ee..85eba90 100644
--- a/src/languageservice/services/yamlCodeActions.ts
+++ b/src/languageservice/services/yamlCodeActions.ts
@@ -348,10 +348,11 @@ export class YamlCodeActions {
         continue;
       }
       for (const value of values) {
+        const cleanedValue = typeof value === 'boolean' ? `${value}` : value;
         results.push(
           CodeAction.create(
-            value,
-            createWorkspaceEdit(document.uri, [TextEdit.replace(diagnostic.range, value)]),
+            cleanedValue,
+            createWorkspaceEdit(document.uri, [TextEdit.replace(diagnostic.range, cleanedValue)]),
             CodeActionKind.QuickFix
           )
         );
```

msivasubramaniaan:
> The quickfix is still broken ( there is a quickfix entry with no label that does nothing when the value is wrong ).
> 
> eg. if you have
> 
> ```yaml
> prop: false
> ```
> 
> with the existing schema it will not suggest changing it to `true`
> 
> I couldn't find an existing test case that has a schema with `"const": true, "type": "boolean"`. It would be nice to have one so that we can detect any regressions.

With the recent fix it will ask to change it to `false` when the schema expects `true`. Please find the attachment and I will add a sperate test case
image

test/schemaValidation.test.ts
```diff
@@ -2094,4 +2094,20 @@ obj:
...
+    schemaProvider.addSchema(SCHEMA_ID, {
+      type: 'object',
+      properties: {
+        prop: {
+          const: true,
+          type: 'boolean',
+        },
+      },
+    });
+    const content = `prop: false`;
```
datho7561, line 2108: The failing case was that `prop: true` was flagged as incorrect. It would be nice to test both `prop: true` and `prop: false` in this test case.

datho7561 (changes requested):

datho7561 (approved):
This fixes the validation issue. The quickfix can be fixed later. Thanks, Muthu!

PR #1072, opened 2025-05-12
Update to latest vscode-json-languageservice

### LINKED ISSUE #1069: Update vscode-json-languageservice to latest version
There are improvements that we could benefit from, such as allowing warnings to be reported on schemas instead of just errors.

Update vscode-json-languageservice to the latest version, 5.5.0 as of the time of writing

- Keep it as is and add the desired features in other ways
- vendor some/all of the package into vscode-yaml, which would allow

### Additional context
It's not trivial, since it's a major version update, and we reference internal code, which has changed significantly

datho7561:
### What does this PR do?
This PR updates vscode-json-languageservice to the latest version.

The primary motivation for doing this is to allow warnings to be reported on schemas. I would like to report warnings on schemas as a part of https://github.com/redhat-developer/yaml-language-server/pull/1065.

When updating the json language service, there were many API changes that I needed to adapt to, and some tests that I changed.

Notably, I needed to copy a lot of the implementation of `JSONSchemaService` into our subclass of `JSONSchemaService`. `JSONSchemaService` turns all schema identifiers into URIs and escapes all `/` in the fragment, and does this in the private `normalizeId` method that we can't override Combined, these behaviours cause many tests to fail.

Closes #1069

### Is it tested? How?
Existing test suite.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "lodash": "4.17.21",
     "prettier": "^3.5.0",
     "request-light": "^0.5.7",
-    "vscode-json-languageservice": "4.1.8",
+    "vscode-json-languageservice": "5.5.0",
     "vscode-languageserver": "^9.0.0",
     "vscode-languageserver-textdocument": "^1.0.1",
     "vscode-languageserver-types": "^3.16.0",
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -12,7 +12,6 @@ import {
   JSONSchemaService,
   SchemaDependencies,
   ISchemaContributions,
-  SchemaHandle,
 } from 'vscode-json-languageservice/lib/umd/services/jsonSchemaService';
 
 import { URI } from 'vscode-uri';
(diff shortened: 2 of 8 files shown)


msivasubramaniaan (approved):

datho7561:
~200 tests are failing when I rebase this PR. This is broken; I need to take another proper look at this.

brian-mulier-p:
Hello! We need this bump for our project, is it something any idea on when you're going to have another look at it? I already tried a bump PR on my end before seeing yours but I didn't dig enough into the code to fix the remaining bits for it to work properly..

datho7561:
I'll do a quick rebase on this PR; there's a lot broken though. We overwrite vscode-json-languageservice internals, and those internals have changed. It's been pretty painful to figure out why certain cases are failing.

I'm starting to think that copying the code we need over might make more sense than trying to rewrite the functionality we need by subclassing an internal class. I don't think we're saving any effort by reusing vscode-json-languageservice, since we need to modify the code extensively and using vscode-json-languageservice makes the code inflexible.

Issue #922, opened 2023-09-10
Numbers in comments


alexvoss:
## Describe the bug

I am using yaml-language-server via RedHat's [VSCode plugin](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) and whenever I put a number (like '7') into a comment that occurs 
at the beginning of the document, the schema does not get loaded:

<img width="1145" alt="image" src="https://github.com/redhat-developer/yaml-language-server/assets/4134224/8abf9702-c34d-4fd2-8a19-d2028ee2217d">

The error also occurs when I have the comment above the document start line ('`---`') or when I add a `%YAML` directive. See [this discussion](https://github.com/redhat-developer/vscode-yaml/issues/859) of the behavior in the VSCode plugin. 

This is not an issue with the plugin since I can reproduce the issue with vim:

<img width="818" alt="image" src="https://github.com/redhat-developer/yaml-language-server/assets/4134224/dd2d67f9-6374-465b-9359-55849adb0478">

## Expected Behavior

The schema should be loaded and used to validate the YAML file.

## Current Behavior

Instead, I get the error message `Unable to parse content from '[absolute path]/testSchema.yaml': Parse error at offset 0.`

## Steps to Reproduce

1. Create this schema file:
```
---
# a comment - that seems to work unless we add a number 123
title: Learning JSON/YAML schema 
type: object
properties:
  id: 
    type: string
```
2. Create this YAML file in the same directory:
```
# yaml-language-server: $schema=./testSchema.yaml
---
id: "123"
description: "a short description"
```

## Environment
- [ ] Windows
- [X] Mac (and vim or VSCode)
- [ ] Linux
- [ ] other (please specify)

alexvoss:
One more observation: the problem disappears when I add a modeline. Hope this helps.

# yaml-language-server: $schema=http://json-schema.org/draft-07/schema

PR #1073, opened 2025-05-14
Fix YAML JSON Schema files starting with %YAML 1.x

dlech:
### What does this PR do?

Fix a bug where YAML files containing a JSON Schema that start with %YAML 1.x or a comment that contains a number would fail to to be parsed.

This is caused by the fact that the JSON Schema parser would contain the number, e.g. 1.2 as the value of `unresolvedJsonSchema.schema` and so the test for `=== undefined` would fail. This would cause the error from the JSON parser to be returned instead of trying to parse the file as a YAML file.

To work around this, also check if `unresolvedJsonSchema.schema` is a number. It if is, it clearly was not a JSON Schema and so we can safely try to parse the file as a YAML file.

Fixes: https://github.com/redhat-developer/yaml-language-server/issues/922

### Is it tested? How?

The modified test failed before the code change and passes after.


datho7561 (approved):
This change fixes the bug for me. Thanks!

Issue #982, opened 2024-07-22
Make "Show all Enum on Hover" Optional


prose-leviathan:
## Describe the bug
I use 

https://github.com/redhat-developer/yaml-language-server/pull/942 introduced a new default behaviour which makes large lists of enum, especially when using enumMarkdown, not usable with the hover. I already reported on the PR, opening this issue now to try having eyes on it.

## Expected Behavior

Do not show the full list if it's too big, and just show the correct description for the value. At least provide the option to enable or disable this beheaviour.

## Current Behavior
What happens is that the entire enum description is shown, with the description for the value being all the way to the bottom. When the list of enum is really big, this means a user won't see the description for the value they are hovering and have to scroll all the way, which they won't

## Steps to Reproduce
Create a json schema with enum and enumMarkdown entries. Make it a large enum (more than 30 items), and see on hover of one option how the whole list appears, and does not display the description for the value.

## Environment
- [ ] Windows
- [ ] Mac
- [ ] Linux
- [ ] other (please specify)

PR #1074, opened 2025-05-15
show the matching enum value on top

msivasubramaniaan:
### What does this PR do?
This PR shows the matching enum value on top of the list. As user wants an one more option to show/hidden the entire ENUM in https://github.com/redhat-developer/yaml-language-server/issues/982 but I don't feel comfort to add an another optional parameter. Instead of showing the matching value on top as readability.

Please refer the screenshot
image

https://github.com/redhat-developer/yaml-language-server/issues/982

### Is it tested? How?
Yes with existing test case

### DIFF
--- a/src/languageservice/jsonSchema.ts
+++ b/src/languageservice/jsonSchema.ts
@@ -49,8 +49,6 @@ export interface JSONSchema {
   not?: JSONSchemaRef;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   enum?: any[];
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  enumMarkdown?: any[];
   format?: string;
 
   // schema draft 06
--- a/src/languageservice/jsonSchema.ts
+++ b/src/languageservice/jsonSchema.ts
@@ -49,6 +49,7 @@ export interface JSONSchema {
   not?: JSONSchemaRef;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   enum?: any[];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   enumMarkdown?: any[];
   format?: string;
 
--- a/src/languageservice/jsonSchema.ts
+++ b/src/languageservice/jsonSchema.ts
@@ -49,6 +49,7 @@ export interface JSONSchema {
   not?: JSONSchemaRef;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   enum?: any[];
+  enumMarkdown?: any[];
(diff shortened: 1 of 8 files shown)


src/languageservice/jsonSchema.ts
```diff
@@ -49,6 +49,8 @@ export interface JSONSchema {
   not?: JSONSchemaRef;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   enum?: any[];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  enumMarkdown?: any[];
```
datho7561, line 53: Is this a field that existing JSON schemas actually use? I don't think it is? I think the user meant `enumDescriptions` and `markdownEnumDescriptions`, which VS Code uses. If you don't think it's worth adding this as a new field, ignore the below text.

---

If this is a field that schemas use, we'd need to also accommodate this by showing the documentation in the completion documentation when completing one of the enums.

eg. you connect this YAML document to this schema and open completion where the `|` is:

YAML document:
```yaml
programmingLanguage: Sca|
```

Schema:
```json
{
  "type": "object",
  "properties": {
    "programmingLanguage": {
      "$ref": "#/definitions/programming-language"
    }
  },
  "definitions": {
    "programming-language": {
      "type": "string",
      "enum": ["Scala"],
      "markdownDescription": "The language that the user is programming in",
      "enumMarkdown": [
        "I _don't_ know enough about Scala to quip about it"
      ]
    }
  }
}
```

It should show the documentation " I _don't_ know enough..." on the right of the enum value, because this is what it does if you change `enumMarkdown` to `enumDescriptions` or `markdownEnumDescriptions`

datho7561 (changes requested):

datho7561:
I think the behaviour to show the description for the enum at the top of the list is a good way to address the user's issue, and it's working well. I only have doubts about the new `enumMarkdown` you've introduced.

msivasubramaniaan:
As user report the bug `onhover` not in a `completion`. Also in the step to reproduce user mentioned that `enumMarkdown` params. I feel it good to an one more param

msivasubramaniaan:
> > in the step to reproduce user mentioned that enumMarkdown params
> 
> Supporting the `enumMarkdown` field is a new feature. When I Googled for this field, the only place it was mentioned was the issue that the user filed. I think the user meant `markdownEnumDescriptions`.
> 
> If you do want to add support for `enumMarkdown`, you should do it properly. It should work for both hover documentation and completion documentation, like it does for `markdownEnumDescriptions`.

I just removed `enumMarkdown` param

Issue #1075, opened 2025-05-15
yaml conversion does not resolve aliases correctly.


pjsk-stripe:
## Describe the bug
Aliases are not resolved after a certain, seemingly arbitrary, point in a large yaml file. This does not appear to be a function of depth of aliases but rather of number of aliases in the file. 

## Expected Behavior
`redhat.vscode-yaml` VSCode extension can process aliases in a large yaml file equally across the entire file. 

## Current Behavior
Tested using `redhad.vscode-yaml` VSCode extension. A large file with N independent aliases will only resolve the properties with the first N entries. 

## Steps to Reproduce
This is a bit complicated, and sorry that it depends on the vscode extension.  

1. Install `redhad.vscode-yaml` in vscode. 
2. Create a large yaml file with a lot of objects including a known field  ([here's the one I used](https://gist.github.com/pjsk-stripe/5aaeba18dce1dc2af590d11153b9a085))
3. Add a schema that ensures the field is required to `yaml.schemas` (in the above case, "name" is inherited from `foo` on all 2000 objects).
  ```json
  {
    "id": "http://stripe.com/foo-yaml-schema#",
    "$schema": "http://json-schema.org/draft-04/schema#",
    "description": "Schema for a foo.yaml",
    "type": "object",
    "additionalProperties": {
      "$ref": "#/definitions/foo"
    },
    "definitions": {
      "foo": {
        "type": "object",
        "required": [":name"],
        "additionalProperties": false,
        "properties": {
          ":name": { "type": "string" }
        }
      }
    }
  }
  ```
4. View the file in vscode:

<img width="366" alt="Image" src="https://github.com/user-attachments/assets/92534d93-ccfd-495f-918f-2f92dd1fe56c" />

## Environment
- [ ] Windows
- [x] Mac
- [x] Linux
- [ ] other (please specify)

pjsk-stripe:
This is solved by https://github.com/redhat-developer/yaml-language-server/pull/1076

PR #1076, opened 2025-05-15
Fix recursion bug in ast-conversion

pjsk-stripe:
### What does this PR do?
Currently, the code to convert AST nodes (in particular, resolve aliases) uses a recursive bit of code that will traverse through YAML aliases to resolve properties. For example:
```yaml
a: &a
  name: "foo"
b: 
  <<: *a
```

Parsing this code will produce two objects `a` and `b` that both have `name: "foo"`. The code that implements this alias resolution will only traverse to a depth of 1000, meaning it will give up on resolving aliases once it's traversed a really really long way down a path. This 1000 is very unlikely in human readable yaml files.

However, if you have a very large yaml files with many aliases, none of which come close to the `depth>1000` case, you can still experience errors due to a bug in the recursion. 

Internally we use the https://github.com/redhat-developer/vscode-yaml VSCode extension and we experienced this bug in a very large yaml file where all aliases in the file after the 1000th alias would report incorrect errors for missing properties. 

This boils down to the recursion correctly calling `refDepth++` to keep track of depth in the algorithm, but **not ever** calling `refDepth--`. Meaning regardless of depth, things fail after so many aliases are observed. 

This PR address the issue and adds some unit tests to validate things. 

https://github.com/redhat-developer/yaml-language-server/issues/1075

### Is it tested? How?
- [x] Added unit test for the depth case
- [x] Added unit test to check that the recursion works correctly
- [x] Tested locally via https://github.com/redhat-developer/vscode-yaml on our very large yaml files.

### DIFF
--- a/src/languageservice/parser/ast-converter.ts
+++ b/src/languageservice/parser/ast-converter.ts
@@ -32,14 +32,18 @@ import {
 
 type NodeRange = [number, number, number];
 
-const maxRefCount = 1000;
-let refDepth = 0;
+// Exported for tests
+export const aliasDepth = {
+  maxRefCount: 1000,
+  currentRefDepth: 0,
+};
+
 const seenAlias = new Set<Alias>();
 
 export function convertAST(parent: ASTNode, node: YamlNode, doc: Document, lineCounter: LineCounter): ASTNode | undefined {
   if (!parent) {
     // first invocation
-    refDepth = 0;
+    aliasDepth.currentRefDepth = 0;
   }
 
   if (!node) {
@@ -57,7 +61,7 @@ export function convertAST(parent: ASTNode, node: YamlNode, doc: Document, lineC
   if (isScalar(node)) {
     return convertScalar(node, parent);
   }
-  if (isAlias(node) && !seenAlias.has(node) && refDepth < maxRefCount) {
+  if (isAlias(node) && !seenAlias.has(node) && aliasDepth.currentRefDepth < aliasDepth.maxRefCount) {
     seenAlias.add(node);
     const converted = convertAlias(node, parent, doc, lineCounter);
     seenAlias.delete(node);
@@ -154,15 +158,18 @@ function convertScalar(node: Scalar, parent: ASTNode): ASTNode {
 }
 
 function convertAlias(node: Alias, parent: ASTNode, doc: Document, lineCounter: LineCounter): ASTNode {
-  refDepth++;
+  aliasDepth.currentRefDepth++;
   const resolvedNode = node.resolve(doc);
+  let ans: ASTNode;
   if (resolvedNode) {
-    return convertAST(parent, resolvedNode, doc, lineCounter);
+    ans = convertAST(parent, resolvedNode, doc, lineCounter);
   } else {
     const resultNode = new StringASTNodeImpl(parent, node, ...toOffsetLength(node.range));
     resultNode.value = node.source;
-    return resultNode;
+    ans = resultNode;
   }
+  aliasDepth.currentRefDepth--;
+  return ans;
 }
 
 export function toOffsetLength(range: NodeRange): [number, number] {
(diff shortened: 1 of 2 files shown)


src/languageservice/parser/ast-converter.ts
```diff
@@ -154,15 +158,18 @@ function convertScalar(node: Scalar, parent: ASTNode): ASTNode {
...
   if (resolvedNode) {
-    return convertAST(parent, resolvedNode, doc, lineCounter);
+    ans = convertAST(parent, resolvedNode, doc, lineCounter);
   } else {
     const resultNode = new StringASTNodeImpl(parent, node, ...toOffsetLength(node.range));
     resultNode.value = node.source;
-    return resultNode;
+    ans = resultNode;
   }
+  aliasDepth.currentRefDepth--;
```
pjsk-stripe, line 171: This is the important bit. I had to restructure things a tiny bit to make this testable.

test/yamlParser.test.ts
```diff
@@ -243,6 +244,65 @@ describe('YAML parser', () => {
...
+      aliasDepth.maxRefCount = 1;
+      const parsedDocument = parse(`
+a: &a
+  foo: "web"
+b: &b
+  <<: *a
+c: &c
+  <<: *b
+`);
+      aliasDepth.maxRefCount = 1000;
```
qaisjp, line 260: If `parse` crashes (due to bad code), `maxRefCount` will globally be left at `1`, which would cause other tests to fail perhaps improperly.

This may be safer:

```suggestion
      try {
        aliasDepth.maxRefCount = 1;
        const parsedDocument = parse(`
a: &a
  foo: "web"
b: &b
  <<: *a
c: &c
  <<: *b
`);
      } finally {
        aliasDepth.maxRefCount = 1000;
      }
```

qaisjp: I very rarely use `finally` so I checked that this actually works as intended in a REPL:

```js
const x = () => {
    console.log("before")
    try {
        console.log("set 1")
        throw new Error("thrown by q");
    } finally {
        console.log("unset")
    }
    console.log("after")
}

x()
```

produces

```
before
set 1
unset
Uncaught Error: thrown by q
    at x (<anonymous>:5:15)
    at <anonymous>:1:1
```

pjsk-stripe: applied manually, had a syntax error

pjsk-stripe:
Hi @msivasubramaniaan, can this get a look over? Thanks!

pjsk-stripe:
Hi @msivasubramaniaan, just following up, any chance I can get this reviewed?

datho7561:
The tests appear to crash because of an out of memory issue before completing, likely during the Billion Laughs test case. Are you sure this is the right fix?

pjsk-stripe:
Hi @datho7561 , good catch!

Yes I am still confident this is the right fix. I believe that test is failing because the original recursion bug was exiting early from the recursion after 1000 _laughs_. Once this is fixed, the recursion would traverse all laughs leading to an OOM. 

I fixed this in https://github.com/redhat-developer/yaml-language-server/pull/1076/commits/f1f1f5c0ab463934d56a9959a2a3525118dbbf03. The original code used a global `seen` variable that kept track of seen nodes which would eventually grow far beyond memory's limits while computing the same aliases over and over again. 

In the fix I switch to a cache that keeps track of what ASTNode each alias resolves to for future iterations. 

The tests now all pass in blazing time, do you mind taking another look?

datho7561:
I'm taking another look; I'm just trying to make sure that it doesn't ever calculate/store the 9 billion character string in memory.

It feels a little slow editing the Billion Laughs file, but this may be an existing bug, so I'll need to verify that.

datho7561:
Completion in the billion laughs file does slow down quite a lot after this fix (~1-2s to 2-5s). However, we just need to mitigate the out of memory effects of the attack, not make it perfect; people shouldn't be editing files with the exploit in it day to day.

datho7561 (approved):
Looks good to me, and addresses the linked issue. Thank you!

datho7561:
Sorry, one second, trying to figure out why the tests are failing CI.

PR #1080, opened 2025-06-05
Fix bool enums

Kosta-Github:
### What does this PR do?

schema validation is failing for `boolean` `enum` values:

schema:
```yaml
type: object
properties:
  test_value:
    enum: [true, false]
```

test:
```yaml
test_value: true # fails validation
```

### What issues does this PR fix or reference?

fix for: https://github.com/redhat-developer/yaml-language-server/issues/1078

this is the same fix as applied to `const` values here: https://github.com/redhat-developer/yaml-language-server/commit/e6165e4cf2b1c53ad800434031bdd30d4f00f6e2

### Is it tested? How?

The first commit in this PR just introduces a failing unit test show casing the issue.

The test is failing with:
```json
[
  {
    "code": 1,
    "data": {
      "schemaUri": [ "file:///default_schema_id.yaml" ],
      "values": [ true, false ]
    },
    "message": "Value is not accepted. Valid values: true, false.",
    "range": {
      "end": { "character": 15, "line": 0 },
      "start": { "character": 11, "line": 0 }
    },
    "severity": 1,
    "source": "yaml-schema: file:///default_schema_id.yaml"
  }
]
```

The second commit is fixing the issue and all tests are passing.

Issue #1078, opened 2025-05-22
Using `bool` values in `enum` and `const` generates a validation error for valid values


Kosta-Github:
## Describe the bug

When using a `boolean` value in an `enum` schema (e.g.: `enum: [true]`) validation fails for a valid value.

The error message is: `Value is not accepted. Valid values: true.`

For example, the `OpenAPI 3.0` spec is using such a mechanism in its schema.

## Expected Behavior

The schema validation should pass without an issue.

## Current Behavior

The schema validation generates this error message: `Value is not accepted. Valid values: true.`

<img width="577" alt="Image" src="https://github.com/user-attachments/assets/37fd4722-9fc1-44cd-a8dc-ca94253e0f35" />

## Steps to Reproduce

Place these 2 files next to each other:

`schema.yaml`:
```yaml
$schema: "http://json-schema.org/draft-07/schema#"
type: object
properties:
  value:
    enum: [true]
```

`test.yaml`:
```yaml
# yaml-language-server: $schema=./schema.yaml

value: true
```

## Environment

- [ ] Windows
- [x] Mac
- [ ] Linux
- [ ] other (please specify)

Latest version of the extension: `1.18.0`

Kosta-Github:
BTW, a similar error is created if using `boolean` in a `const`:

<img width="493" alt="Image" src="https://github.com/user-attachments/assets/f4fba6a9-41eb-46c6-a568-d5a876d4e489" />

## Steps to Reproduce

Place these 2 files next to each other:

`schema.yaml`:
```yaml
$schema: "https://json-schema.org/draft/2020-12/schema"
type: object
properties:
  value:
    const: true
```

`test.yaml`:
```yaml
# yaml-language-server: $schema=./schema.yaml

value: true
```

PR #1083, opened 2025-06-12
fixed Enum issue on schema.enum value

msivasubramaniaan:
### What does this PR do?
This PR fix the schema.enum value on array element

https://github.com/redhat-developer/yaml-language-server/issues/1078

### Is it tested? How?
Yes. Added test case

### DIFF
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -876,7 +876,7 @@ function validate(
       const val = getNodeValue(node);
       let enumValueMatch = false;
       for (const e of schema.enum) {
-        if (val === e || isAutoCompleteEqualMaybe(callFromAutoComplete, node, val, e)) {
+        if (equals(val, schema.enum, node.type) || isAutoCompleteEqualMaybe(callFromAutoComplete, node, val, e)) {
           enumValueMatch = true;
           break;
         }
--- a/src/languageservice/utils/objects.ts
+++ b/src/languageservice/utils/objects.ts
@@ -7,6 +7,15 @@
 export function equals(one: any, other: any, type?: any): boolean {
   if (type === 'boolean') {
     one = JSON.parse(one);
+    if (Array.isArray(other)) {
+      let i: number;
+      for (i = 0; i < other.length; i++) {
+        if (equals(one, other[i], type)) {
+          return true;
+        }
+      }
+      return false;
+    }
   }
   if (one === other) {
     return true;


datho7561:
This causes a regression:

In VS Code, add this to settings.json:
```json
{
  "yaml.schemas": {
    "https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/v1.32.1-standalone-strict/all.json": "*.k8s.yaml"
  }
}
```

Then, create a file `my-pod.k8s.yaml`:

```yaml
apiVersion: v1
kind: Pod
metadata:
  resourceVersion: test
```

`Pod` is listed as invalid, but it is suggested as an autocomplete option and it should be valid

src/languageservice/parser/jsonParser07.ts
```diff
@@ -876,7 +876,7 @@ function validate(
       const val = getNodeValue(node);
       let enumValueMatch = false;
       for (const e of schema.enum) {
-        if (val === e || isAutoCompleteEqualMaybe(callFromAutoComplete, node, val, e)) {
+        if (equals(val, schema.enum, node.type) || isAutoCompleteEqualMaybe(callFromAutoComplete, node, val, e)) {
```
Kosta-Github, line 879: I think this is wrong and should be:

```suggestion
        if (equals(val, e, node.type) || isAutoCompleteEqualMaybe(callFromAutoComplete, node, val, e)) {
```

src/languageservice/utils/objects.ts
```diff
@@ -7,6 +7,15 @@
...
     one = JSON.parse(one);
+    if (Array.isArray(other)) {
+      let i: number;
+      for (i = 0; i < other.length; i++) {
+        if (equals(one, other[i], type)) {
+          return true;
+        }
+      }
+      return false;
+    }
```
Kosta-Github, line 18: Adding this array handling for a boolean is not the right thing to add here.
Iterating over the enum values should be done at the caller site (see my other comment).

Kosta-Github (commented):
IMHO, the root of this issue is actually this location: https://github.com/redhat-developer/yaml-language-server/blob/e6165e4cf2b1c53ad800434031bdd30d4f00f6e2/src/languageservice/parser/jsonParser07.ts#L508-L509

which returns a `boolean` value as a `string` (`"true"` or `"false"`), since `node.source` is of type `string`.

I am not sure, why it is that way and returning `node.value` (which would be of the right type `boolean`) as is done above for `null`, `number`, and `string`?

Kosta-Github:
I still think, this is the more appropriate fix: https://github.com/redhat-developer/yaml-language-server/pull/1080
and all the CI tests were green...

PR #1085, opened 2025-06-22
fix/improve enum value descriptions for merged enum lists

Kosta-Github:
### What does this PR do?

fix/improve enum value descriptions for merged enum lists

This is a follow-up submission for #1028, which removes duplicate enum values in merged enum lists.

This fix/improvement ensures that the first **non-empty** enum description is used for the corresponding enum value.

Before this fix: if the first non-unique enum value was not providing a description but the second one was, it was not picked up and the enum value description stayed empty.

### Is it tested? How?

* added 3 additional unit test cases to the test suite, which ensure the correct/expected behavior
* `npm run test` is passing

### DIFF
--- a/src/languageservice/services/yamlHover.ts
+++ b/src/languageservice/services/yamlHover.ts
@@ -133,11 +133,15 @@ export class YAMLHover {
                   enumValue = JSON.stringify(enumValue);
                 }
                 //insert only if the value is not present yet (avoiding duplicates)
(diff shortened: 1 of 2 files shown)


msivasubramaniaan (approved):
LGTM

datho7561:
@msivasubramaniaan is this good to merge, then?

Issue #1081, opened 2025-06-06
Missing localization in some of the language services


sajeerzeji:
## Describe the bug
Some user-facing strings within the YAML language server are not being localized and appear in English, conflicting with the client-side localization of the library.

## Expected Behavior
All user-facing strings should be localized according to the application's locale settings. The following strings should be localized:

### `yamlValidation.ts`
  **NON-LOCALIZED:**
  - "Matches multiple schemas when only one must validate."

### `yamlCompletion.ts`
  **LOCALIZED:**
  - "Default value"
  **NON-LOCALIZED:**
  - "Inline schema"
  - "Create an item of an array"
  - "- (array item) "
  - "Allowed Values:"
  - "Example:"
  - "Source: [{}]"

### `yamlHover.ts`
  **NON-LOCALIZED:**
  - "JSON Schema"
  - "Allowed Values:"
  - "Example:"
  - "Source: [{}]"

### `yamlCodeActions.ts`
  **NON-LOCALIZED:**
  - "Jump to schema location ({})"
  - "Convert Tab to Spaces"
  - "Convert all Tabs to Spaces"
  - "Delete unused anchor: {}"
  - "Convert to boolean"
  - "Convert to block style {}"
  - "Fix key order for this map"

### `yaml-style.ts`
  **NON-LOCALIZED:**
  - "Flow style mapping is forbidden"
  - "Flow style sequence is forbidden"

### `unused-anchors.ts`
  **NON-LOCALIZED:**
  - "Unused anchor \"{}\""
  - "Unresolved alias \"{}\""

### `map-key-order.ts`
  **NON-LOCALIZED:**
  - "Wrong ordering of key \"{}\" in mapping"

### `yamlSchemaService.ts`
  **LOCALIZED:**
  - "$ref '{0}' in '{1}' can not be resolved."
  **NON-LOCALIZED:**
  - "Schema '{}' is not valid:\n{}"

### `jsonParser07.ts`
  **LOCALIZED:**
  - "Invalid color format. Use #RGB, #RGBA, #RRGGBB or #RRGGBBAA."
  - "String is not a RFC3339 date-time."
  - "String is not a RFC3339 date."
  - "String is not a RFC3339 time."
  - "String is not an e-mail address."
  - "String does not match IPv4 format."
  - "String does not match IPv6 format."
  **NON-LOCALIZED:**
  - "Property {0} is not allowed."
  - "Missing property \"{0}\"."
[...]

PR #1086, opened 2025-06-23
Added l10n bundle for various language and translate the local strings

msivasubramaniaan:
### What does this PR do?
This PR does the localization for YAML

https://github.com/redhat-developer/yaml-language-server/issues/1081

### Is it tested? How?
Yes. Updated existing test cases

### DIFF
--- a/src/yamlServerInit.ts
+++ b/src/yamlServerInit.ts
@@ -42,7 +42,7 @@ export class YAMLServerInit {
      */
     this.connection.onInitialize(async (params: InitializeParams): Promise<InitializeResult> => {
       const l10nPath: string = params.initializationOptions?.l10nPath;
-      const locale: string = params.initializationOptions?.locale;
+      const locale: string = params.locale;
       if (l10nPath) {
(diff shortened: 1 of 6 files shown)


sbouchet:
the Frensh translation is ok. good work

src/yamlServerInit.ts
```diff
@@ -38,9 +40,23 @@ export class YAMLServerInit {
      * Run when the client connects to the server after it is activated.
      * The server receives the root path(s) of the workspace and the client capabilities.
      */
-    this.connection.onInitialize((params: InitializeParams): InitializeResult => {
+    this.connection.onInitialize(async (params: InitializeParams): Promise<InitializeResult> => {
+      const l10nPath: string = params.initializationOptions?.l10nPath;
+      const locale: string = params.initializationOptions?.locale;
```
datho7561, line 45: The locale is already passed from VS Code to the language server as `params.locale`. I think we can read it from there instead, which means we also don't need to send it explicitly in vscode-yaml

msivasubramaniaan:
@datho7561 FYR 
image

PR #1090, opened 2025-06-27
Added test case for l10n localization and all configuration through test helper instead of script file

msivasubramaniaan:
### What does this PR do?
Added test case for l10n localization and all configuration through test helper instead of script file

https://github.com/redhat-developer/vscode-yaml/pull/1131
https://github.com/redhat-developer/vscode-yaml/issues/1132

### Is it tested? How?
Yes added test case


datho7561:
I think this needs to be rebased, it seems like some of these commits have already been merged as a part of https://github.com/redhat-developer/yaml-language-server/pull/1086.

Locally, I rebased and dropped all commits except the latest one "added test cases for bundle and configuration through testhelper class", and the tests are passing for that one

msivasubramaniaan:
Handle through https://github.com/redhat-developer/yaml-language-server/pull/1093

PR #1092, opened 2025-07-03
Fix array of const completion

ShadiestGoat:
### What does this PR do?
This PR focuses on fixing array of const issues described in https://github.com/redhat-developer/vscode-yaml/issues/879
It also takes care not to regress on the error fixed in https://github.com/redhat-developer/yaml-language-server/pull/670, but takes a step further in fixing that issue

https://github.com/redhat-developer/vscode-yaml/issues/879
#620 

### Is it tested? How?
Unit tests (pending real world tests)


datho7561:
Taking a look now...

datho7561:
Something's up. I still get the bug from #620. Try running it in vscode-yaml and open completion here:


```yaml
test:
  - value1
    |
```

ShadiestGoat:
Looking

ShadiestGoat:
I guess that makes sense with my change... Sigh, back to the drawing board I go.

I will say though - the correction made in the other PR is not the right fix for this, as it breaks enum/const array completion

datho7561:
I agree. _Something's_ up.

ShadiestGoat:
When investigating, I found something weird - the closest node discoverer function returns the null value node for when the key is not filled, but returns the map node when the key IS filled. I think my next attempt will be to fix it there.

ShadiestGoat:
I think I have an idea for how to fix it - the idea is "if on an empty line && node is null, don't autocomplete scalar values"

datho7561:
That makes sense to me.

ShadiestGoat:
Hey @datho7561 - I think I got it (mostly?) working. It all works in tests BUT theres a weird case where, in your case, it suggests an empty line, but it wants to add it on the wrong indent. Imo thats out of scope for this PR, since I've experienced this before

ShadiestGoat:
image

(I have l10n disabled bc broken)

image
image

ShadiestGoat:
Wait hold, found issue w/ docs

ShadiestGoat:
I can't look into this further today, but for some reason when testing on a real vscode client, it doesn't load documentation in suggestions:

image

But the test for this passes so idk whatsup

(there is documentation attached)

image

ShadiestGoat:
Scratch previous comments - that was my local env being odd. @datho7561 would you mind taking a look at this PR again?

datho7561 (changes requested):
I tried the updated fix and I still get completions for the `const` values even if there isn't a `-` when trying it from within VS Code. I also tried it in Neovim and same thing.

Can you confirm this is the case on your end? Maybe something is off with my environment too.

ShadiestGoat:
Hey - I'm sorry, I don't think I understand what you mean. Can you share a yaml snippet of where you're getting incorrect completions?

datho7561:
It's the same case I mentioned before that's behaving weird. The schema this file is pointing to is the one you've used in the tests. Open completion where the `|` is:

```yaml
# yaml-language-server: $schema=./array-of-const.json
test:
  - value1
  |
```

You get 3 completion results. The first adds `- `, so I think it's valid. The second two are `value1` and `value2`. They aren't valid, given that a `- ` is needed first. These two completion items are new behaviour introduced by this PR.

We either need to stop the `const` values from being suggested there, or have them also insert the `- `.

ShadiestGoat:
Ill take a look, thanks for clarifying!

ShadiestGoat:
I am able to reproduce this, looks like another leak source, I'll find it

ShadiestGoat:
Heya @datho7561 - I got that one fixed as well ^^ - can you try again?

datho7561:
@ShadiestGoat that case still seems to be a bit broken in VS Code. Something's different between the test cases and actually running it in VS Code. That's not a good sign, that probably warrants an investigation (in a separate PR, I can do that).

I did some debugging and I think that I figured out a solution; here's what I have:

```diff
diff --git a/src/languageservice/services/yamlCompletion.ts b/src/languageservice/services/yamlCompletion.ts
index ccfb9d1..f83907b 100644
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -516,8 +516,7 @@ export class YamlCompletion {
       const ignoreScalars =
         textBuffer.getLineContent(overwriteRange.start.line).trim().length === 0 &&
         originalNode &&
-        isScalar(originalNode) &&
-        originalNode.value === null;
+        ((isScalar(originalNode) && originalNode.value === null) || isSeq(originalNode));

       // completion for object keys
       if (node && isMap(node)) {
@@ -939,7 +938,7 @@ export class YamlCompletion {
                   separatorAfter,
                   collector,
                   types,
-                  false,
+                  ignoreScalars,
                   typeof s.schema.items === 'object' &&
                     (s.schema.items.type === 'object' || isAnyOfAllOfOneOfType(s.schema.items))
```

ShadiestGoat:
that makes total sense - good catch. I assume this works just because if we are in an empty line & we are in a seq, then we can always ignore scalars. I mean why would we suggest a scalar in an such a case?

I can patch this in and test in my evening, thanks for the testing and suggestions!

ShadiestGoat:
Thats weird I'm still getting the same issue

ShadiestGoat:
image

ShadiestGoat:
Never mind... I don't know how to properly develop vscode extensions - it looks like it works when launching using the vscode launcher

image
image
image
image
image

ShadiestGoat:
So... I think this is ready to merge! Please let me know if I need to do anything else ^^

ShadiestGoat:
As a side note, a follow up here would be to improve how non-scalars are added:

Eg. this works:

image
image

But, this works but imo its not great:
image
image

ShadiestGoat:
Also, object completion:

image
image

This is _fine_:

image
image

Also broken:
image
image

datho7561 (approved):
Looks good to me and works well. Thanks, Lucy!

PR #1095, opened 2025-07-10
Fix NPM broken

msivasubramaniaan:
### What does this PR do?
NPM broken https://github.com/redhat-developer/yaml-language-server/actions/runs/16197165836/job/45730613032#step:8:15

https://github.com/redhat-developer/yaml-language-server/actions/runs/16197165836/job/45730613032#step:8:15
https://github.com/redhat-developer/yaml-language-server/issues/1097

### Is it tested? How?
NA


datho7561:
what does `always-auth` actually do? I can't find the documentation on the npm website.

[The documentation for using tokens](https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow#set-the-token-as-an-environment-variable-on-the-cicd-server) suggests that maybe the the environment variable should be named `NPM_TOKEN` instead of `NODE_AUTH_TOKEN`

datho7561:
It seems `always-auth` was deprecated and removed in `npm` v8: https://stackoverflow.com/questions/74618972/why-is-npms-npm-config-set-always-auth-not-a-valid-option-npm-8

msivasubramaniaan:
> It seems `always-auth` was deprecated and removed in `npm` v8: https://stackoverflow.com/questions/74618972/why-is-npms-npm-config-set-always-auth-not-a-valid-option-npm-8

Trying with https://github.com/actions/setup-node/issues/81#issuecomment-623018474

datho7561:
Okay, that looks like it will solve the problem

.gitignore
```diff
@@ -8,5 +8,4 @@ out/server
 yarn-error.log
 coverage
 lib/
-package-lock.json
```
datho7561, line 11: add yarn.lock here so that we don't accidentally commit it in the future

Dockerfile
```diff
@@ -4,8 +4,8 @@ WORKDIR /yaml-language-server
 
 COPY . .
 
-RUN yarn install && \
-    yarn run build
+RUN npm run install && \
```
datho7561, line 7: It should be `npm install`

README.md
```diff
@@ -400,11 +400,11 @@ This repository only contains the server implementation. Here are some known cli
 3. Install the dependencies
    ```bash
    cd yaml-language-server
-   $ yarn install
+   $ npm run install
```
datho7561, line 403: `npm install`

package.json
```diff
@@ -24,7 +24,7 @@
   ],
   "repository": {
     "type": "git",
-    "url": "https://github.com/redhat-developer/yaml-language-server.git"
+    "url": "git+https://github.com/redhat-developer/yaml-language-server.git"
```
datho7561, line 27: I'm pretty sure this is just used for the link on the npm website. I think it should remain as-is instead of having `git+` at the beginning, because adding `git+` will break that link

datho7561 (changes requested):

PR #1096, opened 2025-07-10
@vscode/l10n is a runtime dependency

polyzen:
### What does this PR do?

Moves @vscode/l10n to runtime dependencies.

From my Neovim lsp.log:
```
[ERROR][2025-07-10 13:40:08] ...p/_transport.lua:36 "rpc" "yaml-language-server"  "stderr"  "node:internal/modules/cjs/loader:1404\n  throw err;\n  ^\n\nError: Cannot find module '@vscode/l10n'\nRequire stack:\n- /usr/lib/node_modules/yaml-language-server/out/server/src/languageservice/services/yamlSchemaService.js\n- /usr/lib/node_modules/yaml-language-server/out/server/src/languageservice/yamlLanguageService.js\n- /usr/lib/node_modules/yaml-language-server/out/server/src/yamlServerInit.js\n- /usr/lib/node_modules/yaml-language-server/out/server/src/server.js\n- /usr/lib/node_modules/yaml-language-server/bin/yaml-language-server\n    at Function._resolveFilename (node:internal/modules/cjs/loader:1401:15)\n    at defaultResolveImpl (node:internal/modules/cjs/loader:1057:19)\n    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1062:22)\n    at Function._load (node:internal/modules/cjs/loader:1211:37)\n    at TracingChannel.traceSync (node:diagnostics_channel:322:14)\n    at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)\n    at Module.require (node:internal/modules/cjs/loader:1487:12)\n    at require (node:internal/modules/helpers:135:16)\n    at Object.<anonymous> (/usr/lib/node_modules/yaml-language-server/out/server/src/languageservice/services/yamlSchemaService.js:12:14)\n    at Module._compile (node:internal/modules/cjs/loader:1730:14) {\n  code: 'MODULE_NOT_FOUND',\n  requireStack: [\n    '/usr/lib/node_modules/yaml-language-server/out/server/src/languageservice/services/yamlSchemaService.js',\n    '/usr/lib/node_modules/yaml-language-server/out/server/src/languageservice/yamlLanguageService.js',\n    '/usr/lib/node_modules/yaml-language-server/out/server/src/yamlServerInit.js',\n    '/usr/lib/node_modules/yaml-language-server/out/server/src/server.js',\n    '/usr/lib/node_modules/yaml-language-server/bin/yaml-language-server'\n  ]\n}\n\nNode.js v22.17.0\n"
```

### Is it tested? How?
Applied in https://gitlab.archlinux.org/archlinux/packaging/packages/yaml-language-server/-/commit/5888f52cadeb012d6c6df73aaad4babd939212ff and tested on a `.gitlab-ci.yml` file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "url": "https://github.com/redhat-developer/yaml-language-server.git"
   },
   "dependencies": {
+    "@vscode/l10n": "^0.0.9",
     "ajv": "^8.17.1",
     "ajv-draft-04": "^1.0.0",
     "lodash": "4.17.21",
@@ -48,7 +49,6 @@
     "@types/sinon-chai": "^3.2.5",
     "@typescript-eslint/eslint-plugin": "^5.38.0",
     "@typescript-eslint/parser": "^5.38.0",
-    "@vscode/l10n": "^0.0.9",
     "@vscode/l10n-dev": "^0.0.35",
     "chai": "^4.2.0",
     "coveralls": "3.1.1",


datho7561:
I was able to reproduce the bug, and I'm checking the solution now. (To reproduce the bug, run `yarn --production` after building, which will remove the development dependencies).

datho7561:
This resolves the problem about the missing library. However, I get an error when running in neovim; it expects the l10n bundle to be in a different directory than it is. This prevents the server from starting. I think this is because the current working directory of the language server isn't configured the way the server expects it to be or something like that. I'll take a look into that.

datho7561 (approved):
This is needed to get the language server working properly when it's running outside of a development environment

lotheac:
> This resolves the problem about the missing library. However, I get an error when running in neovim; it expects the l10n bundle to be in a different directory than it is. This prevents the server from starting. I think this is because the current working directory of the language server isn't configured the way the server expects it to be or something like that. I'll take a look into that.

I worked around this in Alpine by [putting the l10n bundle directory at `../l10n/` relative to the yaml-language-server bundle](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/87215). It would feel less hacky if the bundles could be loaded from `./l10n/` instead for my case, but I'm not sure what kind of directory structures other packagers are using.

datho7561:
@lotheac @msivasubramaniaan @polyzen I guess we should replace the fallback path here:

https://github.com/redhat-developer/yaml-language-server/blob/main/src/yamlServerInit.ts#L62

to be `../../../l10n`, since the built `server.js` is under `./out/server/src/server.js`. I've tried this locally by commenting out https://github.com/redhat-developer/vscode-yaml/blob/main/src/extension.ts#L134 and also with neovim; it works as expected in both cases.

@lotheac it seems you're bundling the extension with esbuild and producing a completely different directory structure. We don't bundle at all in our build. We're expecting most people to be installing through `npm i -g`, and not through the distribution's package manager. I think the hack that you found is the best way to accommodate the different directory structure that you produce.

datho7561:
See #1104

lotheac:
> @lotheac it seems you're bundling the extension with esbuild and producing a completely different directory structure. We don't bundle at all in our build. We're expecting most people to be installing through `npm i -g`, and not through the distribution's package manager. I think the hack that you found is the best way to accommodate the different directory structure that you produce.

thanks for your comment; your observation is correct. I didn't want the Alpine package to ship files that are unnecessary to run yaml-language-server, so I went off the beaten path and used esbuild :)

I suppose I'll adapt it to use a similar path as the npm installation for the entrypoint server.js, as well as the l10n json files, following your patches.

PR #1107, opened 2025-07-15
Remove coveralls dependency

### LINKED ISSUE #1108: Remove `coveralls` as a dependency
## Describe the bug
We've included `coveralls` as a developer dependency, but I don't think it's actually used anywhere. I think all we need is the GitHub Action in order to get coverage reporting.

## Expected Behavior
No coveralls dependency

## Current Behavior
coveralls dependency

## Steps to Reproduce
N/A

## Environment
- [x] Windows
- [x] Mac
- [x] Linux
- [ ] other (please specify)

datho7561:
### What does this PR do?
Fixes #1108

Remove `coveralls` as a developer dependency

From my understanding, we just need the coveralls GitHub Action in order for things to work.

`coveralls` is not used anywhere, so removing it should reduce the size of `./node_modules` for people developing yaml-language-server.

Also, `coveralls` might get flagged as a false positive security bug because it depends on `request`. By removing it, we won't get that false positive.

### Is it tested? How?
CI

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,6 @@
         "@typescript-eslint/parser": "^5.38.0",
         "@vscode/l10n-dev": "^0.0.35",
         "chai": "^4.2.0",
-        "coveralls": "3.1.1",
         "eslint": "^8.24.0",
         "eslint-config-prettier": "^9.0.0",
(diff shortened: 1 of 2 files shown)


msivasubramaniaan (approved):
LGTM

PR #1110, opened 2025-07-18
Added encoding dependency

msivasubramaniaan:
### What does this PR do?
This PR resolves the https://github.com/redhat-developer/vscode-yaml/issues/1139

https://github.com/redhat-developer/vscode-yaml/issues/1139

### Is it tested? How?
Yes


datho7561 (approved):
I can't reproduce https://github.com/redhat-developer/vscode-yaml/issues/1139, I don't know what's different about my setup though.

I think this change is save to make; the code still compiles after it.

PR #1113, opened 2025-07-21
Fixed broken link to JSON schema website

webdevred:
### What does this PR do?

This PR fixes a broken link in the README.
The previous link to https://www.schemastore.org/json/ is no longer valid (returns 404).

Updated with the correct working URLs:
- https://json.schemastore.org/
- https://schemastore.org/api/json/catalog.json

Both links are valid and point to the full JSON Schema catalog used by the YAML language server.
Let me know if you'd prefer just one of them in the README, or if it makes sense to include both.


datho7561:
I think pointing to schemastore.org is fine, given it provides a search bar to search through the schemas.

PR #1127, opened 2025-09-23
Fix release GitHub Action

datho7561:
### What does this PR do?
- setup-node@v5 creates a .npmrc if you provide a registry, so there should be no need to do that separately
- after setup-node@v5 is run, `npm publish` expects the token to be accessible from the environment variable `NODE_AUTH_TOKEN`
- also fix coveralls so that it only publishes results on ubuntu. If results are published from multiple platforms, then the 2nd and 3rd publish attempts will fail.

See https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#publish-to-npmjs-and-gpr-with-npm

This should fix the release GitHub Action (as well as the prereleases)

### Is it tested? How?
No, I don't think there's a good way to test it without merging first.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -35,9 +35,10 @@ jobs:
 
       # Set up Node
       - name: Use Node 20
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v5
         with:
           node-version: 20
+          registry-url: 'https://registry.npmjs.org'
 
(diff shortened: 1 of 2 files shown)


datho7561:
Supposedly the coveralls failures are related to an ongoing outage, however I believe the changes I made will mitigate failures anyways.

rgrunber (approved):

PR #1133, opened 2025-10-01
Refactoring to change strings into block strings

### LINKED ISSUE #1119: Code action: convert inline strings to block (and vice versa)

Our deployments (uncontrollably) convert block strings to long inline strings. When editing the resulting files in the live server, it would be great to be able to reliably convert them back to a block string.

Wrapping long lines to folded scalars would be welcome as well.

Given 

```
field: "<1000 word novel>"
```

If the cursor is in the string, invoking code actions should present two options:
* convert to block scalar
* convert to flow scalar

Because I assume that sometimes the opposite would be desirable:
```
field: |
  <100 lines of prose>
```

If the cursor is in the string, code actions should include:
* convert to inline scalar

A series of search and replace? Very carefully editing the string without changing it.

datho7561:
### What does this PR do?
Adds two refactorings that change inline strings (i.e. ~~unquoted~~, single quoted, double quoted) into block strings. One for folded block string and one for literal block string.

eg. if you have the following

```yaml
root: "value1\value2"
```

There are two refactoring actions so that it can be turned into

```yaml
root: >-
  value1

  value2
```

or

```yaml
root: |-
  value1
  value2
```

Fixes #1119

### Is it tested? How?
- [x] Unit tests

### DIFF
--- a/test/yamlCodeActions.test.ts
+++ b/test/yamlCodeActions.test.ts
@@ -865,5 +865,28 @@ animals: [dog , cat , mouse]  `;
       const result = actions.getCodeAction(doc, params);
       expect(result).to.have.length(0);
     });
+    it('process block string conversion only til closing quote', function () {
+      const docContent =
+        'longline: "ine1thisisaverylonglinethisisaverylonglineine1thisisaverylonglinethisisaverylongline"\n\n# a line';
+      const doc = setupTextDocument(docContent);
+      const params: CodeActionParams = {
+        context: CodeActionContext.create([]),
+        range: Range.create(Position.create(0, 11), Position.create(0, 12)),
+        textDocument: TextDocumentIdentifier.create(TEST_URI),
+      };
+      const actions = new YamlCodeActions(clientCapabilities);
+      const result = actions.getCodeAction(doc, params);
+      expect(result).to.have.length(1);
+      const edit0: WorkspaceEdit = {
+        changes: {},
+      };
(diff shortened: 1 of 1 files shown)


datho7561:
@nyarly if you have time, do you mind trying this out to see if it works well for you? I think it's mostly working, but there might be some edge cases I haven't considered.

If you need help getting the project set up, free free to ask about that.

fbricon:
You get as many code action pairs as there are new lines in string literals

Screenshot 2025-12-02 at 16 35 40

datho7561:
> You get as many code action pairs as there are new lines in string literals

Should be fixed, I wasn't checking against the cursor position/selection range so I was suggesting all potential string conversions in the document.

nyarly:
@datho7561 Let me see about getting this set up

shin19991207:
I tested manually and it works well! I only noticed the block-string conversion consumes the newline after a quoted scalar when one exists. For example:
```
longline: "ine1thisisaverylonglinethisisaverylonglineine1thisisaverylonglinethisisaverylongline"

# a line
```
After converting to folded block string, it becomes:
```
longline: >-
  ine1thisisaverylonglinethisisaverylonglineine1thisisaverylonglinethisisaverylongline
# a line
```

I made a small change so the code action only replaces up to the closing quote and preserves everything after it.

datho7561:
I'll add a test for that

PR #1137, opened 2025-10-09
Address 'no scope' warning

datho7561:
### What does this PR do?
A scope should be provided when accessing `[yml]` settings, otherwise a warning is emitted. This change provides `file:///` as the scope.

Fixes https://github.com/redhat-developer/vscode-yaml/issues/972

### Is it tested? How?
Yes; verified manually that the warning goes away in VS Code.

### DIFF
--- a/src/languageserver/handlers/settingsHandlers.ts
+++ b/src/languageserver/handlers/settingsHandlers.ts
@@ -40,7 +40,7 @@ export class SettingsHandler {
     const config = await this.connection.workspace.getConfiguration([
       { section: 'yaml' },
       { section: 'http' },
-      { section: '[yaml]' },
+      { section: '[yaml]', scopeUri: 'null' },
       { section: 'editor' },
       { section: 'files' },
     ]);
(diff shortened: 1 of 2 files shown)


datho7561:
I should test that setting the scope to `file:///` works on Windows, but unfortunately I don't have access to my Windows computer today.

fbricon:
sending `file:///` seems wrong. The error message from https://github.com/redhat-developer/vscode-yaml/issues/972 says:

> provide the URI of a resource or 'null' for any resource.

so I would use `null`

datho7561:
I double checked on Windows, and this doesn't fix the problem there, but it also doesn't cause any new problems there.

fbricon:
Tested on my mac with :
> scopeUri: null

problem remains, but with: 
> scopeUri: 'null' 

warnings are gone, and [yaml] section settings are still picked up.

datho7561:
I'll try it on Windows and push the change if it works

fbricon:
thanks @datho7561

PR #1138, opened 2025-10-10
feat: add errorMessage from schema when the property is missing

p-spacek:
### What does this PR do?
adds an errorMessage from the schema when the property is missing
allows customization of the errorMessage in more complicated scenarios, like 

```
anyOf: [
          {
            required: ['title'],
            errorMessage: 'At least one of `title` or `icon` must be defined.',
          },
          {
            required: ['icon'],
          },
        ],
```

### Is it tested? How?
test

### DIFF
--- a/src/languageservice/parser/jsonParser07.ts
+++ b/src/languageservice/parser/jsonParser07.ts
@@ -1251,9 +1251,7 @@ function validate(
           validationResult.problems.push({
             location: location,
             severity: DiagnosticSeverity.Warning,
-            message:
-              getWarningMessage(ProblemType.missingRequiredPropWarning, [propertyName]) +
-              (schema.errorMessage ? ` ${schema.errorMessage}` : ''),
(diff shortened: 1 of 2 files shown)


datho7561:
~~Does any other YAML/JSON validation service support this? It seems like an odd thing for us to introduce in the yaml-language-server unless it's in the JSON Schema specification or there's an established convention for doing this.~~

Yes, VS Code's JSON language server supports it.

~~I can see that vscode-json-languageservice has it listed here: https://github.com/microsoft/vscode-json-languageservice/blob/e0f39f09a3e693f3a0895a051efd4895cbc6a746/src/jsonSchema.ts#L78, but I can't seem to get it to actually show the message I set when working with JSON in VS Code.~~

Nevermind, got it working in VS Code. The specific scenario I tried initially didn't work properly, though.

Also, it seems we already support `errorMessage` for many scenarios, my bad.

Issue #1145, opened 2025-11-20
BigInt keys false reported as duplicate in map objects


jovfer:
## Describe the bug
If a map contains several keys represented as large integers, false-positive key duplication error is reported

## Expected Behavior
No error

## Current Behavior
Key duplication error

## Steps to Reproduce
The following yaml should be valid, but validation failed
```
123456789012345671: 1
123456789012345672: 2
```

## Environment
- [ ] Windows
- [x] Mac
- [ ] Linux
- [ ] other (please specify)
Tested on Mac, likely general issue

datho7561:
~~This is a limitation of eemeli/yaml, which we use for parsing files. Here's some added context: https://github.com/eemeli/yaml/issues/657#issuecomment-3895966623, https://github.com/eemeli/yaml/issues/660~~

datho7561:
Took a look at the PR you opened, it seems to fix the issue

PR #1146, opened 2025-11-20
Add regression test for equivalent bigint map keys

jovfer:
### What does this PR do?
Draft attempt to fix an issue, generated by Codex

#1145 

### Is it tested? How?
Added a couple of tests for the particular scenario.
`should not report big integer key duplication error` is failing without the patch

### DIFF
--- a/test/schemaValidation.test.ts
+++ b/test/schemaValidation.test.ts
@@ -557,6 +557,34 @@ describe('Validation Tests', () => {
         })
         .then(done, done);
     });
+
+    it('Big integer value should validate as type:integer', async () => {
+      schemaProvider.addSchema(SCHEMA_ID, {
+        type: 'object',
+        properties: {
+          bignum: {
+            type: 'integer',
+          },
+        },
+      });
+      const content = 'bignum: 123456789012345671';
+      const result = await parseSetup(content);
+      assert.equal(result.length, 0);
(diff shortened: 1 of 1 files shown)


datho7561:
This seems to fix the issue. Is there anything missing from this PR, or is it ready to go? Do you mind rebasing it?

jovfer:
@datho7561 thanks for checking the issue and the PR. The main reason I marked it as draft is my confidence level in the fix.

I managed to isolate the issue, cover it with a test, and generate a hotfix with AI, but honestly I’m not yet confident enough to say whether this workaround is good enough, or whether the better path would be fixing the underlying library and adding a patch on top.

Any suggestions on additional tests, maybe at a higher level, that we could add to validate correctness better?

And sure, I’ll rebase the patch later this week and we can at least verify that it still passes the existing tests.

datho7561:
> I’ll rebase the patch later this week

Thanks!

> I managed to isolate the issue, cover it with a test, and generate a hotfix with AI, but honestly I’m not yet confident enough to say whether this workaround is good enough, or whether the better path would be fixing the underlying library and adding a patch on top.

Thinking about it more, there's many places where we're checking the type of the `value` of a node. One thing to test against might be if JSON schema-based validation of large integers still works properly.

PR #1150, opened 2025-12-05
Add anchor preview in hover

Hugo-Hache:
### What does this PR do?

It displays the preview of an anchor when hovering it.

image

This is a very rough proposal to check if you would be interested in merging this feature. If this is the case, I'll put more time in writing the test and improving the code based on your remarks.

### DIFF
--- a/src/languageservice/services/yamlHover.ts
+++ b/src/languageservice/services/yamlHover.ts
@@ -229,7 +229,7 @@ export class YAMLHover {
     const unprocessedProperties: PropertyASTNode[] = [...node.properties];
 
     while (unprocessedProperties.length > 0) {
-      const propertyNode = unprocessedProperties.pop();
+      const propertyNode = unprocessedProperties.shift();
       const key = propertyNode.keyNode.value;
 
       if (key === '<<' && propertyNode.valueNode) {


datho7561:
I think this sounds helpful. I can find time to review it, though I can't promise I'll be fast.

I tried it out, and I have a good initial impression of it. I haven't tested it thoroughly or looked closely at the code yet though.

I noticed it was disabled by default, any reason for that? It seems like a helpful feature, and I can't think of drawbacks of enabling it by default. People need to go out of their way to find any features that aren't enabled by default, which could mean it doesn't get used.

Hugo-Hache:
As I didn't know if it was a common use case I conservatively set the default to false, but I agree with you that plenty of users might discover this hover by serendipity and enjoy it.

For the resolveYamlMapMergeKeys function I guess we would need to limit the number of recursion to prevent performance issue in very large YAML (but I noticed there is a maxItemsComputed, so maybe it's enough to prevent it).

Anyway it's good news it looks helpful to you. Thanks for the time you'll invest in testing/reviewing it.

Hugo-Hache:
* Changed shouldHoverAnchor to default as true
* Added MAX_MERGE_RECURSION_LEVEL (and simplified code by removing resolveYamlMapMergeKeys or seenKeys)

I committed a max recursion level of 10, does it look good to you? (screenshot with a modified local limit of 3)

Screenshot 2025-12-10 at 18 05 39

datho7561 (dismissed):
Looks good and works well! Thanks, Hugo!

I'm going to resolve the merge conflict, seems it's just in the imports

datho7561:
Ahh shoot it's susceptible to [billion laughs](https://en.wikipedia.org/wiki/Billion_laughs_attack). Oh well, not a huge deal.

Hugo-Hache:
> Ahh shoot it's susceptible to [billion laughs](https://en.wikipedia.org/wiki/Billion_laughs_attack). Oh well, not a huge deal.

TIL billion laughs. If I understand correctly in addition to exploration depth (MAX_MERGE_RECURSION_LEVEL = 10), we should add a limit on width (ie number of aliases resolved at the same level)? 

In my understanding even we if we put a width like 5 (already restrictive as it looks rather common to have 5 anchors at the same level), we still have 5^9 (1.2M) nodes. 

Maybe a global maximum of nodes explored would make more sense as it would prevent any performance issue, while working on cases where either depth or width is >=5?

datho7561:
Yeah, the usual solution is to maintain a maximum number of nodes explored, and stop exploring past some upper bound. It's kind of tricky to implement. However, it's an edge case and I don't think it's pressing to address. If the user runs into it all that happens is memory is momentarily eaten up by VS Code/whatever editor until they close it.

PR #1153, opened 2025-12-10
Allow .editorconfig for prettier

joedevivo:
### What does this PR do?

`prettier` wasn't respecting my `.editorconfig`. This PR creates an option `format.editorConfig`, and when it's `true`, will pass that into `prettier`, respecting those settings if you opt in.

### Is it tested? How?
Built locally, tested in neovim with `nvim-lspconfig`.

```lua
opts = {
  servers = {
    yamlls = {
      cmd = "path/to/local/yaml-language-server",
      settings = {
        yaml = {
          format = {
            enable = true,
            editorConfig = true,
          },
        },
      },
    },
  },
}
```

### DIFF
--- a/src/languageservice/services/yamlFormatter.ts
+++ b/src/languageservice/services/yamlFormatter.ts
@@ -35,6 +35,7 @@ export class YAMLFormatter {
       const prettierOptions: Options = {
         parser: 'yaml',
         plugins: [yamlPlugin, estreePlugin],
+        editorConfig: options.editorConfig,
 
         // --- FormattingOptions ---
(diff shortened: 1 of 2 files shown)


datho7561:
In the current configuration, what does this setting default to? I think it would be helpful to enable it by default, unless you think that will be annoying (if someone has configured formatting settings in an .editorconfig they probably want them respected)

datho7561 (changes requested):

datho7561:
We no longer use prettier for formatting internally.

Issue #856, opened 2023-03-06
Support for json-schema > draft-7


mitodrummer:
## Summary
Hello, I am wondering if there is a plan to update to the newest json schema (2020-12 at the time of this ticket). I am needing features from 2019 schema.

TY

mitodrummer:
https://github.com/microsoft/vscode-json-languageservice/pull/133

Looks like maybe version 4.2.1 of json-languageservice should have it :)

remcohaszing:
FYI #637 explains why the version is locked. That issue includes a number of related references.

As the reporter of that issue I would totally understand if the version is unlocked in order to support newer JSON schema features.

gorkem:
@remcohaszing I think we will have to move to a later vscode-json-languageservice not only to get schema support for more recent schema versions but also for issues like #823 are also fixed by later versions.

gorkem:
No problem. Unfortunately, there are breaking changes on vscode-language-service that we need to adjust yaml-language-server for.

amimas:
Any update available for this?

apupier:
> [microsoft/vscode-json-languageservice#133](https://github.com/microsoft/vscode-json-languageservice/pull/133)
> 
> Looks like maybe version 4.2.1 of json-languageservice should have it :)

Unfortunately, it seems that it is not workign as expected, this issue is linked from the pointed issue https://github.com/microsoft/vscode/issues/98724

remcohaszing:
I don’t know what’s the hold up for [SchemaStore/schemastore#2858](https://github.com/SchemaStore/schemastore/pull/2858), but it’s taking a long time. It only affects 3 schemas, of which I manage 1, so I can create a workaround for that.

I think it’s best to just update. This will break for a couple of users, but benefit the majority.

michaeltlombardi:
Updating would be very helpful, as there's a good number of useful keywords in 2019-09 that work even without full support for dynamic references/anchors/vocabulary - notable, the `unevaluated*` and `dependent*` keywords for schema composition/extension/conditional validation.

Cerebus:
Updating is a must, b/c I've got schemas I'd like to use that are on draft 2019-09 and 2020-12. It's a royal PITA when completion is busted for half my stuff b/c `yamlls` won't grok the schema.

baywet:
@datho7561 could you please articulate in broad strokes where the changes need to be made for this? Sort of an introduction to the code base + considerations to keep in mind.
[...]

PR #1158, opened 2025-12-17
[WIP] proof-of-concept for using hyperjump for validation

datho7561:
### What does this PR do?
This is a proof of concept that replaces the internal json schema validator with a 

It's able to point out the correct error range but not much else; the error messages are not helpful and it only works if the schema ends in `.schema.yaml`

#856

### Is it tested? How?
I tried out some basic features, it needs a lot of work before it can be merged.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -6,7 +6,7 @@
     {
       "label": "watch typescript",
       "type": "shell",
-      "command": "yarn run watch",
+      "command": "npm run watch",
       "presentation": {
         "reveal": "never"
       },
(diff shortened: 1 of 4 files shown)


jdesrosiers (commented):
Is it really just about providing validation diagnostics? If that's all that needs to be done, this should be fairly easy to knock out. I assumed the hard part would be integrating editor completions and such. Is the plan to just use the vscode implementation for that part? Maybe that works well enough even though it doesn't fully understand draft 2020-12?

datho7561:
> Is it really just about providing validation diagnostics?

I think we'll likely need to adapt our existing validator to accommodate the new features, because, as you mentioned, we'll need to understand and use the schemas for other parts of the language server outside of just validation.

Issue #1151, opened 2025-12-05
Some links are escaped making them unusable


klonuo:
In the markdown provided by hover action, dots and dashes are escaped, making at least links unusable:

```json
{"contents": {"kind": "markdown", "value": "#### GitHub Workflow\n\nSelects an action to run as part of a step in your job\\. An action is a reusable unit of code\\. You can use an action defined in the same repository as the workflow, a public repository, or in a published Docker container image \\(https://hub\\.docker\\.com/\\)\\.\n\nWe strongly recommend that you include the version of the action you are using by specifying a Git ref, SHA, or Docker tag number\\. If you don't specify a version, it could break your workflows or cause unexpected behavior when the action owner publishes an update\\.\n\n\\- Using the commit SHA of a released action version is the safest for stability and security\\.\n\n\\- Using the specific major action version allows you to receive critical fixes and security patches while still maintaining compatibility\\. It also assures that your workflow should still work\\.\n\n\\- Using the master branch of an action may be convenient, but if someone releases a new major version with a breaking change, your workflow could break\\.\n\nSome actions require inputs that you must set using the with keyword\\. Review the action's README file to determine the inputs required\\.\n\nActions are either JavaScript files or Docker containers\\. If the action you're using is a Docker container you must run the job in a Linux virtual environment\\. For more details, see https://help\\.github\\.com/en/articles/virtual\\-environments\\-for\\-github\\-actions\\.\n\nSource: [github-workflow.json](https://www.schemastore.org/github-workflow.json)"}, "range": {"start": {"line": 17, "character": 14}, "end": {"line": 17, "character": 37}}}
```

Schema is correct: https://www.schemastore.org/github-workflow.json

## Steps to Reproduce

Here is sample github workflow for test (for example hover on checkout action on the last line):

```yaml
name: Daily

on:
  schedule:
    # Runs daily at 10:15 CET (09:15 UTC)
    - cron: '15 9 * * *'
  workflow_dispatch:  # Allow manual triggering

jobs:
  update-data:
    runs-on: ubuntu-latest
    
    steps:
    - name: Checkout repository
      uses: actions/checkout@v4
```

Initially wrongly issued here: https://github.com/sublimelsp/LSP-yaml/issues/81

rchl:
With 1.20.0 in ST I see the following raw markdown content on hovering `uses`:


```
#### GitHub Workflow

[...]

PR #1159, opened 2025-12-18
Fix hover markdown escaping so links are usable

shin19991207:
### What does this PR do?

Hover content is rendered as Markdown, so plain-text schema descriptions must be escaped for Markdown. The current escaping escapes punctuation like `.` and `()`, which can, for example, change a URL at the end of a sentence into `https://hub\.docker\.com/\.`. When the text is being auto-linkified by the hover renderer, it then include the backslash in the auto-detected link, making it end with \ and become unusable (for example, `https://hub.docker.com/\`).

This is fixed by updating YamlHover.toMarkdown() to stop escaping punctuation commonly used with URLs i.e. `.` `(` `)`.

**Note**: Best practice is for schemas to wrap URLs in `<...>` (for example, `<https://example.com>`), so the renderer treats them as explicit autolinks and they reliably stay usable.

- https://github.com/redhat-developer/yaml-language-server/issues/1151

### Is it tested? How?

Automated test:
- Added hover tests to ensure `()` parentheses and `.` dots remain literal even without escaping in descriptions.
- Added hover tests to ensure URLs next to () or . stay usable.

Manual test:
- Create a file `.github/workflows/daily.yaml`
- Add content:
  ```yaml
  name: Daily
  
  on:
    schedule:
      # Runs daily at 10:15 CET (09:15 UTC)
      - cron: '15 9 * * *'
    workflow_dispatch:  # Allow manual triggering
  
  jobs:
    update-data:
      runs-on: ubuntu-latest
      
      steps:
      - name: Checkout repository
        uses: actions/checkout@v4
  ```
- Hover over to the text `actions/checkout@v4`. Links in there should be usable.

### DIFF
--- a/test/hover.test.ts
+++ b/test/hover.test.ts
@@ -906,21 +906,21 @@ Source: [${SCHEMA_ID}](file:///${SCHEMA_ID})`
       schemaProvider.addSchema(SCHEMA_ID, {
         type: 'object',
         properties: {
-          registry: {
+          repo: {
             type: 'string',
             description: 'The YAML Language Server at https://github.com/redhat-developer/yaml-language-server',
           },
         },
       });
 
-      const content = 'regi|s|try: x';
+      const content = 'rep|o|: x';
       const hover = await parseSetup(content);
 
       assert.strictEqual(MarkupContent.is(hover.contents), true);
       assert.strictEqual((hover.contents as MarkupContent).kind, 'markdown');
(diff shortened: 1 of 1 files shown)


datho7561 (approved):
Looks good and addresses the bug. It's pretty safe to not escape these special characters, I think they're used exclusively for links and ordered lists. Thanks, Morgan!

PR #1160, opened 2025-12-18
[WIP] use ajv to validate YAML files

datho7561:
### What does this PR do?
Validate YAML against its schema with ajv instead of our validator.

This is a proof of concept. Many existing features are broken slightly, especially those related to validation specific to the VS Code extensions to JSON Schema. ajv provides human readable error messages and JSON Pointers to the invalid data out of the box, which makes it suitable to use to validate YAML, and slightly easier to use than `@hyperjump/json-schema`.

https://github.com/redhat-developer/yaml-language-server/issues/856

### Is it tested? How?
Existing tests (they are failing)

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -6,7 +6,7 @@
     {
       "label": "watch typescript",
       "type": "shell",
-      "command": "yarn run watch",
+      "command": "npm run watch",
       "presentation": {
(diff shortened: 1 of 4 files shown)


datho7561:
See https://github.com/redhat-developer/yaml-language-server/pull/1158#issuecomment-3725982111

PR #1165, opened 2026-01-15
Fix loading translations when running on web

datho7561:
### What does this PR do?
1. Move the calls to read to the translations from the filesystem to the node-specific code. This prevents the crash when attempting to start the extension using a web worker, since the node-specific methods aren't used.
2. When running as a web worker, receive the translations through the message passing mechanism, then configure the translations before starting the language server. This is the mechanism that VS Code's JSON extension uses to configure translations when running as a web worker.
3. Use the English translation instead of an id in the calls to `l10n.t()`. VS Code's tools to load translations don't load translations if the language is English. This could also be helpful if loading the English translations fails.

Fixes https://github.com/redhat-developer/vscode-yaml/issues/1191

### Is it tested? How?
Manually. I'm hoping to write some smoke tests for running as a web worker in vscode-yaml.

### DIFF
--- a/src/languageservice/services/validation/yaml-style.ts
+++ b/src/languageservice/services/validation/yaml-style.ts
@@ -22,7 +22,7 @@ export class YAMLStyleValidator implements AdditionalValidator {
         result.push(
           Diagnostic.create(
             this.getRangeOf(document, node.srcToken),
-            l10n.t('Flow style mapping is forbidden', 'Flow style mapping is forbidden'),
+            l10n.t('Flow style mapping is forbidden'),
             DiagnosticSeverity.Error,
             'flowMap'
           )


src/languageservice/services/yamlCodeActions.ts
```diff
@@ -233,7 +233,7 @@ export class YamlCodeActions {
           const rewriter = new FlowStyleRewriter(this.indentation);
           results.push(
             CodeAction.create(
-              l10n.t('convertToBlockStyle', 'Convert to block style {0}', blockTypeDescription),
+              l10n.t('Convert to block style {0}', 'Convert to block style {0}', blockTypeDescription),
```
shin19991207:
```suggestion
              l10n.t('Convert to block style {0}', blockTypeDescription),
```

shin19991207 (changes requested):
I reviewed the code and the implementation makes sense to me overall. I noticed a few minor typos/redundancies, plus one question about how the worker bootstraps.

datho7561:
To test the yaml language PR in vscode-yaml on web, you'll need https://github.com/redhat-developer/vscode-yaml/pull/1193/files (to send the message with the translations to the web worker). Then, to run it on web, there's `npm run run-in-chromium`

shin19991207 (approved):
I tested this manually and it works good. Thanks, David!

PR #1169, opened 2026-01-31
fix unset schemaStore.url not using the default

guille:
### What does this PR do?

Prior to this commit, enabling schema store (via `schemaStore.enable`) required also setting `schemaStore.url`, either to an empty string or to a valid URL.

When the setting was unset, the value of the url was undefined, and the checks were not properly catching the case.

This commit makes it so an unset `schemaStore.url` acts the same as an empty string and uses the default URL.

See above.

### Is it tested? How?

I ran `npm run compile`, pointed my LSP client to `node $path_to_project/out/server/src/server.js --stdio` and commented out `yaml.schemaStore.url` and opened a yml file. It was able to fetch its schema.

Unit tests are also passing.

### DIFF
--- a/src/languageserver/handlers/settingsHandlers.ts
+++ b/src/languageserver/handlers/settingsHandlers.ts
@@ -83,7 +83,7 @@ export class SettingsHandler {
 
       if (settings.yaml.schemaStore) {
         this.yamlSettings.schemaStoreEnabled = settings.yaml.schemaStore.enable;
-        if (settings.yaml.schemaStore.url?.length !== 0) {
+        if (settings.yaml.schemaStore.url) {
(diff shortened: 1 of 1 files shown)


datho7561 (approved):
To test, I used vscode-yaml. I removed the setting from the package.json so that VS Code would send nothing to the server (instead of the default URL or the empty string). Before this change, this breaks the schema store support. With this change, it works fine.

datho7561:
@guille do you mind rebasing this PR? For some reason, GitHub won't let me merge until this is rebased.

guille:
Done @datho7561 , thanks for reviewing & testing

PR #1173, opened 2026-02-05
Fix resolving schemas on web

datho7561:
### What does this PR do?
Use URI instead of filepath to retrieve schemas from workspace on web, since the web environment sometimes uses URIs with schemes other than `file://` for the workspace files

Fixes https://github.com/redhat-developer/vscode-yaml/issues/1194

### Is it tested? How?
Manually, integration tests/smoke tests in vscode-yaml


shin19991207 (approved):
LGTM! I tested manually and it works good. Thanks, David!

PR #1166, opened 2026-01-16
Support JSON Schema 2019 and 2020

shin19991207:
#### Refactor / architecture
- Extract instance validation into a validator layer (base + draft-specific validators)
- Keep backwards compatibility where reasonable (draft-07 behavior remains the baseline)

#### [Draft 2019-09 support](https://json-schema.org/draft/2019-09)

keyword/item | implementation | tests | note
-- | -- | -- | --
`$anchor` resolution | ✅ | ✅ from [JSON Schema Test Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests) | Draft-07 and earlier should resolve plain‑name fragments defined by $id, see **Additional fixes<sup>2</sup>**
`$defs` resolution (renamed + `definitions` compat) | -  | - | already covered (it resolved to the same way as `definitions`)
`$id` | -  | - | covered by meta-validation[^1]
`$recursiveAnchor` / `$recursiveRef` | ✅ | ✅ from [JSON Schema Test Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests)
`$ref` resolution updated to support sibling keywords | ✅ | ✅ from [JSON Schema Test Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests) | Draft-07 and earlier should ignore siblings, see **Additional fixes<sup>1</sup>**
`$vocabulary` | -  | - | covered by meta-validation[^1]
`dependentSchemas` | ✅ | ✅ from [JSON Schema Test Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests)
`unevaluatedItems` | ✅ | ✅ from [JSON Schema Test Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests)
`unevaluatedProperties` | ✅ | ✅ from [JSON Schema Test Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests)
`dependentRequired` | ✅ | ✅ from [JSON Schema Test Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests)
`contains` + `minContains` / `maxContains` | ✅ | ✅ from [JSON Schema Test Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests)

#### [Draft 2020-12 support](https://json-schema.org/draft/2020-12)

keyword/item | implementation | tests | note
[...]

PR #1176, opened 2026-02-10
fix: pass loadSchema callback to pickSchemaDialect

elasticjava:
## Summary

`pickSchemaDialect` is a module-level function but references `this.loadSchema(s)` (line 1338), where `this` is `undefined` in strict mode. This causes a `TypeError` when resolving schemas with custom (non-standard) `$schema` URIs that require loading the meta-schema to infer the base dialect.

**Root cause:** The function was introduced as part of #1166 (Support JSON Schema 2019 and 2020). It lives at module scope but uses `this` as if it were a class method.

**Fix:** Add an optional `loadSchema` callback parameter. The caller in `resolveSchemaContent` passes `this.loadSchema.bind(this)` through a closure variable. When `loadSchema` is not provided, the function returns `undefined` for unknown dialects instead of crashing.

**Changed lines:** +9/-3 in `yamlSchemaService.ts`

## Reproduction

Any schema with a custom `$schema` URI that is not one of the well-known drafts triggers the bug:

```json
{
  "$schema": "https://example.com/my-custom-meta-schema/v1",
  "type": "object",
  "properties": { "name": { "type": "string" } }
}
```

Also affects schemas embedding K8s definitions with `"$schema": "http://json-schema.org/schema#"` in nested `$defs`.

## Test plan

- [x] New test `pickSchemaDialect.test.ts` with 4 test cases:
  - Custom `$schema` URI does not throw TypeError
  - Known draft URIs still work (draft-07, 2020-12)
  - Nested `$schema` on sub-definitions does not crash
- [x] All 1229 existing + new tests pass (0 failing, 5 pending — all pre-existing Windows-only skips)

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,5 +1,5 @@
 {
-  "timeout": "5000",
+  "timeout": "10000",
   "ui":  "bdd",
   "extension": ["ts"],
   "package": "./package.json",


datho7561:
Confirmed this is a problem. There's a build error caused by formatting that prettier doesn't like. I'll fix that locally and make sure that this fix works.

datho7561:
Once the build error is fixed, I think this'll be good to merge.

datho7561:
The tests seem to pass even without the fix; I'll try to figure out what's going on.

datho7561:
> I'll try to figure out what's going on.

That turned into a day long sidequest

elasticjava:
> I've modified the test case so that it fails without the fix, and left just the case where we check the broken behaviour.
> 
> @elasticjava Sorry for modifying your PR so much. I'd love another look at my work if you have time. If not, I'll go ahead and merge.

Thank you @datho7561 — really appreciate the improvements.

The new `invalid-metaschema.test.ts` is spot on: 
much closer to the real-life flow where this actually breaks.

Also great catch on the missing `await` — 
I guess I can’t *await* to validate my schema now. 🙂

Thanks for merging!

PR #1177, opened 2026-02-10
Fix range of 'key ordering' error

### LINKED ISSUE #1172: Weird range on key ordering errors
## Describe the bug
The range of key ordering errors is weird. Sometimes it starts before the out of order key, sometimes it ends after the out of order key.

Here is an example YAML that shows this behaviour:

```yaml
one:
  child: moo
two:
  mild: meh
  child: moo
three:
  child: moo
```

## Expected Behavior
There are two errors, one that covers `two:` and one that covers `mild:`.

## Current Behavior
There are two errors, one that starts at `two:` and continues to the next line for two characters, and one that starts at `mild:`  but continues past the `:` by one character.

## Steps to Reproduce
1. Enable key ordering errors, eg. `"yaml.keyOrdering": true,`
2. Make the above YAML
3. Observe the wonky error ranges

## Environment
- [x] Windows
- [x] Mac
- [x] Linux
- [ ] other (please specify)

datho7561:
### What does this PR do?
Instead of reporting the range of the key and value, just report the range of the key. This helps to reduce visual clutter in the IDE due to the errors.

Fixes #1172

### Is it tested? How?
Existing and new unit tests

### DIFF
--- a/src/languageservice/services/validation/map-key-order.ts
+++ b/src/languageservice/services/validation/map-key-order.ts
@@ -27,6 +27,7 @@ export class MapKeyOrderValidator implements AdditionalValidator {
                 'mapKeyOrder'
               )
             );
+            break;
           }
         }
       }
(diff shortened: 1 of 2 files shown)


datho7561:
Yes, that makes sense to me. Thanks for the review, Morgan!

PR #1186, opened 2026-02-19
Prefer local sibling schema resolution for relative $ref before remote $id lookup

### LINKED ISSUE #1184: Support resolution of local schemas
With the new 1.20.0 release a lot of the incorrect resolution behavior seems to be fixed.
The issue I see with the new resolution is that schemas defined in local files are not resolved correctly.

Assume you have the following two schemas defined in a local directory:

_schemas/primary.json_
```json
{
  "$id": "https://example.com/schemas/primary.json",
  "$ref": "secondary.json"
}
```
_schemas/secondary.json_
```json
{
  "$id": "https://example.com/schemas/secondary.json",
  "type": "string",
}
```

If you now map a yaml file to the primary.json schema file, resolution for the `$ref` will fail as it tries to find the schema `https://example.com/schemas/secondary.json`.
The resolution would work if the secondary.json file was actually available at the specified location but the specification does not require that.

To allow local development of schemas that reference other local schemas it would be helpful to add a configuration option that specifies schema files to load and register internally.

Something like:
```json
{
  "yaml.localSchemas": [
    "./schemas/*.json"
  ]
}
```

The loaded schemas should precede any attempt of loading schemas via http.

Bundling the schemas according to https://json-schema.org/draft/2020-12/json-schema-core#name-compound-documents works but is impractical during development.

### Additional context
I am willing to contribute here if someone can give me some guidance/input.

shin19991207:
### What does this PR do?
Resolves relative `$ref` using local sibling schema path before remote `$id` ref

Fixes https://github.com/redhat-developer/yaml-language-server/issues/1184

### Is it tested? How?

Manually and with automated tests.
```json
{
  "$id": "https://example.com/schemas/primary.json",
  "type": "object",
  "properties": {
    "mode": {
      "$ref": "secondary.json"
    }
  },
  "required": ["mode"],
  "additionalProperties": false
}
```
```json
{
  "$id": "https://example.com/schemas/secondary.json",
  "type": "string",
  "enum": ["dev", "prod"]
}
```
```yaml
# yaml-language-server: $schema=./primary.json
mode: stage
# expect 'Value is not accepted. Valid values: "dev", "prod".' instead of 'Unable to load ...'
```

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -131,6 +131,7 @@ export class YAMLSchemaService extends JSONSchemaService {
   public schemaPriorityMapping: Map<string, Set<SchemaPriority>>;
 
   private schemaUriToNameAndDescription = new Map<string, SchemaStoreSchema>();
+  private schemaIdPrefixToLocalDir = new Map<string, string>();
 
   constructor(
     requestService: SchemaRequestService,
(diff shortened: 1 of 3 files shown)


datho7561:
It's working okay, but it seems like the schema is cached somehow, so that if the second schema is changed, the changes aren't loaded properly until you close and reopen VS Code. Caching makes sense if it's a remote schema, but not really if it's local.

datho7561:
I think it'd be helpful to add a test with an absolute reference as well, eg.

`primary.json`

```json
{
  "$id": "https://example.com/schemas/primary.json",
  "$ref": "/schemas/secondary.json"
}
```

`secondary.json` (in the same folder as primary)

```json
{
  "$schema": "https://json-schema.org/draft/2019-09/schema",
  "$id": "https://example.com/schemas/secondary.json",
  "type": "object",
  "properties": {
    "name": { "type": "string" },
    "age": { "type": "integer", "minimum": 0 }
  },
  "required": ["name", "age"]
}
```

DorianCzichotzki:
Will this break bundled schemas? Or is that resolution done even earlier.

A bundled version of the example would be:
```json
{
  "$id": "https://example.com/schemas/primary.json",
  "type": "object",
  "properties": {
    "mode": {
      "$ref": "secondary.json"
    }
  },
  "required": ["mode"],
  "$defs": {
    "https://example.com/schemas/secondary.json": {
      "$id": "https://example.com/schemas/secondary.json",
      "type": "string",
      "enum": ["dev", "prod"]
    }
  }
}
```

In this case the embedded schema should be used.

shin19991207:
@datho7561 I revisited the 'local schema changes not reloaded' issue with the code and thought of a better approach. When resolving `$id`, try a local sibling file first: `_preferLocalBaseForRemoteId()` builds a file URI from the current schema’s directory + the `$id` filename, uses it if it can be loaded; otherwise falls back to `$id` resolution.

I added test cases for reloading local schema after changes & absolute reference as well.

@DorianCzichotzki Bundled schemas should not be affected. They are resolved before any external fetch. In yamlSchemaService, we index embedded resources first then resolve $ref against that index.

datho7561:
It seems like the example in https://github.com/redhat-developer/yaml-language-server/pull/1186#issuecomment-3928387637 doesn't work properly from my testing.

shin19991207:
Okay. I reviewed and updated how schema resolution should work when attempting to load schemas from a local file before falling back to remote URIs. It comes in two cases:

**1. Relative reference**

Given:
_/some/path/schemas/primary.json_
```
{
  "$id": "https://example.com/schemas/primary.json",
  "$ref": "secondary.json"
}
```

_/some/path/schemas/secondary.json_
```
{
  "$id": "https://example.com/schemas/secondary.json",
  "type": "string"
}
```

YAML file:
```
# yaml-language-server: $schema=/some/path/schemas/primary.json
...
```

Here, `"$ref": "secondary.json"` first resolve relative to the retrieval URI of `primary.json` (`/some/path/schemas/primary.json`), producing `/some/path/schemas/secondary.json`.

If that local file is missing, the next attempt is against the $id field (`https://example.com/schemas/primary.json`) as https://example.com/schemas/secondary.json.

**2. Absolute reference**

Given:
_/some/path/schemas/primary.json_
```
{
  "$id": "https://example.com/schemas/primary.json",
  "$ref": "/schemas/secondary.json"
}
```

_/some/path/schemas/secondary.json_
```
{
  "$schema": "https://json-schema.org/draft/2019-09/schema",
  "$id": "https://example.com/schemas/secondary.json",
  "type": "object",
  "properties": {
    "name": { "type": "string" },
    "age": { "type": "integer", "minimum": 0 }
  },
  "required": ["name", "age"]
}
```

Under standard JSON Schema rules, an absolute JSON Pointer-like path resolves against the base URI, not the local filesystem. Since `primary.json` has `$id = https://example.com/schemas/primary.json`, resolving the absolute reference `"/schemas/secondary.json"` against that produces https://example.com/schemas/secondary.json.

From there, the relative path `secondary.json` is tried locally to `/some/path/schemas/secondary.json` as first attempt to resolve (this avoids interpreting `/schemas/secondary.json` as a filesystem path such as `file:///schemas/secondary.json`). 

Second attempt falls back to https://example.com/schemas/secondary.json.

datho7561 (approved):
Seems to be working as described in the spec now. Thanks, Morgan!

PR #1188, opened 2026-02-20
Allow language-overridable `yaml.validate` and `yaml.format.enable` settings

shin19991207:
### What does this PR do?

Allow language-overridable `yaml.validate` and `yaml.format.enable` settings

https://github.com/redhat-developer/vscode-yaml/issues/1204#issuecomment-3930761456

### Is it tested? How?

Manually and through automated tests


datho7561 (changes requested):
Works well! I didn't understand at first why we needed changes in yaml-language-server, but now I do. I think it's probably a good idea to use sinon for stubbing functions.

datho7561 (approved):
Looks good! Thanks, Morgan!

PR #1189, opened 2026-02-21
Add yaml-lint-disable comment to suppress diagnostics per-line

X-Guardian:
### What does this PR do?

Add support for suppressing linter diagnostics on a per-line basis using a `yaml-lint-disable` comment placed on the line immediately before the one producing the diagnostic.

The comment supports three forms: suppress all diagnostics, suppress by message substring, or suppress by multiple comma-separated substrings (case-insensitive). Implemented via LSP client middleware that intercepts diagnostics from the yaml-language-server and filters out suppressed ones before they reach VS Code.

- Resolves https://github.com/redhat-developer/vscode-yaml/issues/666

### Is it tested? How?

- Ran unit tests: `npm test`
- In VSCode, F5 launch the extension, open a YAML file with schema validation errors, add `# yaml-lint-disable` above a line with errors and verified they were suppressed
- Verified `# yaml-lint-disable <substring>` only suppresses matching diagnostics
- Verified diagnostics reappear when the comment is removed
- Verified no impact on files without disable comments

### DIFF
--- a/test/diagnostic-filter.test.ts
+++ b/test/diagnostic-filter.test.ts
@@ -184,7 +184,13 @@ describe('filterSuppressedDiagnostics', () => {
   });
 
   it('should handle multiple disable comments for different lines', () => {
-    const lines = linesOf(['# yaml-language-server-disable', 'line1: bad', 'line2: ok', '# yaml-language-server-disable', 'line4: also-bad']);
+    const lines = linesOf([
+      '# yaml-language-server-disable',
+      'line1: bad',
+      'line2: ok',
+      '# yaml-language-server-disable',
+      'line4: also-bad',
+    ]);
     const diagnostics = [makeDiag(1, 'error on line 1'), makeDiag(2, 'error on line 2'), makeDiag(4, 'error on line 4')];
 
     const result = filter(diagnostics, lines);


datho7561:
I'm iffy about using `yaml-lint-disable`, since `yamllint` is an existing python tool that's unrelated to this extension. Maybe `yaml-language-server-disable` instead?

The strategy of using the text of the diagnostic to specify which error to silence is a bit messy. We provide translations for a few different locales, so the text in the comment will need to match the text in the translation that's currently loaded.

In the long term, it might be a good idea to [use the LSP `code` field](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic) to specify an error code for each diagnostic that gives the general "class" of error (eg. `type`, `parse`, `key-order`, etc.) then silence using that code (eg. `# yaml-lint-disable type` to silence a type error). That would take a decent amount of work to implement in the first place though.

For now, I think using the error text is a good solution.

X-Guardian:
Done @datho7561.

datho7561:
Seems like the build's failing due to a formatting-related lint problem

X-Guardian:
Formatting fixed.

X-Guardian:
Hi @datho7561, looks like the MacOS tests are flaky. Can you re-trigger the workflow?

datho7561:
coveralls.io DNS is down, so Ubuntu is failing because the coverage report can't be uploaded

datho7561:
(there's some sort of rule where I can't merge until CI is passing, hoping coveralls comes back up soon)

PR #1212, opened 2026-03-12
Remove annoying log output related to modelines

datho7561:
### What does this PR do?
Currently, a log entry is emitted whenever a modeline comment is parsed. This is because, in https://github.com/redhat-developer/yaml-language-server/pull/1193, the logic for how modelines were parsed was changed. On this line: https://github.com/redhat-developer/yaml-language-server/pull/1193/changes#diff-dc843f2ff5a3d197853464e616d6e413820becbc82e0e9a3aa6fc6d996976bb2L21 there is always 2 entries if a parsable modeline is present. This is because the index 0 entry contains the entire match, and the index 1 entry contains the group match (that contains the URI of the schema). This means that the `console.log()` will always run, incorrectly alerting the user that they tried to specify two schemas in the modeline.

Even if the message were accurate, I don't think logging is the best way to alert the user to this problem.
This should ideally be a warning or error instead. For now, I'm removing the `console.log()` entirely to address this.

Fixes #1211

### Is it tested? How?
Manually

### DIFF
--- a/src/languageservice/services/modelineUtil.ts
+++ b/src/languageservice/services/modelineUtil.ts
@@ -17,14 +17,9 @@ export function getSchemaFromModeline(doc: SingleYAMLDocument | JSONDocument): s
       return isModeline(lineComment);
     });
     if (yamlLanguageServerModeline != undefined) {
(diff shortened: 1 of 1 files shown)


shin19991207 (approved):
LGTM and fixes the issue. Thanks, David!

PR #1214, opened 2026-03-13
Update Kubernetes schema to 1.33

datho7561:
### What does this PR do?
The oldest supported Kubernetes version is 1.33, so I've updated the referenced schema to 1.33.

### Is it tested? How?
uhhhh


datho7561:
Morgan's PR upgraded to 1.34 and allowed changing the version used via a setting

PR #1222, opened 2026-03-27
Use yaml package for formatting

datho7561:
### What does this PR do?
Rebases #983

Original text from that issue:

This replaces Prettier as the LSP formatting solution with the `yaml` package. The `yaml` package can format YAML just fine, and we already have it as a dependency. Prettier is a big dependency.

If people want to use Prettier, other Prettier integrations are probably better suited for them. For example, the YAML language server doesn’t respect Prettier configuration files.

This is a proof of concept. I tried to map existing options to the new implementation. A more ideal solution might be to change the formatting options.

Refs #933 , Rebases #983

### Is it tested? How?
Existing tests


shin19991207 (approved):
Looks good to me. Thanks, David!

datho7561:
Thanks for the review!

PR #1223, opened 2026-04-06
Base k8s schema resolution off of GroupVersionKind

datho7561:
### What does this PR do?
This should improve validation and completion of Kubernetes resources drastically.

Fixes #1213

### Is it tested? How?
- [x] Existing unit tests
- [x] New unit tests


datho7561:
I looked into removing the custom validation and completion logic for files marked as "kubernetes", but that caused problems in files that don't have a GVK declared yet. We'll need a more robust solution some time in the future for this pain point; I don't think it's worth modifying the completion/validation behaviour just to get this one schema working as expected (it's a hack).

It might be interesting to have a generic "kubernetes object" schema to fall back on that includes the bare minimum (completion/validation for GVK and some metadata) so that we don't have to use the large "every kubernetes object" schema.

shin19991207 (approved):
Looks good and works well for me. Thanks, David!

PR #1230, opened 2026-04-14
Fix local sibling `$ref` resolution when `$id` does not end with file name

### LINKED ISSUE #1228: Issue with local refs in local schema when ID does not end with file name
## Describe the bug
When a local JSON schema file has an ID that does not end with the file name, it is unable to resolve any local references to other local schemas. 

This is a new issue since version 1.20.0, is not fixed in 1.20.1 and I think is similar to #1184.

## Expected Behavior
`$ref: "./repro_defs.json" `should resolve relative to the local schema file, find the other local schema, and provide validation.

## Current Behavior
The extension resolves ./repro_defs.json against the schema's $id (https://example.com/schemas/repro-main-v1), producing https://example.com/schemas/repro_defs.json, then fails with:

`Problems loading reference 'https://example.com/schemas/repro_defs.json': Unable to load schema from 'https://example.com/schemas/repro_defs.json': No content.`

## Steps to Reproduce
Create repro_defs.json:
```
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://example.com/schemas/repro-defs-v1",
  "title": "Shared Defs",
      "type": "object",
      "properties": {
        "name": { "type": "string" },
        "age": { "type": "integer" }
      },
      "required": ["name"]
}
```

Create repro_main_schema.json in the same directory:
```
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://example.com/schemas/repro-main-v1",
  "title": "Main Schema",
  "type": "object",
  "properties": {
    "team": { "type": "string" },
    "members": {
      "type": "array",
      "items": {
        "$ref": "./repro_defs.json#/$defs/Person"
      }
    }
  },
  "required": ["team", "members"]
}
```
Create repro_test.yaml in the same directory:
```
# yaml-language-server: $schema=/absolute/path/to/repro_main_schema.json
team: platform
members:
  - name: Alice
    age: 30
```
Open repro_test.yaml in VS Code — error appears on team.

This however is fixed, if you replace the IDs of the primary schema with `https://example.com/schemas/repro_main_schema.json`. There should be no requirement for the ID of the schema to match the local filename based on how I understand the spec. 

## Environment
- [ ] Windows
- [x] Mac
- [ ] Linux
- [ ] other (please specify)

shin19991207:
### What does this PR do?
Implemented `fallbackBaseURL` so local schema like `file:///.../repro_main_schema.json` can still resolve `$ref: ./repro_defs.json` from the local directory even if its top-level `$id` is something unrelated like `https://example.com/schemas/repro-main-v1`.

`fallbackBaseURL` represents the actual schema resource we loaded from disk and is used for local-sibling fallback.

Fixes https://github.com/redhat-developer/yaml-language-server/issues/1228

### Is it tested? How?

- [x] New automated tests
- [x] Test manually

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -534,6 +534,7 @@ export class YAMLSchemaService extends JSONSchemaService {
       uri: string,
       linkPath: string,
       parentSchemaURL: string,
+      fallbackBaseURL: string,
       parentSchemaDependencies: SchemaDependencies,
       resolutionStack: Set<string>,
       recursiveAnchorBase: string,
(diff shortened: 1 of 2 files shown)


datho7561 (approved):
Looks good and fixes the issue. Thanks, Morgan!

PR #1238, opened 2026-04-23
Fix schema resolution when `$id` base URI differs from the loaded schema URI

### LINKED ISSUE #1236: Bundled schema resolution fails for schemas with $id matching a file name
## Describe the bug
I was testing with bundled schemas and encountered an odd error that took me a while to figure out. The issue is that schema resolution seems to fail when the $id of the schema ends in an existing schema file name. I am not sure if this happens only for bundled schemas, but that is where I found the issue.

Take the following local schema defined in a file called `test_schema.json`:
```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://example.com/test_schema.json",
  "$defs": {
    "https://example.com/subschema/schema.json": {
      "$schema": "https://json-schema.org/draft/2020-12/schema",
      "$id": "https://example.com/subschema/schema.json",
      "$defs": {
        "sub_options": {
          "type": "string"
        }
      },
      "type": "object",
      "properties": {
        "suboption": {
          "$ref": "#/$defs/sub_options"
        }
      }
    }
  },
  "properties": {
    "option": { "$ref": "subschema/schema.json" }
  }
}
```

Note that the $id of the schema ends in `test_schema.json`. No matter what file I try to validate with this schema it returns the following error:

```
$ref '/$defs/sub_options' in 'https://example.com/subschema/schema.json' cannot be resolved.
```

If I now change the $id of the root schema to anything else the resolution works and I can validate the YAML file correctly. The only change is the $id of the schema:
```yaml
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://example.com/other_schema.json",
  "$defs": {
    "https://example.com/subschema/schema.json": {
      "$schema": "https://json-schema.org/draft/2020-12/schema",
      "$id": "https://example.com/subschema/schema.json",
      "$defs": {
        "sub_options": {
          "type": "string"
        }
      },
      "type": "object",
      "properties": {
        "suboption": {
          "$ref": "#/$defs/sub_options"
        }
      }
    }
  },
  "properties": {
    "option": { "$ref": "subschema/schema.json" }
  }
}
```

The YAML file I tried to validate is this:
```yaml
option:
  suboption: "some text"
```

I guess this has to do with the new file/bundled schema resolution introduced a few weeks ago.

## Environment
- [ ] Windows
- [x] Mac
- [ ] Linux
- [ ] other (please specify)

shin19991207:
### What does this PR do?

The regression was introduced in https://github.com/redhat-developer/yaml-language-server/pull/1186, where the added `_preferLocalBaseForRemoteId` sets a schema node’s baseUri to the local file path when a schema is loaded from a local file whose root `$id` ends with the same filename. In this case, the resolver effectively treated the local file URI as the base for resolving relative references.
This behavior caused problems for embedded resource resolution. Because the resolver used the local file URI as the effective base, relative references could resolve against the wrong location. In particular, embedded resource references failed because `resourceIndexByUri` stores embedded resources under their original `$id`, not under a locally rewritten URI. As a result, internal references such as `#/$defs/...` could not find a corresponding item in `resourceIndexByUri` and thus could no longer be resolved correctly.

The fix separates the schema’s logical resolution base from the source URI it was loaded from. This removes the behavior that preferred a same-named local file as the effective base for a remote `$id`. Instead, the resolver now keeps these concerns distinct and only uses the source URI as a fallback when resolving external references to local sibling files. Now, we have:
- `_baseUri`: the logical resource URI derived from the schema’s `$id`
- `_sourceUri`: the URI of the file the schema was loaded from
This separation ensures that logical resolution remains consistent with the schema’s `$id`, while still allowing the source URI to be used for local file resolution.

Fixes https://github.com/redhat-developer/yaml-language-server/issues/1236

### Is it tested? How?

- [x] Manual test from https://github.com/redhat-developer/yaml-language-server/issues/1236
- [x] Automated tests

### DIFF
--- a/src/languageservice/jsonSchema.ts
+++ b/src/languageservice/jsonSchema.ts
@@ -17,7 +17,8 @@ export enum SchemaDialect {
 export interface JSONSchema {
   // for internal use
   _dialect?: SchemaDialect;
-  _baseUrl?: string;
+  _baseUri?: string;
+  _sourceUri?: string;
   _$ref?: string;
 
   id?: string;
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -63,7 +63,8 @@ const REF_SIBLING_NONCONSTRAINT_KEYS = new Set([
   '$schema',
   '$id',
   'id',
-  '_baseUrl',
+  '_baseUri',
+  '_sourceUri',
   '_dialect',
   '$anchor',
   '$dynamicAnchor',
@@ -216,8 +217,13 @@ export class YAMLSchemaService extends JSONSchemaService {
       return new ResolvedSchema({}, resolveErrors);
     }
 
+    const _setSourceUri = (node: JSONSchema, sourceUri: string | undefined): void => {
+      if (!node || typeof node !== 'object' || !sourceUri) return;
+      node._sourceUri = sourceUri;
+    };
+
     const _cloneSchema = (
-      value: unknown,
+      value: JSONSchema,
       seen: Map<object, unknown>,
       stopCondition?: (val: unknown, seenSize: number) => unknown | undefined
     ): unknown => {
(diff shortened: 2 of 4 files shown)


datho7561 (approved):
Looks good and addresses the linked issue. Thanks, Morgan!

PR #1240, opened 2026-04-29
Revert "Use yaml package for formatting"

### LINKED ISSUE #1233: YAML formatter automatically convert long integers to floats
## Describe the bug
YAML formatter is automatically converting long integers to a float value.

Workaround: use string value (but it modifies the data type)

## Expected Behavior
Long integers shall be kept as is; the reader may use BigInt and be able to read the exact value.
Or at least there should be an option to disable this automatic conversion.
(my use case: long barcode numbers, using Code 128 format)

## Current Behavior
- 12345678901234567890 is converted to 12345678901234567000
- 12345678901234567890123 is converted to 1.2345678901234568e+22

## Steps to Reproduce
Create a YAML file, with "test: 12345678901234567890" and save

## Environment
- [ ] Windows
- [ ] Mac
- [x] Linux
- [ ] other (please specify)

### LINKED ISSUE #1234: `proseWrap: “preserve”` is ignored, and text is always auto wrapped since v1.22.0
## Describe the bug
`proseWrap: “preserve”` is ignored, and text is always auto wrapped since v1.22.0.
If you revert to version 1.21.0 or earlier, it will work as expected.

Although `preserve` should be the default value for `proseWrap`, it wasn't applied, and it didn't work even when I set it explicitly.
If the text exceeds the default value of 80 for `printWidth`, a line break is forced.

## Environment
- [ ] Windows
- [x] Mac
- [ ] Linux
- [ ] other (please specify)

### LINKED ISSUE #1235: YAML formatter transforms escaped newlines in double-quoted strings
### Description

The YAML formatter automatically transforms escaped newline characters (`\n`) in double-quoted strings into actual multi-line formatting, which breaks the intended syntax for certain use cases.

### Steps to Reproduce

1. Create a YAML(Ansible) file with the following content:
```yaml
- name: Example task
  set_fact:
    my_var: "{{ content | regex_replace('\\\\n', '\n') }}"
```
OR
```yaml
- name: Example task
  set_fact:
    my_var: "{{ content | regex_replace('\\\\n', '\u000a') }}"
```

2. Save the file (with format on save enabled and YAML formatter enabled)
3. The formatter transforms it to:

```yaml
- name: Example task
  set_fact:
    my_var: "{{ content | regex_replace('\\\\n', '
    
      ') }}"
```

**Expected Behavior**

The escaped newline '\n' should be preserved as-is in double-quoted strings, since it has semantic meaning different from an actual line break.

**Actual Behavior**

The formatter interprets '\n' as a literal newline and reformats the string to span multiple lines, breaking the YAML syntax and changing the semantic meaning.

**Additional Context**

This also affects Unicode escape sequences like \u000a
Using # prettier-ignore or similar comments doesn't prevent the transformation

This is particularly problematic for Ansible playbooks where regex replacements need to convert literal \n strings into actual newlines

YAML spec allows escaped characters in double-quoted strings, and the formatter should preserve them
Environment

Version: 1.116.0 (Universal)
OS: Darwin arm64 25.4.0
vscode-yaml extension version: 1.22

**Workaround**
Disabling the YAML formatter.

**Suggested Fix**
The formatter should recognize that escaped characters in double-quoted strings have semantic meaning and should not be transformed into their literal representations during formatting.

### LINKED ISSUE #1237: The new formatter strips trailing zeroes from exponents, which are required for YAML 1.1 floats.
## Summary
tl;dr the fix is to set `"yaml.yamlVersion": "1.1"` but this was not previously necessary and may impact a large number of python + YAML users.

The current extension's YAML formatter will replace  `1.0e+4` with `1e+4`.

In YAML 1.1, `1e+4` is not a valid float.
In YAML 1.2, it is.

The -- to my knowledge -- standard and widely-used interface between python and YAML, PyYAML, is still on YAML 1.1
As a result, exponents are interpreted as strings if they lack a `.0`
```python
>>> import yaml
>>> yaml.safe_load("1.0e+4") 
10000.0
>>> yaml.safe_load("1e+4")  
'1e+4'
```

Adding more details to the formatter changelog entry ("The new formatter may perform breaking changes if `yaml.yamlVersion` is not set correctly, such as `1.0e+4` -> `1e+4` which are equivalent in YAML 1.2 but not YAML 1.1.") would help limit the fallout.

(PyYAML does not look likely to fix it on their side anytime soon https://github.com/yaml/pyyaml/issues/173)

datho7561:
### What does this PR do?
This reverts commit 22d01b401da90192846d4d3938360f288f687ee8.

There were many bugs in the new formatter implementation, so we're reverting for now.

In order to move forwards with this, we'll either need to adjust the behaviour of `toString` in the YAML parser through contributing patches, or write our own formatter.

- Fixes https://github.com/redhat-developer/vscode-yaml/issues/1237
- Fixes https://github.com/redhat-developer/vscode-yaml/issues/1235
- Fixes https://github.com/redhat-developer/vscode-yaml/issues/1234
- Fixes https://github.com/redhat-developer/vscode-yaml/issues/1233
- Fixes https://github.com/redhat-developer/yaml-language-server/issues/1237

### Is it tested? How?
Existing tests


holyspectral:
For an extra data point, `"yaml.format.singleQuote": true` option is also broken in [vscode-yaml](https://github.com/redhat-developer/vscode-yaml/) 1.22.0.  Things like this wouldn't get formatted to single quote.  1.21.0 works fine. 

```
  - it: "should include default selector"
```

PR #1241, opened 2026-04-29
Add tests for bugs in reverted formatter changes

datho7561:
### What does this PR do?
- Readd the fix that doesn't return any text edits if the formatter changes nothing
- Add a bunch of tests for the cases where the new formatter was failing

N/A

### Is it tested? How?
Unit tests of the formatter


shin19991207 (approved):
LGTM. Thanks, David!

PR #1251, opened 2026-05-12
Add setting to disable automatic schema detection for matching YAML files

### LINKED ISSUE #245: Disable schema detection for certain files
First of all, great extension. It is really useful to have yaml files validated (and auto-completed).

However, I have a file in my codebase called `some.thing-sam.yml` and I get red squiggle marks (linting errors) on valid properties. I believe what happens is that this extension treats the file as a AWS CloudFormation Serverless Application Model (SAM) file, which it isn't.

I have found a similar issue #98 where the schemaStore validation is disabled completely. But this is not what I want. Ideally I could provide a re-mapping of the file extension, as I believe the filename pattern is the culprit. Renaming the file to anything else but `*sam.yml` works.

Paraphrasing: Naming a yaml file `*sam.yml` will trick this extension (or the yaml-language-server) into treating the file as an AWS CloudFormation Serverless Application Model. There should be a way to fix this for the user.

I tried to provide a mapping in my VSCode settings myself, but without any luck:

```json
{
    "yaml.schemas": {
        "https://raw.githubusercontent.com/awslabs/goformation/master/schema/sam.schema.json": "foo.bar"
    }
}
```

Also noteworthy, the `catalog.json` has this configuration in place:

```json
    [...]
    {

      "name": "AWS CloudFormation Serverless Application Model (SAM)",

      "description": "The AWS Serverless Application Model (AWS SAM, previously known as Project Flourish) extends AWS CloudFormation to provide a simplified way of defining the Amazon API Gateway APIs, AWS Lambda functions, and Amazon DynamoDB tables needed by your serverless application.",

      "fileMatch": [

        "*.sam.json",

        "*.sam.yml",

        "*.sam.yaml",

        "sam.json",

        "sam.yml",

        "sam.yaml"

      ],

      "url": "https://raw.githubusercontent.com/awslabs/goformation/master/schema/sam.schema.json"

    },
    [...]
```

Judging from the glob patterns my file (`some.thing-sam.yml`) shouldn't have been picked up for linting with that schema in the first place, should it?

shin19991207:
### What does this PR do?

Adds the `yaml.disableSchemaDetection` setting for disabling automatically detected schemas for specific YAML files by glob pattern. The expected value is a glob pattern or list of glob patterns.

This is intended for cases where schemas from Schema Store or schema association notifications are incorrectly applied to files with generic names. Explicit schemas still take precedence, so users can continue to opt into validation with the `yaml.schemas` setting or a modeline.

Also update https://github.com/redhat-developer/yaml-language-server/pull/1251/changes#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R116 because absolute paths in `yaml.schemas` does work fine.

Fixes https://github.com/redhat-developer/vscode-yaml/issues/245

### Is it tested? How?

- [x] New automated tests
- [x] Manual testing. Tried with the setting:
   ```json
   "yaml.disableSchemaDetection": ["*sam*"],
   ```
   or
   ```json
   "yaml.disableSchemaDetection": "*sam*",
   ```
   and a file named `some.sam.yml` with following content:
    ```yaml
    name: example
    version: 1
    enabled: true
    items:
      - one
      - two
    ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ The following settings are supported:
 - `yaml.hover`: Enable/disable hover
 - `yaml.completion`: Enable/disable autocompletion
 - `yaml.schemas`: Helps you associate schemas with files in a glob pattern
-- `yaml.schemaDetectionDisableFor`: Disables automatically detected schemas for matching YAML files. Schemas explicitly configured with `yaml.schemas` or a modeline still apply.
+- `yaml.disableSchemaDetection`: Disables schema detection for matching YAML files. Modelines still apply.
 - `yaml.schemaStore.enable`: When set to true the YAML language server will pull in all available schemas from [JSON Schema Store](https://www.schemastore.org)
 - `yaml.schemaStore.url`: URL of a schema store catalog to use when downloading schemas.
 - `yaml.kubernetesCRDStore.enable`: When set to true the YAML language server will parse Kubernetes CRDs automatically and download them from the [CRD store](https://github.com/datreeio/CRDs-catalog).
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -294,14 +295,33 @@ or IntelliJ compatible format:
...
 ### Schema priority
 
 The following is the priority of schema association in highest to lowest priority:
 1. Modeline
 2. CustomSchemaProvider API
-3. yaml.settings
-4. Schema association notification
-5. Schema Store
+3. `yaml.schemas`
+4. `yaml.schemaDetectionDisableFor`
```
datho7561: These should be the other way around; if a schema is configured in `yaml.schemas` for a file and the file is also covered in `yaml.schemaDetectionDisableFor`, then the schema configured in `yaml.schemas` is ignored.

datho7561 (changes requested):
I think the name of the setting could be better.

Maybe `disableSchemaDetection`?

datho7561 (approved):
Looks good and works well. Thanks, Morgan!

We might need to comb through the issue list to find similar bugs and close them, I know this was a pain point for a bunch of users.

thernstig:
@shin19991207 @datho7561 would it not make sense to auto-disable it for common schemas in some way?

E.g. if the "Docker DX" extension is installed, it does not care about:

```json
      "compose*y*ml",
      "docker-compose*y*ml",
```

I know other extensions does similar things for this, and it is a good practice because then thousand developers do not need to manually update their settings for common files and extensions.

The new solution is good, but I do definitely think it can be more DX friendly.

datho7561:
That sounds like a good idea.

We'd need some mechanism to detect that the docker extension is installed. We'd probably also want GitHub Actions. I can't think of any others off the top of my head. We'd probably need to do this on the VS Code side, then, which means the mechanism wouldn't work for eg. neovim, helix, zed etc.. but I still think it's worth doing it.

thernstig:
@datho7561 I do know that https://github.com/hverlin/mise-vscode does discover if other extensions are installed, I have not investigated how or if what they do is idiomatic though.

PR #1252, opened 2026-05-14
Fix indent on enter for nested maps

datho7561:
### What does this PR do?
- Fixes indent on enter for nested maps
- Simplify code so that it adds or removes spaces based on what's already present instead of completely replacing it

Fixes redhat-developer/vscode-yaml#1226

### Is it tested? How?
Modified existing unit tests and added new ones for the nested cases

### DIFF
--- a/src/languageservice/services/yamlOnTypeFormatting.ts
+++ b/src/languageservice/services/yamlOnTypeFormatting.ts
@@ -4,8 +4,8 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { DocumentOnTypeFormattingParams } from 'vscode-languageserver';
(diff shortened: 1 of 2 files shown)


shin19991207 (approved):
I couldn't reproduce the issue on my end but the code looks good to me!

datho7561:
Wait, really? Hmm, I'll double check before and after on my Mac locally before merging.

datho7561:
I think maybe it's not on by default, I had to add `"editor.formatOnType": true`, then the bug appeared. Otherwise I wasn't getting any auto indentation, which seemed weird to me, since I thought that the TextMate grammar normally provides that.

PR #1255, opened 2026-05-19
Support return types for `yaml.customTags`

### LINKED ISSUE #867: CloudFormation ref in Fn:GetAtt and Fn:FindInMap shows validations errors Incorrect type. Expected "string". yaml-schema: aws://cloudformation.schema.json
## Describe the bug
I have "Incorrect type. Expected "string".yaml-schema: aws://cloudformation.schema.json" error, but aws is working normally with this template.
Screenshot 2023-01-06 at 11 26 41
Screenshot 2023-01-06 at 11 26 04

## Expected Behavior
Nothing shoud be highlighted, because it's working template

## Environment
- VsCode mac

shin19991207:
### What does this PR do?

This PR adds optional return types to the `yaml.customTags` setting.

Previously, a custom tag could only describe the YAML node type the tag is written on. For example, `"!Seq-example sequence"` tells the parser that `!Seq-example` would map to a sequence custom tag, but schema validation still treats the tagged value as a sequence.

This PR adds a new form `"!Tag nodeType:returnType"`, where `nodeType` specifies the YAML node type that the tag is written on, and `returnType` specifies the schema type that the tagged value evaluates to.

`yaml.customTags` now supports three forms:
```
!Tag
!Tag nodeType
!Tag nodeType:returnType
```

Existing configurations remain supported, so this is backward compatible. Behavior only changes when a user explicitly adds `:returnType`.

Fixes https://github.com/redhat-developer/vscode-yaml/issues/867

Note that `yaml.customTags` only applies to YAML short-form tags. It does not apply to CloudFormation long form such as `Fn::...`. Long form values are normal YAML mappings, not YAML tags.

### Is it tested? How?

- [x] New automated tests
- [x] Manual testing:

1. Add the following `yaml.customTags` configuration in workspace setting (a list of all intrinsic functions in short form with return types added, based on the [AWS CloudFormation intrinsic function docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html)):
```json
{
  "yaml.customTags": [
    "!Base64 scalar:string",
    "!Cidr sequence:array",
    "!And sequence:boolean",
    "!Equals sequence:boolean",
    "!If sequence",
    "!Not sequence:boolean",
    "!Or sequence:boolean",
    "!Condition scalar:boolean",
    "!FindInMap sequence:string",
    "!GetAtt scalar:string",
    "!GetAtt sequence:string",
    "!GetAZs scalar:array",
    "!ImportValue scalar:string",
    "!Join sequence:string",
    "!Select sequence:string",
    "!Split sequence:array",
    "!Sub scalar:string",
    "!Sub sequence:string",
    "!Transform mapping",
    "!Ref scalar:string"
  ]
}
```

2. Create a file named `template.yaml` with the following content:
```yaml
Resources:
  EC2Instance:
    Type: AWS::EC2::Instance
    Properties:
      # example of short-form tag with block sequence input
      AvailabilityZone: !GetAtt
        - Ec2SecurityGroup
        - GroupId

      # example of short-form tag with flow sequence input
      ImageId: !FindInMap [AWSRegionArch2AMI, !Ref "AWS::Region", HVM64]

      # example of short-form tag with scalar input
      InstanceType: !Ref InstanceType

      SecurityGroupIds:
        # example of short-form tag inside an array item
        - !GetAtt
          - Ec2SecurityGroup
          - GroupId
```
3. Expect no errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ The following settings are supported:
 - `yaml.schemaStore.url`: URL of a schema store catalog to use when downloading schemas.
 - `yaml.kubernetesCRDStore.enable`: When set to true the YAML language server will parse Kubernetes CRDs automatically and download them from the [CRD store](https://github.com/datreeio/CRDs-catalog).
 - `yaml.kubernetesCRDStore.url`: URL of a crd store catalog to use when downloading schemas. Defaults to `https://raw.githubusercontent.com/datreeio/CRDs-catalog/main`.
-- `yaml.customTags`: Array of custom tags that the parser will validate against. It has two ways to be used. Either an item in the array is a custom tag such as "!Ref" and it will automatically map !Ref to scalar or you can specify the type of the object !Ref should be e.g. "!Ref sequence". The type of object can be either scalar (for strings and booleans), sequence (for arrays), map (for objects).
+- `yaml.customTags`: Array of custom tags that the parser will validate against. It has three ways to be used. A tag without a type, such as "!Ref", is treated as a scalar tag. A tag with a node type, such as "!Ref sequence", specifies the YAML node type that the tag is written on. A tag with a node type and return type, such as "!FindInMap sequence:string", also specifies the schema type that the tagged value evaluates to. Supported node types are scalar, sequence, and mapping. Supported return types are string, number, integer, boolean, null, array, and object. The return type aliases scalar, sequence, and mapping are accepted as string, array, and object.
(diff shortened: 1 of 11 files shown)


datho7561 (approved):
Looks good, works well and seems really helpful. Thanks, Morgan!

PR #1263, opened 2026-05-28
Fix auto-completion for array with default

### LINKED ISSUE #1187: Array with default and required values are not rendered correctly
## Describe the bug

I have used the below schema. This schema has a property (foo) that is required, an array and have default values.

When I am in a new file that uses this schema and press Ctrl + Space to see every required property for my schema, I only see `${1}` for my array:

Image

When I press Tab to get this schema, I only get `hello` (the default value for my string property) in the spaces of the array.

This would therefore require any user of my schema to notice that the default value is completly wrong and then manuelly correct them.

The property foo in the below Schema exists in multiple ways, because I tested if one of them works (none work).

The normal way how I use default values with arrays is `foo-two-default`.

```json
{
  "properties": {
    "bar": {
      "type": "string",
      "default": "hello"
    },
    "num": {
      "type": "number",
      "default": "42"
    },
    "foo-two-default": {
      "type": "array",
      "default": [
        "foo-two-default"
      ],
      "items": {
        "type": "string",
        "default": "foo-two-default"
      }
    },
    "foo-default-in-items": {
      "type": "array",
      "items": {
        "type": "string",
        "default": "foo-default-in-items"
      }
    },
    "foo-default-outside": {
      "type": "array",
      "default": [
        "foo-default-outside"
      ],
      "items": {
        "type": "string"
      }
    }
  },
  "required": [
    "foo-two-default",
    "foo-default-in-items",
    "foo-default-outside",
    "bar",
    "num"
  ],
  "title": "dummy"
}
```

## Expected Behavior
I would expect that my array default values will be given normally as normal string or number properties are.

## Current Behavior
- In the preview, some placeholder values `${1}` are visible
- In the inital creation of the yaml document, the default value of my string value will be used. 
- When deleting the wrong default value and pressing CTRL + Space, the correct default values are found and rendered (OK)

## Steps to Reproduce
1. Use above schema in VS Code for any file
2. In the file: Press Ctrl + Space and see the preview
3. Use the preview to create the YAML content

Image

## Environment
- [x] Windows
- [ ] Mac
- [ ] Linux
- [ ] other (please specify)

shin19991207:
### What does this PR do?

Fixes auto-completion for array properties that define defaults either on the array property itself or on the array item schema.

Before this fix, auto-completing an array property inserted an empty item placeholder even when the schema provided a default.

Completion behavior is now:
- If the array property has `default`, use that value when completing the array property
- If the array property has no `default` but its `items` schema has `default`, complete the array property as an array with one item using the items default
- For subsequent array items, suggest items default when it exists; otherwise provide no suggestion
- Inserted default values are snippet placeholders, so users can tab through them and edit/replace the values easily

Fixes https://github.com/redhat-developer/vscode-yaml/issues/1187

### Is it tested? How?

- [x] New automated tests
- [x] Manual testing with the example from the linked issue

### DIFF
--- a/src/languageservice/services/yamlCompletion.ts
+++ b/src/languageservice/services/yamlCompletion.ts
@@ -708,7 +708,7 @@ export class YamlCompletion {
     const lineContent = textBuffer.getLineContent(overwriteRange.start.line);
     const hasOnlyWhitespace = lineContent.trim().length === 0;
     const hasColon = lineContent.indexOf(':') !== -1;
-    const isInArray = lineContent.trimLeft().indexOf('-') === 0;
+    const isInArray = lineContent.trimStart().indexOf('-') === 0;
     const nodeParent = doc.getParent(node);
     const matchOriginal = matchingSchemas.find((it) => it.node.internalNode === originalNode && it.schema.properties);
     const oneOfSchema = matchingSchemas.filter((schema) => schema.schema.oneOf).map((oneOfSchema) => oneOfSchema.schema.oneOf)[0];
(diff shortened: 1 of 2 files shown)


datho7561 (approved):

PR #1265, opened 2026-05-31
Fix to show SchemaStore entries without `fileMatch` field in schema selection

### LINKED ISSUE #1130: Missing schemas from schemastore
## Describe the bug

Trying to apply Traefik v3 schema to currently editing `traefik.yaml` file but only getting Traefik v2 proposal, even though Traefik v3 is defined in https://www.schemastore.org/api/json/catalog.json

Image

## Expected Behavior

Traefik v3 should be in the list

## Current Behavior

Only Traefik v2 is listed

## Steps to Reproduce

1. Install vscode extension
2. edit traefik.yaml file
3. Syntax is matched to Traefik v2 syntax with no option to change it to Traefik v3

## Environment
- [ ] Windows
- [x] Mac
- [ ] Linux
- [ ] other (please specify)

shin19991207:
### What does this PR do?
SchemaStore entries were only added to schema settings when they define a `fileMatch` field. Schemas like Traefik v3, which exist in SchemaStore but do not define `fileMatch`, were dropped before they could appear in the schema picker UI.

This fix keeps all SchemaStore entries even without a `fileMatch` field, so all schemas from SchemaStore remain selectable by users.

Fixes https://github.com/redhat-developer/vscode-yaml/issues/1130

### Is it tested? How?

- [x] New automated tests
- [x] Tested manually: Expect schemas without a `fileMatch` field to be displayed in the schema picker UI, e.g. search for "Traefik v3" and it should appear.

### DIFF
--- a/src/languageserver/handlers/settingsHandlers.ts
+++ b/src/languageserver/handlers/settingsHandlers.ts
@@ -257,10 +257,21 @@ export class SettingsHandler {
 
     for (const schemaIndex in schemas.schemas) {
       const schema = schemas.schemas[schemaIndex];
-
-      if (schema && schema.fileMatch) {
-        for (const fileMatch in schema.fileMatch) {
-          const currFileMatch: string = schema.fileMatch[fileMatch];
+      if (!schema.url) {
+        continue;
+      }
+      const fileMatches = schema.fileMatch ?? [];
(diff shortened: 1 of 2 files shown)


datho7561 (approved):
Looks good and fixes the issue. Thanks, Morgan!

PR #1268, opened 2026-06-02
Fix autocomplete suggestions from conditional if schema branches

### LINKED ISSUE #1266: Strange completions in github workflows
## Describe the bug

I'm seeing weird `Github Workflow` completions in a github workflow schema, inside `workflow_dispatch` event inputs, for example.

https://github.com/user-attachments/assets/fbe18e81-c030-4ec1-91ae-ef46ee5eb4de

Those insert a `type: ...` snippets of a kind.

For one, I don't think it's really needed if it's on purpose and two: the names of those are very bad.

## Expected Behavior

No such `Github Workflow` completions present.

## Current Behavior

Image

## Steps to Reproduce

1. Open workflow like:
```yml
name: Check configuration schema changes

on:
  workflow_dispatch:
    inputs:
      sdaf:
       <cursor>   

jobs:
  foo:
    uses: repo/test.yml@main
```

2. Place the cursor in place of `<cursor>` and trigger completions

## Environment
- [ ] Windows
- [x] Mac
- [ ] Linux
- [ ] other (please specify)

shin19991207:
### What does this PR do?
Fix to avoid auto-completion from using `if` condition schemas as completion sources
Since `if` schema is only a guard used to decide whether a `then` or an `else` branch applies, it should not contribute completion candidates itself.

Fixes https://github.com/redhat-developer/yaml-language-server/issues/1266

### Is it tested? How?

- [x] New automated test
- [x] Tested manually with the case reported in the linked issue

### DIFF
--- a/src/languageservice/parser/schemaValidation/baseValidator.ts
+++ b/src/languageservice/parser/schemaValidation/baseValidator.ts
@@ -703,7 +703,9 @@ export abstract class BaseValidator {
       const subMatchingSchemas = matchingSchemas.newSub();
 
       this.validateNode(node, subSchema, original, subValidationResult, subMatchingSchemas, options);
-      matchingSchemas.merge(subMatchingSchemas);
+      if (!callFromAutoComplete) {
+        matchingSchemas.merge(subMatchingSchemas);
+      }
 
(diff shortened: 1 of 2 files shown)


datho7561 (approved):

PR #1272, opened 2026-06-07
Use modern TypeScript module resolution

remcohaszing:
### What does this PR do?

This project was still using the deprecated TypeScript module resolution `node10` (aka `node`). This change updates the module resolution to `nodenext`. Imports were fixed accordingly.

Notably, `yaml/dist/parse/cst` isn’t exported, but this is exported as the `CST` namespace in the `yaml` module.

This is a preparation for updating to LSP 3.18, because some related packages no longer support the old module resolution anymore. It’s also a preparation to go ESM-only at some point.

N/A

### Is it tested? How?

```
tsc
```


datho7561 (approved):
Looks good and it's an important step towards es module support. Thanks, Remco!

PR #1274, opened 2026-06-09
Support disabling schema validation via modeline

### LINKED ISSUE #446: A way to disable schema loading for a specific file using modeline
Sometimes I want to be able to create a YAML file without a schema, for example to hold generic config. In some cases the naming of this file would match some generic pattern, and yaml-language-server will then try to validate this file against that schema. My editor is then filled with validation errors and bogus autocomplete suggestions.

I wish I could use the modeline to declare that this file is without a schema.

```yaml
# Some ideas:
# yaml-language-server: $schema=none
# yaml-language-server: disable
# yaml-language-server: schemaless
```

I already tried the following, none of which worked:

```yaml
# yaml-language-server: $schema=/dev/null
# yaml-language-server: $schema=null
# yaml-language-server: $schema=undefined
# yaml-language-server: $schema=
```

shin19991207:
### What does this PR do?
Adds support for disabling schema validation for a single YAML file with a modeline:

```yaml
# yaml-language-server: $schema=none
```

Since we support disabling schema detection for specific files using the `yaml.disableSchemaDetection` setting, I think it makes sense to offer the same control through a modeline.

Fixes https://github.com/redhat-developer/yaml-language-server/issues/446

### Is it tested? How?

New automated test + tested manually

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ The following settings are supported:
 
 ## Suppressing diagnostics
 
-You can suppress specific validation warnings on a per-line basis by adding a `# yaml-language-server-disable` comment on the line immediately before the one producing the diagnostic.
+You can suppress specific validation warnings on a per-line basis by adding a `# yaml-language-server-disable` comment on the line immediately before the one producing the diagnostic. To disable schema validation for an entire file, use a [modeline schema association](#using-modeline).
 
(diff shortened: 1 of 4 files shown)


datho7561 (approved):

PR #1275, opened 2026-06-10
Update to TypeScript 6

remcohaszing:
### What does this PR do?

TypeScrit 5 and 6 both removed support for some deprecated features. So the config and code have been minimally adjusted.

The `module` option for the CJS build was changed to `nodenext`. This implies a matching `moduleResolution` and
`allowSyntheticDefaultImports`, to these are now redundant and were removed. `strict` is now true. All non-strict options the code is not compliant with have been disabled explicitly.
`forceConsistentCasingInFileNames` is now true by default, to it was removed.

The correct `module` option for ESM is also `nodenext`, but with `"type": "module"` in `package.json. Unfortunately we can’t use this yet for dual publishing, so as a middle ground, `module` was left at `esnext` and we have to use the `bundler` module resolution. `rootDir` detection was removed, so this is now configured explicitly, with a matching `include` option. `downlevelIteration` is deprecated and was removed.

N/A

### Is it tested? How?

```sh
tsc
tsc -P tsconfig.esm.json
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,7 +48,7 @@
         "sinon-chai": "^3.5.0",
         "source-map-support": "^0.5.19",
         "ts-node": "^10.0.0",
-        "typescript": "^4.8.3",
+        "typescript": "^6.0.0",
         "typescript-eslint": "^8.59.4"
       }
     },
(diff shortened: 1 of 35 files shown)


datho7561 (approved):
This seems like an important step for this project in terms of modernizing it and preparing for ES Modules.

In the future, we might want to reenable some of the strict mode settings that were enabled by default as a part of TypeScript 6, probably one by one.

remcohaszing:
> we might want to reenable

Just to be clear, they were never enabled. TypeScript 6 enables them by default, meaning they now need to be disabled explicitly instead of implicitly.

But yes, I strongly recommend working towards enabling them.

datho7561:
@remcohaszing when you have a chance, do you mind rebasing this? I'll merge soon after it's rebased

remcohaszing:
Rebased!

PR #1277, opened 2026-06-12
Add file extensions to all import statements

remcohaszing:
### What does this PR do?

This is literally a regex replace done in VSCode, with some minor fixes for directory imports.

On its own this change is not very useful. This is a preparation for ESM. ESM imports require file extensions in Node.js. The goal of this change is to prevent all the noise this causes when the actual more meaningful changes are made.

I chose to use `.ts` extensions and `rewriteRelativeImportExtensions` to align with #1257.

### Is it tested? How?

```sh
tsc
```


datho7561:
It seems ts-node doesn't like the .ts file extensions, and it tries to use node's type striping to load the files instead of ts-node.

remcohaszing:
Yes, this did not work out the way I had hoped. I provided a couple more pull requests to to work towards this goal that do pass CI.

PR #1280, opened 2026-06-13
docs: include default values for language server settings

### LINKED ISSUE #1055: Default settings
The documentation lists all available settings, but it doesn't specify their default values. Could you include those defaults so we have a clear understanding?

SSDWGG:
Fixes #1055

## Summary
- convert the language server settings list into a table
- document the default value for each listed setting from the current settings state and schema URL constants

## Validation
- Not run; documentation-only change prepared via GitHub API without a local checkout.

AI assistance disclosure: This pull request was prepared with the assistance of AI and has not yet been reviewed by a human.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,37 +29,37 @@ Starting from `1.0.0` the language server uses [eemeli/yaml](https://github.com/
 
 The following settings are supported:
 
-| Setting | Default | Description |
-| --- | --- | --- |
-| `yaml.yamlVersion` | `1.2` | Set default YAML spec version (`1.2` or `1.1`). |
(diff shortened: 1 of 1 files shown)


datho7561:
Looking at the markdown preview, the table doesn't look all that great on GitHub. The last column is as wide as the first two and gets line wrapped, which makes it hard to read. I think it would be better to add the default in parentheses after the name of the setting.

SSDWGG:
Updated the settings table to fold defaults into the setting name column, so the description column has more room in GitHub's markdown preview.

PR #1281, opened 2026-06-13
Update to ESLint 10

remcohaszing:
### What does this PR do?

I tried to keep the config in the same spirit as the old configuration.

I replaced `eslint-plugin-import` with `eslint-plugin-import-x`. `eslint-plugin-import` is getting outdated due to its philosophy to keep support with ancient versions of Node.js and ESLint.

I removed rules that are covered by Prettier.

I disabled `import-x/no-unresolved`. This rules causes false positives and catches problems that TypeScript already catches anyway.

Now all files are linted, by removing the file paths from the scripts in `package.json`. The path defaults to `.`.

N/A

### Is it tested? How?

```sh
eslint
```


remcohaszing:
It may also be time to remove `@microsoft/eslint-formatter-sarif`. It pulls in the `eslint@8` dependency for no reason. This is solved by https://github.com/microsoft/sarif-js-sdk/pull/92, but the project appears to be abandoned.

datho7561 (approved):
Looks good. Thanks, Remco!

PR #1282, opened 2026-06-13
Consistent type imports

remcohaszing:
### What does this PR do?

With the goal in mind to use Node.js type stripping, we will have to use
type imports for types.

The ESLint rule `@typescript-eslint/consistent-type-imports` enforces us
to use type imports. The rule `import-x/consistent-type-specifier-style`
enforces a consistent style how to write type imports. I chose
`prefer-top-level`, because that means TypeScript and Node.js understand
they can strip the entire import statement instead of just the members.
So that yields better runtime performance and it may even be required in
some cases.

This builds on top of #1281

### Is it tested? How?

```
eslint
tsc
```


remcohaszing:
Fiendly ping @datho7561. I figured you might have missed that I rebased  this :)

datho7561:
Sorry, I got caught up in other work. I'll try to get to it tomorrow afternoon.

datho7561 (approved):
Looks good. Thanks, Remco!

Going to confirm with Morgan if I can merge this, it'll be a pretty annoying rebase for her PR

remcohaszing:
I see how this can cause merge conflicts, but they should be easy to resolve. Just pick the other change, then autofix the imports with `eslint --fix`.

datho7561:
Ahhh okay with an autofix that'll be much easier

PR #1283, opened 2026-06-13
Sort imports

remcohaszing:
### What does this PR do?

This change enforces some ESLint rules to ensure consistent import sorting. In my experience this helps identifying how imports, files, and dependencies are used.

The `import-x/first` rule ensures all imports are on top of the file.

The `import-x/order` rule forces a consistent order of import statements. Imports are grouped by Node.js builtins, external dependencies, and relative imports. Type imports are sorted before values. Within each group, the statements are alphabetized.

The `sort-imports` rule sorts the imported members.

The `import/newline-after-import` rule enforces a newline after the last import.

This is built on top of #1282

### Is it tested? How?

```sh
eslint
tsc
```

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -33,10 +33,22 @@ export default defineConfig(
       },
     },
     rules: {
+      'sort-imports': ['error', { ignoreDeclarationSort: true }],
       '@typescript-eslint/consistent-type-imports': 'error',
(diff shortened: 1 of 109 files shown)


remcohaszing:
The failing check does not fail for me locally. Maybe look at the related PRs before considering this change. Then I’ll rebase if those are accepted.

remcohaszing:
Note that this change is purely stylistic.

PR #1284, opened 2026-06-13
Rewrite Mocha tests to use promises

remcohaszing:
### What does this PR do?

The goal is to eventually move to the builtin Node.js test runner. This test runner uses a syntax very similar to Mocha, but it doesn’t provide a done callback. Instead, we must use promises.

Mocha also supports promises, so we can already rewrite the tests like that.

Also promise based tests just look nicer. :)

N/A

### Is it tested? How?

```sh
npm test
```


remcohaszing:
I recommend reviewing this with whitespace changes disabled.

datho7561 (approved):
Looks good! Thanks, Remco!

PR #1290, opened 2026-06-30
Fix: `yaml/get/all/jsonSchemas` returns schema association declared with inline `$schema`

shin19991207:
### What does this PR do?

Fix: update `yaml/get/all/jsonSchemas` API when a schema is declared via inline `$schema` obejct

Fixes a bug from https://github.com/redhat-developer/yaml-language-server/pull/970

### Is it tested? How?

- [x] New automated tests
- [x] Tested manually with https://github.com/redhat-developer/vscode-yaml/pull/1263


datho7561 (approved):
Looks good. Thanks, Morgan!

PR #1296, opened 2026-07-17
Fix fileMatch single-star glob matching across directories

### LINKED ISSUE #1295: Schema fileMatch regression in 1.24.0: single `*` in glob patterns matches across path separators (picomatch `bash: true`)
### Describe the bug

Since yaml-language-server 1.24.0 (shipped in vscode-yaml 1.24.0), a single `*` in schema `fileMatch` glob patterns matches across `/` path separators, i.e. it behaves like `**`. This changes which schemas from SchemaStore (and from user `yaml.schemas` settings) get associated with which files, and produces false validation errors.

Concrete real-world impact: the SchemaStore "Ansible Playbook" schema declares

```json
"fileMatch": ["playbook.yml", "playbook.yaml", "site.yml", "site.yaml", "**/playbooks/*.yml", "**/playbooks/*.yaml"]
```

Up to 1.23.0, `**/playbooks/*.yml` matched only direct children of a `playbooks/` directory. In 1.24.0 it matches **every** `.yml` file anywhere below any `playbooks/` directory. In a common Ansible layout where roles live under the playbooks directory (`playbooks/roles/<role>/tasks/main.yml`), every tasks/handlers file now additionally gets the *playbook* schema applied on top of the correct *tasks* schema, so every task is flagged with:

```
Missing property "ansible.builtin.import_playbook".
```

Downgrading vscode-yaml to 1.23.0 makes the errors disappear.

### Root cause

1.24.0 replaced the `vscode-json-languageservice`-based file pattern matching with picomatch. `FilePatternAssociation` now compiles patterns with the `bash: true` option:

https://github.com/redhat-developer/yaml-language-server/blob/1.24.0/src/languageservice/services/yamlSchemaService.ts#L211-L214

```ts
this.isMatch = picomatch(processedPatterns, {
  bash: true,
  noglobstar: false,
});
```

Per the picomatch documentation, `bash: true` "treats single stars as globstars (`**`)" (see the `star = opts.bash === true ? globstar(opts) : STAR` handling in `picomatch/lib/parse.js`). So every `*` in every `fileMatch` pattern is silently promoted to `**`.

The previous implementation (`vscode-json-languageservice` `FilePatternAssociation`, used via `createRegex(..., { extended: true, globstar: true })`) kept the conventional glob semantics where `*` does not cross path separators. VS Code's built-in JSON language service still behaves that way, so 1.24.0 diverges from both its own previous behavior and from how `fileMatch` works for JSON files.

### Steps to reproduce

1. Install vscode-yaml 1.24.0 with default settings (SchemaStore enabled).
2. Create this file tree:

   ```
   playbooks/
   ├── site.yml
   └── roles/
       └── demo/
           └── tasks/
               └── main.yml
   ```

3. Put a trivial tasks file in `playbooks/roles/demo/tasks/main.yml`:

   ```yaml
   ---
   - name: Ping
     ansible.builtin.ping:
   ```

4. Open `main.yml`.

### Actual behavior

The task is flagged with `Missing property "ansible.builtin.import_playbook".` — the file is matched by the Ansible **Playbook** schema (`**/playbooks/*.yml`) in addition to the Ansible **Tasks** schema (`**/tasks/*.yml`).

### Expected behavior

`**/playbooks/*.yml` must not match `playbooks/roles/demo/tasks/main.yml`; a single `*` should not cross `/`, matching the pre-1.24.0 behavior and standard glob semantics. Only the tasks schema should apply, and the file should validate cleanly (as it does in 1.23.0).

### Suggested fix

Drop `bash: true` from the picomatch options in `FilePatternAssociation` (or set it to `false` explicitly). `noglobstar: false` already provides `**` support; `bash: true` only adds the unwanted star promotion (and disables backslash escapes).

### Environment

- vscode-yaml: 1.24.0 (regression), 1.23.0 (works)
- yaml-language-server: 1.24.0
- OS: any (reproduced on Linux / WSL2)

### Workaround

Pin vscode-yaml to 1.23.0, or override the association via explicit `yaml.schemas` settings.

shin19991207:
### What does this PR do?
The configured picomatch with `bash: true` makes a single `*` behave like `**` and match across directory separators.
Removed `bash: true` so `*` only matches within one directory level.

Fixes https://github.com/redhat-developer/yaml-language-server/issues/1295

### Is it tested? How?

- [x] Existing tests
- [x] Manually tested with the case reported in https://github.com/redhat-developer/yaml-language-server/issues/1295
- [x] Added a test covering direct and nested paths

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -209,7 +209,6 @@ export class FilePatternAssociation {
         .filter((p) => p.length > 0);
 
       this.isMatch = picomatch(processedPatterns, {
-        bash: true,
         noglobstar: false,
       });
 
--- a/test/yamlSchemaService.test.ts
+++ b/test/yamlSchemaService.test.ts
@@ -896,5 +896,17 @@ spec:
       const nestedConfigUris = service.getSchemaURIsForResource('nested/repo/.github/ISSUE_TEMPLATE/config.yml');
       expect(nestedConfigUris).to.not.include('https://json.schemastore.org/github-issue-forms.json');
     });
+
+    it('should not match a single star across path separators', () => {
+      const service = new SchemaService.YAMLSchemaService(requestServiceMock);
+      const playbookSchemaUri = 'https://json.schemastore.org/ansible-playbook.json';
+      service.registerExternalSchema(playbookSchemaUri, ['**/playbooks/*.yml'], {});
+
+      const playbookUris = service.getSchemaURIsForResource('project/playbooks/site.yml');
+      expect(playbookUris).to.include(playbookSchemaUri);
+
+      const roleTaskUris = service.getSchemaURIsForResource('project/playbooks/roles/demo/tasks/main.yml');
+      expect(roleTaskUris).to.not.include(playbookSchemaUri);
+    });
   });
 });


datho7561 (approved):
Looks good and works well from my testing. Thanks, Morgan!

PR #1299, opened 2026-07-20
Prepare for OIDC authentication for publishing

### LINKED ISSUE #1292: Prepare to use OIDC for publishing in GitHub Actions
In order to support OIDC for publishing to npm, (OIDC is the newer, recommended way of authenticating with npmjs), we'll need to make some changes to the GitHub Actions: https://docs.npmjs.com/trusted-publishers#github-actions-configuration

datho7561:
### What does this PR do?
Adds the necessary permissions to the GitHub Actions to authenticate using OIDC, so that we can switch over authentication for publishing to NPM over to OIDC.

Fixes #1292

### Is it tested? How?
Nope, we'll have to try it once we have OIDC set up.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -8,6 +8,10 @@ on:
     tags:
       - '*'
 
+permissions:
+  id-token: write  # Required for OIDC
+  contents: read
+
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
fbricon, line 56: are you sure this is still necessary?

datho7561: No, it's not needed, I can remove it in a separate PR.

PR #1301, opened 2026-07-21
fix: do not depend on specific working dir within container image

ProbstDJakob:
### What does this PR do?
It replaces the relative path within the `ENTRYPOINT` instruction with a absolute one and removes the `WORKDIR`.

None

### Is it tested? How?
I use the container image with a patched `ENTRYPOINT` and set `--workdir` on command line for quite some time without issues.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,10 @@
 FROM node:lts
 
-WORKDIR /yaml-language-server
+COPY . /yaml-language-server
 
-COPY . .
-
-RUN npm install && \
+RUN cd /yaml-language-server && \
+    npm install && \
     npm run build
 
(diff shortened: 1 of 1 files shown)


datho7561:
What's the benefit of doing this vs. the way it's currently set up?

ProbstDJakob:
When changing the working directory within the container via `--workdir` the application does not work as the `ENTRYPOINT` is relativ and most likely does not exist in the new work dir. I change the working directory because I have a wrapper script setting up many language server and formatter and passing the project in as bind and changing the work dir to it as some of the tools depend on it. This breaks with the current way this image is built.

datho7561 (approved):
Looks good, and seems like a good idea. Thanks, @ProbstDJakob !

PR #1308, opened 2026-07-28
Improve handling of schemas written using YAML

### LINKED ISSUE #1275: First key is quoted makes the schema unparsable
## Describe the bug

See https://github.com/apis-json/api-json/blob/develop/spec/schema_0.21.yml

This is not usable for schema validation with the vscode extension, since it doesn't like if the first key is quoted. Omitting the quotes on the first key make it work, if any subsequent key is wrapped with quotes, it works fine though, so it's just about the first one. 

Error when it has quotes:

```
Unable to parse content from '[..]/apisjson-schema_0.21.yml': Parse error at line: 1 column: 1YAML(65536)
````

## Environment
- [x] Windows
- [ ] Mac
- [ ] Linux
- [ ] other (please specify)

datho7561:
### What does this PR do?
- Use the file extension of the schema to determine if the schema was written using JSON or YAML
- Refactor the schema resolution logic to be a bit simpler
  - Avoid requesting the content twice

Fixes redhat-developer/vscode-yaml#1275

### Is it tested? How?
- [x] Unit tests

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -1702,17 +1702,8 @@ export class YAMLSchemaService implements IJSONSchemaService {
     return this.getAssociatedSchemas(resource);
   }
 
-  private async loadJSONSchema(url: string): Promise<UnresolvedSchema> {
-    if (!this.requestService) {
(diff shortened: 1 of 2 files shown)


shin19991207 (approved):
Looks good and works well for me. Thanks, David!

PR #1309, opened 2026-07-29
Fix Kubernetes schema priority after resource-specific schema resolution

### LINKED ISSUE #1298: Confusing interaction of schemastore and kubernetes manifests
## Describe the bug

Integration with schemastore (enabled by default) sometimes results in weird/incorrect behavior for kubernetes manifests. I found that, if I place the manifest inside "policies/" directory, it makes kubernetes validation work incorrectly, which I traced back to schemastore integration. Offending schemastore section:

```json
    {
      "name": "Aerleon Policy",
      "description": "Aerleon policy file",
      "fileMatch": ["**/policies/**/*.yaml"],
      "url": "https://raw.githubusercontent.com/aerleon/aerleon/main/schemas/aerleon-policies.schema.json"
    },
```

So, the validation works correctly for file `templates/foo.yaml`, but not for `templates/policies/foo.yaml`.

Confusingly, the behavior is different based on document order. If I have a file with the following contents, it correctly validates Pod, but doesn't work for MutatingAdmissionPolicy:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: irrelevant
spec:
  foo1: bar # finds the error here correctly
  containers:
    - name: irrelevant
---
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingAdmissionPolicy
metadata:
  name: irrelevant
spec:
  foo2: bar # doesn't find the error
  failurePolicy: Fail
```

But if I swap document order, it doesn't work for any of them anymore:

```yaml
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingAdmissionPolicy
metadata:
  name: irrelevant
spec:
  foo2: bar # doesn't find the error
  failurePolicy: Fail
---
apiVersion: v1
kind: Pod
metadata:
  name: irrelevant
spec:
  foo1: bar # doesn't find the error
  containers:
    - name: irrelevant
```

Setting `yaml.schemaStore.enable=false` removes the problem. However, I don't want to disable it completely since I find it useful for other files. I just want kubernetes manifests to not be treated as anything else.

The problem is reproduced for other fileMatch patterns from schemastore as well. For example, if I rename the directory to "slices", it starts complaining that the whole MutatingAdmissionPolicy is wrong.

## Environment
- [ ] Windows
- [x] Mac
- [ ] Linux
- [ ] other (please specify)

yaml-language-server version 1.24.0

Config:

```json
{
    "yaml": {
        "kubernetesVersion": "1.36.1",
        "schemas": {
            "kubernetes": "templates/**"
        }
    }
}
```

`kubernetesVersion` is needed to recognize MutatingAdmissionPolicy v1. `schemas.kubernetes` is only set for testing. Originally, the issue was discovered through helm-ls integration, which sets kubernetes file match in a same way: https://github.com/mrjosh/helm-ls/blob/v0.5.4/internal/util/config.go#L162

shin19991207:
### What does this PR do?

Kubernetes auto-detection replaces the configured `all.json` URI with a document-specific `_definitions.json` or CRD URI. However, schema priorities are registered against the original `all.json` URI. Looking up priority using the document-specific URI caused it to receive the lowest priority, allowing other schema sources such as SchemaStore to incorrectly win.

This PR fixes the issue above and includes several related enhancements to Kubernetes support:

- Treats an omitted Kubernetes API group as the internal `core` group. For example, `apiVersion: v1` is parsed as `{ group: "core", version: "v1" }`, allowing a `Pod` to match the `io.k8s.api.core.v1.Pod` definition. Unknown core resources fall back to `all.json` instead of being searched for in the CRD catalog
- Creates a separate CodeLens entry at the beginning of each document in a multi-document YAML file
- Displays specific CodeLens titles for Kubernetes resources, such as `Pod (Kubernetes v1.36.1)` and `MutatingAdmissionPolicy (Kubernetes v1.36.1)`, instead of displaying the general `all.json` / `_definitions.json`
- Preserves extensionless schema URI names instead of incorrectly appending `.json`. For example, a schema URI ending in `schema` is displayed as `schema`, not `schema.json`, because schema resources may be JSON, YAML, or extensionless

Fixes https://github.com/redhat-developer/yaml-language-server/issues/1298

### Is it tested? How?

- [x] Automated tests
- [x] Tested the case reported in https://github.com/redhat-developer/yaml-language-server/issues/1298. Verified that:
  1. Both documents are validated against their corresponding Kubernetes schemas, and validation errors are reported for both.
  2. A CodeLens appears at the top of each document and links to the corresponding schema.

- [x] Tested that an unknown core kind falls back to `all.json` instead of being resolved through the CRD catalog. For example, expected the following YAML reports a diagnostic on `Po`: `Value is not accepted. Valid values: "MutatingAdmissionPolicy", "MutatingAdmissionPolicyBinding", "MutatingWebhookConfiguration"...`, rather than attempting CRD validation:
  ```yaml
  apiVersion: v1
  kind: Po
  metadata:
    name: irrelevant
  ```

### DIFF
--- a/test/yamlSchemaService.test.ts
+++ b/test/yamlSchemaService.test.ts
@@ -8,11 +8,13 @@ import * as path from 'path';
 import * as sinon from 'sinon';
 import sinonChai from 'sinon-chai';
 import * as url from 'url';
+import { TextDocument } from 'vscode-languageserver-textdocument';
 import * as YAML from 'yaml';
 import type { JSONSchema } from '../src/languageservice/jsonSchema';
 import { parse } from '../src/languageservice/parser/yamlParser07';
 import * as SchemaService from '../src/languageservice/services/yamlSchemaService';
 import { DEFAULT_KUBERNETES_SCHEMA_VERSION, getSchemaUrls } from '../src/languageservice/utils/schemaUrls';
+import { SchemaPriority } from '../src/languageservice/yamlLanguageService';
 import { SettingsState } from '../src/yamlSettings';
 
 const BASE_KUBERNETES_SCHEMA_URL = `https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/${DEFAULT_KUBERNETES_SCHEMA_VERSION}-standalone-strict/`;
@@ -663,6 +665,20 @@ spec:
       expect(requestServiceMock.callCount).equals(4);
     });
 
+    it('should fall back to all.json instead of the CRD catalog for an unknown core resource', async () => {
+      const yamlDock = parse('apiVersion: v1\nkind: UnknownCoreResource');
+      const settings = new SettingsState();
+      settings.kubernetesCRDStoreEnabled = true;
+      requestServiceMock = sandbox.fake.resolves('{"oneOf": []}');
+
+      const service = new SchemaService.YAMLSchemaService(requestServiceMock, undefined, undefined, settings);
+      service.registerExternalSchema(KUBERNETES_SCHEMA_URL, ['*.yaml']);
+      const resolvedSchema = await service.getSchemaForResource('test.yaml', yamlDock.documents[0]);
+
+      expect(resolvedSchema.schema.url).equals(KUBERNETES_SCHEMA_URL);
+      expect(requestServiceMock).calledOnceWithExactly(KUBERNETES_SCHEMA_URL);
+    });
+
     it('should not get schema from crd catalog if definition in kubernetes schema (multiple oneOf)', async () => {
       const documentContent = 'apiVersion: apps/v1\nkind: Deployment';
       const content = `${documentContent}`;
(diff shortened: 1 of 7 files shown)


datho7561 (dismissed):
Works really well and addresses the issue. I think the changes to CodeLens especially are super important. Thanks, Morgan!

One thing that might be interesting, to align with the other changes to CodeLens in this PR, is to have it say "Kubernetes v1.36.1" instead of "all.json" when it's validating against Kubernetes but not any specific resource kind. Do you think that's interesting, and if so, do you want to implement that here or in a different PR?

shin19991207:
@datho7561 Yes that makes sense! I'll implement it in this PR as well.

test/yamlSchemaService.test.ts
```diff
@@ -869,6 +884,86 @@ spec:
...
+      service.addSchemaPriority(KUBERNETES_SCHEMA_URL, SchemaPriority.SchemaAssociation);
+      service.registerExternalSchema(aerleonSchemaURI, ['**/policies/**/*.yaml'], { type: 'object' });
+      service.addSchemaPriority(aerleonSchemaURI, SchemaPriority.SchemaStore);
+
+      const pod = `apiVersion: v1
+kind: Pod
+metadata:
+  name: irrelevant
+spec:
+  foo1: bar
```
artem-nefedov, line 897: Maybe the test should also confirm that these incorrect fields are detected as such?
Otherwise, I don't understand why they're present here.

artem-nefedov:
The PR title says "Fix ... for multi-document YAML", but the problem I reported is also reproduced with a single-document file (e.g. one containing just MutatingAdmissionPolicy).

datho7561:
Here's an edge case I ran into:
image
The code lens ends up on the previous line for the third document that's empty. I don't think this is a big deal, but it might be worth fixing.

datho7561 (dismissed):
It works really well and it's really helpful. If you think https://github.com/redhat-developer/yaml-language-server/pull/1309#issuecomment-5134389760 is worth addressing now, that might be worth it. Otherwise, I think this PR is good to merge, so feel free to go ahead and do that.

shin19991207:
@datho7561 Thanks for catching this edge case! I fixed the issue so the CodeLens is now placed on the line after a trailing `---` when a following line exists.

datho7561 (approved):
Looks good! Thanks, Morgan!

PR #1311, opened 2026-07-31
fix: recognize Kubernetes CRD schema as builtin

### LINKED ISSUE #1310: Server incorrectly treats kubernetes CRD as a CRD
OK, I know the title sounds stupid, but that's the actual problem.

CustomResourceDefinition is a kubernetes-native resource, but the server tries to fetch its schema from CRD store.

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: a.b.c
```

Error:

```
customresourcedefinition_v1.json:1:1: Unable to load schema from 'https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/apiextensions.k8s.io/customresourcedefinition_v1.json': No content.
```

This is reproduced on latest stable release 1.24.0, and also on a latest main "1.25.1-2b95f5a.0".

Pybsama:
### What does this PR do?

Kubernetes publishes `CustomResourceDefinition` under the
`io.k8s.apiextensions-apiserver.pkg.apis` definition prefix rather than the
usual `io.k8s.api` prefix. The current exact matcher therefore misses this
builtin resource and falls back to a nonexistent entry in the external CRD
catalog.

This change:

- maps the exact `apiextensions.k8s.io` API group to its generated Kubernetes
  definition prefix;
- keeps the existing exact `io.k8s.api` matching behavior for every other
  group;
- falls back to `all.json` when an apiextensions kind or version is absent from
  the selected Kubernetes schema instead of treating it as a custom resource;
- preserves CRD catalog lookup for similarly named custom groups.

Fixes #1310

### Is it tested? How?

- `npm run build`
- Node.js 22: ESLint, main/UMD/ESM TypeScript compilation
- 9 focused Kubernetes schema-selection scenarios
- 1,305 passing full-suite tests and 5 existing pending tests

The full suite currently has one unrelated live SchemaStore assertion for the
Drone schema. It reproduces unchanged on the exact `main` base commit because
the externally served diagnostic ordering no longer matches the fixture; the
1,305 remaining tests pass under Node.js 22.

AI assistance was used during implementation and review; I verified the
changes and all reported test results.

### DIFF
--- a/src/languageservice/services/k8sSchemaUtil.ts
+++ b/src/languageservice/services/k8sSchemaUtil.ts
@@ -28,8 +28,8 @@ export function autoDetectKubernetesSchema(
   if (builtinResource) {
     return builtinResource;
   }
-  // core resources cannot be CRDs: fall back to all.json for unknown core kinds
-  if (gvk.group === 'core') {
+  // core and apiextensions resources cannot be CRDs: fall back to all.json for unknown kinds or versions
+  if (gvk.group === 'core' || gvk.group === 'apiextensions.k8s.io') {
     return undefined;
   }
   const customResource = autoDetectCustomResource(gvk, crdCatalogURI);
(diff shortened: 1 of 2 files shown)


datho7561:
This seems to be the right direction. @shin19991207 may have also started something, I don't know for sure though.

shin19991207:
@datho7561 Yes, this seems to be the right direction. And I haven't started anything on this so please feel free to start reviewing on this PR!

datho7561 (approved):
Looks good and works well! Thanks for the contribution! Is there anything else that needs to be done on here?

artem-nefedov:
@datho7561 looks like this can be merged?

datho7561 (approved):

PR #1312, opened 2026-07-31
Fix schema priority lookup for normalized file URIs

shin19991207:
### What does this PR do?

Before the fix, schema registration normalizes URIs to skip encodings, however, schema priorities were stored using the original URI. For an encoded Windows file URI, this produced different keys:

```text
Priority key:  file:///c%3A/Users/.../schema.json
Registered ID: file:///c:/Users/.../schema.json
```

As a result, the configured schema's priority could not be found and defaulted to `0`. Another matching schema could then incorrectly take precedence over the schema configured through `yaml.schemas`.

This PR normalizes URIs consistently when recording schema priorities, ensuring encoded and unencoded forms of the same URI use the same key.

Related to https://github.com/redhat-developer/vscode-yaml/issues/1274

### Is it tested? How?

- [x] New automated tests
- [x] Tested manually on Mac using the following setting:
  ```json
  "yaml.schemas": {
      "file:///Users/.../encoded%3Aschema.json": "test.yaml",
      "file:///Users/.../schema.json": "test.yaml"
  }
  ```
  Expected `test.yaml` to be validated against `Multiple JSON Schemas` (both `encoded:schema.json` and `schema.json`), instead of just `schema.json`.

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -1483,12 +1483,13 @@ export class YAMLSchemaService implements IJSONSchemaService {
 
   // Set the priority of a schema in the schema service
   public addSchemaPriority(uri: string, priority: number): void {
-    let currSchemaArray = this.schemaPriorityMapping.get(uri);
+    const id = normalizeId(uri);
+    const currSchemaArray = this.schemaPriorityMapping.get(id);
     if (currSchemaArray) {
-      currSchemaArray = currSchemaArray.add(priority);
-      this.schemaPriorityMapping.set(uri, currSchemaArray);
+      currSchemaArray.add(priority);
+      this.schemaPriorityMapping.set(id, currSchemaArray);
     } else {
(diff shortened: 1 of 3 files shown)


datho7561 (approved):
I confirmed that overriding the schema with an encoded colon causes it to not be used when there's two schemas assgined to a file, and that this fixes that.

I'll play around with it on Windows too to see if I can replicate that problem and if this fixes it. Do you want to merge this as-is, or wait to hear back from phaumer?

shin19991207:
@datho7561 I think this PR can be merged as-is. I'll open another PR for the original reported issue when we can actually reproduce and identify the issue.

datho7561:
Sounds good, let's merge, then

wpolicarpo:
@datho7561 apologies for the direct ping, but do we have an ETA for [1.25.0
](https://github.com/redhat-developer/yaml-language-server/milestone/38?closed=1) to be released with this fix? I've been hit by this issue in the last few weeks and just want to keep an eye on the next release. Thanks.

PR #1316, opened 2026-08-05
Fix resolution of percent-encoded schema paths in modelines

shin19991207:
### What does this PR do?
Prevents percent-encoded schema paths in modelines from being encoded a second time when converted to file URIs.

For example, this following percent-encoded modeline should resolve to a schema named `encoded:schema.json`:

```yaml
# yaml-language-server: $schema=./encoded%3Aschema.json
```

Previously, the existing percent escape was encoded again:

```text
./encoded%3Aschema.json
        ↓
file:///.../encoded%253Aschema.json
```

Consequently, the schema request targeted a file literally named `encoded%3Aschema.json` instead of decoding `%3A` to `:`.

Related to https://github.com/redhat-developer/vscode-yaml/issues/1274
This is a separate modeline path case discovered while investigating encoded schema URIs in `yaml.schemas`.

### Is it tested? How?

- [x] New automated tests
- [x] Tested manually:
    1. Created a schema named `encoded:schema.json` (For Windows, use `encoded schema.json` because `:` is invalid in filenames on Windows)
    2. Create a yaml file with a percent-encoded modeline that points to the file `# yaml-language-server: $schema=./encoded%3Aschema.json` (or `# yaml-language-server: $schema=./encoded%20schema.json` for Windows)
    3. Expect the yaml to resolve against the schema correctly

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -483,9 +483,9 @@ export class YAMLSchemaService implements IJSONSchemaService {
       }
       if (!path.isAbsolute(schemaRef)) {
         const resUri = URI.parse(resource);
-        schemaRef = URI.file(path.resolve(path.parse(resUri.fsPath).dir, schemaRef)).toString();
+        schemaRef = URI.file(path.resolve(path.parse(resUri.fsPath).dir, schemaRef)).toString(true);
       } else {
-        schemaRef = URI.file(schemaRef).toString();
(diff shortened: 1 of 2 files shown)


datho7561:
Tried this out on Linux and it was working well; I'll also check it on Windows before approving...

datho7561:
Works on Windows too!

datho7561 (approved):
Thanks, Morgan!

sirgagahandino946-afk (approved):

Issue #766, opened 2022-08-11
Editor shows syntax errors for helm charts with flow control statements (if/else/with/range)


MahatmaFatalError:
## Describe the bug
Eclipse editor shows errors for flow control statements https://helm.sh/docs/chart_template_guide/control_structures/ 

## Expected Behavior
Editor opens the file, no syntax errors.

## Current Behavior
Editor opens the file, but shows syntax errors for lines with helm flow control statements:
```
Multiple markers at this line
- Block collections are not allowed within flow collections
- Missing , or : between flow map items
- Implicit map keys need to be followed by map values
- Implicit keys need to be on a single line
```

## Steps to Reproduce
Open
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: {{ template "fullname" . }}
  labels:
    app: {{ template "name" . }}
    chart: {{ .Chart.Name }}
    {{- if .Values.something.enabled }}
    something: "true"       
    {{- end}}
```
in eclipse generic text editor (wild wild web developer).
Errors in line 8 and 10.

## Environment
- [ ] Windows
- [x] Mac
- [ ] Linux
- [ ] other (please specify)

gorkem:
@mickaelistria Is this relevant to `yaml-language-server`?

mickaelistria:
Yes. Eclipse IDE embed yaml-language-server, and opening the file makes the language server return the described errors.

Issue #449, opened 2021-04-27
Helm conditional and loop expressions


derba:
Helm, the leading cloud native application delivery and configuration management solution, has its templating syntax. I think its support could be improved.

Lets consider this file, copied from [here](https://helm.sh/docs/chart_template_guide/control_structures/):

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ .Release.Name }}-configmap
data:
  myvalue: "Hello World"
  drink: {{ .Values.favorite.drink | default "tea" | quote }}
  food: {{ .Values.favorite.food | upper | quote }}
  {{- if eq .Values.favorite.drink "coffee" }}
  mug: true
  {{- end }}
```
In VS Code on the line of `if` the following problem is printed: "missed comma between flow collection entries".
On the next line with "mug" this appears: "end of the stream or a document separator is expected".
Both are incorrect.

Sometimes, for the sake of easier readability, the conditional and loop lines are not indented according to YAML. E.g:

```yaml
data:
{{- if eq .Values.favorite.drink "coffee" }}
  mug: true
{{- else }}
  cup: true
{{- end }}
```

Probably the easiest solution would be ignoring anything between `{{` and `}}`.

JPinkney:
Helm isn't valid yaml. If you want to use helm support in vscode use https://marketplace.visualstudio.com/items?itemName=ms-kubernetes-tools.vscode-kubernetes-tools. It interfaces with this extension and has its own built-in helm parser IIRC

derba:
You are right. Helm is not a valid YAML. That is why I thin this is not a bug, but a possible improvement.

Telling the truth, I am a bit confused, as Kubernetes plugin depends on this YAML plugin.

Sad fact: If I enable Kubernetes plugin, then JSON schema validation does not work any longer. That is of course an improvement request for Kubernetes plugin.

Issue #220, opened 2020-01-21
Support for .tpf (Helm chart Yaml-based template files)


mickaelistria:
It seems like Helm Chart files are basically Kubernetes yaml files with templating capability in `{{...}}` blocks.
It would be nice if Yaml Language Server could provide support for those files (by pre-setting schema, allowing validation/completion in `{{...}}` and so on.

nklmilojevic:
What is needed for this to be added - how we can help? There is https://github.com/Azure/vscode-kubernetes-tools/blob/master/syntaxes/helm.tmLanguage.json which VSCode uses to correctly highlight Helm template syntax.

valorl:
I was recently looking into improving working with Helm templates in my neovim setup and came across this tree-sitter grammar for go templates. 

Could be relevant here. With tree-sitter and this grammar, the template blocks can be fully parsed into an AST that could be useful for functionality beyond highlighting, such as diagnostics, autocomplete, etc.

https://github.com/ngalaiko/tree-sitter-go-template

luisdavim:
It would be nice to make this work with templates in general and not just helm templates, `{{...}}` is the default delimiter set but it can be customised, and we also have Jinja and ERB....

Kerwood:
Would be nice to see this feature.

luisdavim:
There is there now: https://github.com/mrjosh/helm-ls

luisdavim:
This is the setup I landed on, install these 2 plugins:
* https://github.com/mrjosh/helm-ls - helm language server
* https://github.com/towolf/vim-helm - adds syntax higlighting for helm templates and sets the file type to `helm` (this is the important part)

Then configure the language server to run helm-ls for helm files, in my case, I use https://github.com/neoclide/coc.nvim so I have:

```jsonc
"languageserver": {
  "helm": {
    "command": "helm_ls",
    "args": ["serve"],
    "filetypes": ["helm", "helmfile"],
    "rootPatterns": ["Chart.yaml"]
  }
}
```

This solves the problems with all the diagnostics errors from yaml ls but still uses it under the covers if installed.

UgaTheDev:
This issue is six years old, so before proposing anything, here is what has already been tried and why it went nowhere:

- Issue #53 and Issue #449 asked for Helm support and were both closed with the same reasoning: "helm yaml is not compatible to yaml parser". The suggested alternative was vscode-kubernetes-tools and its own Helm parser.
[...]

Issue #53, opened 2018-03-15
Is it possible to support Helm.sh templating in option please?


lakano:
Hi !
Helm is now the official package manager of Kubernetes. K8s configuration is a lot based under yaml files, and Helm have a templating system to help to replace values in the yaml configuration, but also to control the flow ( {{ if ... }} {{ end }} ).
Of course, it's not YAML compatible ( at least fo the flow control https://docs.helm.sh/chart_template_guide/#flow-control ).
May be it's possible to bypass the errors/warning, based on the extension (eg: file.helm) ?
Regards,

andxu:
in fact, the helm yaml is not a yaml compatible to yaml parser, I will do the parse work(helm languange server) which will be used in vscode kubernetes  tools.

andxu:
@JPinkney yaml-language-server has not been published for over 2 months, when is the planed release date for the next publish(see https://www.npmjs.com/package/yaml-language-server)?

JPinkney:
@andxu Next release will hopefully be in 10 days. We just need to close a few issues first.

JPinkney:
Going to close this issue because helm yaml is not compatible to yaml parser.

PR #1321, opened 2026-08-13
Add opt-in masking of template expressions before parsing

UgaTheDev:
### What does this PR do?

Adds an opt-in `yaml.template` setting. When set to `helm`, Go-template expressions are masked with inert text of exactly the same length before the document is parsed, so templated files such as Helm charts parse as plain YAML instead of reporting a cascade of syntax errors.

This implements the design I proposed in Issue #220. It deliberately does not do either of the two things previous attempts asked for and maintainers declined: it teaches the server no Helm semantics, and it does not touch or swap the YAML parser.

**How masking works**

- A line whose content is only a template expression (`{{- if .Values.enabled }}`, `{{- end }}`) becomes a comment of identical length, so control flow drops out of the document structure.
- An inline expression (`replicas: {{ .Values.replicaCount }}`) becomes a plain scalar of identical length, so the value parses as a string.
- A span crossing line boundaries is masked on every line it covers, each line classified by the rules above. An unclosed span is masked to the end of its line.

Because every replacement is length-preserving, all offsets survive. Diagnostics, hover, completion, and symbol ranges map back to the real document with no position-translation layer anywhere.

**Where it hooks in**

Masking happens in `YamlDocuments.ensureCache`, the single choke point that validation, completion, hover, folding, symbols, code lens, links, and definition all read through. The cache-invalidation check gained the template mode so that changing the setting re-parses open documents.

One design note worth flagging for review: the mode is stored on `YamlDocuments` rather than added to `ParserOptions`. Most callers of `getYamlDocument` pass no options at all and get `defaultOptions`; only validation and completion pass real ones. Since the cache is keyed by URI alone, threading the mode through `ParserOptions` would have made masking depend on which feature happened to fill the cache first. Happy to change this if you would rather it live somewhere else.

The formatter is the one feature that bypasses the cache, so it returns no edits for documents containing template expressions when the setting is on. Prettier has no notion of template syntax and would rewrite or reject it.

**Honest limits**

- No completion, hover, or validation inside `{{ }}`. That stays helm-ls territory. This only stops template syntax from breaking the document around it.
- A fully templated value validates as a string, so a schema expecting a number or boolean at that position still reports a type error. This could later be handled by suppressing type diagnostics whose range overlaps a masked span, but I left that out to keep this change small.
- `{{ include "chart.labels" . | nindent 4 }}` expanding to block content, and `if`/`else` branches that each define the same key, can still produce false positives. Masking cannot know what a template expands to.
- Multi-line expressions are masked line by line. Positions stay correct; the intermediate text is just uglier.

I did not add a CHANGELOG entry, since there is no `1.25.0` section yet and I did not want to create the heading. Happy to add one wherever you prefer.

Implements the proposal in Issue #220. Also covers the parsing half of Issue #766. Related closed issues: Issue #53, Issue #449.

### Is it tested? How?

Yes, `test/templateMasking.test.ts` adds 11 tests. The full suite is green at 1319 passing, 0 failing, and `npm run build` (clean, lint, compile, UMD, ESM) exits 0.

The tests include the before/after pair directly, so the fix cannot silently regress into a no-op. On this chart:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: {{ include "mychart.fullname" . }}
  labels:
    {{- include "mychart.labels" . | nindent 4 }}
spec:
  {{- if not .Values.autoscaling.enabled }}
  replicas: {{ .Values.replicaCount }}
  {{- end }}
  selector:
    matchLabels:
      app: {{ .Chart.Name }}
```

with `yaml.template: none`, the parser reports 10 errors:

```
Block collections are not allowed within flow collections
Missing , or : between flow map items
Block collections are not allowed within flow collections
Missing , or : between flow map items
All mapping items must start at the same column
All mapping items must start at the same column
Implicit keys need to be on a single line
Block collections are not allowed within flow collections
Missing , or : between flow map items
Implicit map keys need to be followed by map values
```

with `yaml.template: helm`, it reports 0.

Other cases covered: text without templates is returned unchanged (identity, not a rebuild); length and line count are preserved; indentation is preserved on commented-out lines; spans crossing line boundaries and unclosed spans; quoted expressions; offsets still resolve to the correct source text (`kind: Deployment` is located by node offset against the original, unmasked document); and changing the setting re-parses a document whose version has not changed.

To try it manually, set `yaml.template` to `helm` and open any chart under `templates/`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ The server supports the following settings supplied by LSP clients:
 
 - `yaml.yamlVersion`: Set default YAML spec version (`1.2` or `1.1`). Defaults to `1.2`.
 - `yaml.maxItemsComputed`: The maximum number of document symbols and folding regions computed (limited for performance reasons). Defaults to `5000`.
+- `yaml.template`: Mask templating expressions before parsing, so that templated documents such as Helm charts parse as plain YAML instead of reporting syntax errors on every `{{ ... }}`. Set to `helm` to enable, `none` to disable. Defaults to `none`. See [Templated documents](#templated-documents).
 - `yaml.format.enable`: Enable/disable the default YAML formatter. Defaults to `true`.
 - `yaml.format.singleQuote`: Use single quotes instead of double quotes. Defaults to `false`.
 - `yaml.format.bracketSpacing`: Print spaces between brackets in objects. Defaults to `true`.
@@ -292,6 +293,24 @@ When multiple schema sources or schema-disabling mechanisms apply to the same YA
 5. `yaml.schemas`
 6. `json/schemaAssociations` notification
 7. SchemaStore
+
+## Templated documents
+
+Templating languages that wrap YAML, such as Helm's Go templates, produce files that are not valid YAML until they are rendered. A control-flow line like `{{- if .Values.autoscaling.enabled }}` breaks the document structure, and the parser reports a cascade of syntax errors on every line that follows.
+
+Setting `yaml.template` to `helm` masks `{{ ... }}` expressions with inert text of exactly the same length before the document is parsed:
+
+- A line containing only a template expression becomes a comment, so control flow drops out of the document structure.
+- An inline expression such as `replicas: {{ .Values.replicaCount }}` becomes a plain scalar, so the value parses as a string.
+
+Because masking preserves length, all offsets are unchanged, and diagnostics, hover, completion, and symbol ranges continue to point at the real document. Schema validation, duplicate key detection, completion, hover, folding, and document symbols keep working on the untemplated parts of the file.
+
+The setting is off by default. Its limits are worth knowing before enabling it:
+
+- There is no completion, hover, or validation inside `{{ }}`. Masking only stops template syntax from breaking the surrounding document.
+- A fully templated value validates as a string, so a schema expecting a number or boolean at that position still reports a type error.
+- Expressions that expand to block content, such as `{{ include "chart.labels" . | nindent 4 }}`, and `if`/`else` branches that each define the same key, can still produce false positives. Masking cannot know what a template expands to.
+- Formatting is disabled for documents containing template expressions, since the formatter would rewrite the template syntax.
 
 ## Adding custom tags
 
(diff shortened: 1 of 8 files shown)


datho7561:
This seems like a promising approach for validating simple Helm charts, but I think there's some pretty hard limitations to what we can do based on what you've described. Complex charts (anything with if/else) might require a different approach to get working, and providing Helm-specific features like completion doesn't seem possible without a node in the AST to represent the Helm template segment.

I found this case that doesn't validate properly:

```helm
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ .Release.Name }}-configmap
data:
  myvalue: "Hello World"
  drink: {{ .Values.favorite.drink | default "tea" | quote }}
  food: {{ .Values.favorite.food | upper | quote }}
  {{ if eq .Values.favorite.drink "coffee" }}mug: "true"{{ end }}
```

My guess is for this and similar cases, we'll need to be a bit more aware of how Helm works and apply that logic when masking/stripping the template. Do you mind seeing if you can get that case working?

Something good about this approach is that we can likely expand it to cover Jinja, a templating language used in Python, too.

UgaTheDev:
Good catch — the inline `{{ if }}…{{ end }}` line was taking the inline-filler path, which left filler after the closing quote and broke the parse. Fixed in 4508ed8: lines that start with a control-flow expression are now commented out like standalone control-flow lines, since their content is conditional and can't be validated as always-present anyway. Added your ConfigMap as a test case.

Agreed on Jinja — the masking pass is dialect-agnostic apart from the span regex, so it should slot in as another `TemplateMode`.

One known limitation worth stating: diagnostics *inside* a conditional line disappear (a typo in `mug:` won't be flagged) — validating conditional content properly would need the AST-level Helm awareness you mentioned.

UgaTheDev:
Checking in on this after a few quiet weeks. CI is green and the masking remains opt-in (default off), so current behavior is unchanged unless the setting is enabled. Happy to rebase, rename the setting, or adjust scope if the approach needs discussion.

PR #1323, opened 2026-08-13
Fix enum and const validation for nested boolean values

### LINKED ISSUE #1320: `getNodeValue()` reads raw source text for booleans, breaking `const`/`enum` validation (e.g. `enum: [true]` rejects a valid `true`)
A boolean value compared against a `const` or `enum` schema constraint is read from its raw YAML source text rather than its parsed value, causing valid documents to be rejected.

### Reproduction

Schema:
```json
{
  "type": "object",
  "properties": {
    "required": { "type": "boolean", "enum": [true] }
  }
}
```

YAML:
```yaml
required: true
```

**Expected:** no diagnostics — `true` satisfies `enum: [true]`.
**Actual:** `Value is not accepted. Valid values: true.` — a false positive on a value that already matches the constraint.

This isn't synthetic: it's the exact shape OpenAPI 3.0's official schema uses to require `required: true` on path parameters (`definitions.ParameterLocation`, discriminated `oneOf` with `enum: [true]`), so any OAS 3.0 document with a valid path parameter fails validation.

### Root cause

`getNodeValue()` in `lib/esm/languageservice/parser/astNodeUtils.js` returns the node's raw source text for booleans instead of its parsed value:

```js
export function getNodeValue(node) {
  switch (node.type) {
    ...
    case 'null':
    case 'string':
    case 'number':
      return node.value;
    case 'boolean':
      return node.source;   // <-- returns the string "true", not the boolean true
    default:
      ...
  }
}
```

Note the adjacent `'string'`/`'number'` cases already correctly return `node.value`; only `'boolean'` returns `.source` instead. So `required: true` is compared against the constraint as the string `"true"`, which fails a strict `===`/`deepEqual` check against the boolean `true` in `enum: [true]`.

Confirmed present in `yaml-language-server@1.23.0`, and downstream in every `monaco-yaml` version that bundles it (verified in `monaco-yaml@5.4.0` and `5.5.0`, where this function is inlined into the built worker under the name `getNodeValue3`).

### Fix

```diff
     case 'boolean':
-      return node.source;
+      return node.value;
```

`node.value` already carries the correctly-parsed boolean at this point — `.source` is the raw text, kept around for other purposes (e.g. formatting/round-tripping), not for value comparison.

### Downstream impact

This also affects `monaco-yaml`, which bundles this function into its worker build — filed there as well: <link once monaco-yaml issue is filed>.

### Environment

- `yaml-language-server`: 1.23.0 (also present in earlier versions going back at least to what `monaco-yaml@5.4.0` bundles)

shin19991207:
### What does this PR do?

Updates `getNodeValue()` to return the parsed value for boolean nodes instead of their raw YAML source

Previously, booleans nested inside object or array `enum`/`const` values were reconstructed as strings. For example, the YAML value:

```yaml
parameter:
  required: true
```

was compared as `{ "required": "true" }` against the schema value `{ "required": true }`. This caused the diagnostic `Value is not accepted. Valid values: {"required":true}.`.

Fixes https://github.com/redhat-developer/yaml-language-server/issues/1320

### Is it tested? How?

- [x] New automated tests by codex
- [x] Manual test: https://github.com/redhat-developer/yaml-language-server/issues/1320#issuecomment-5285388916

### DIFF
--- a/src/languageservice/utils/astNodeUtils.ts
+++ b/src/languageservice/utils/astNodeUtils.ts
@@ -21,7 +21,7 @@ export function getNodeValue(node: ASTNode): any {
     case 'number':
       return node.value;
     case 'boolean':
-      return node.source;
+      return node.value;
     default:
       return undefined;
   }
--- a/test/schemaValidation.test.ts
+++ b/test/schemaValidation.test.ts
@@ -208,6 +208,72 @@ describe('Validation Tests', () => {
       assert.deepStrictEqual(result, []);
     });
 
+    it('Test object enum value containing a nested boolean', async () => {
+      schemaProvider.addSchema(SCHEMA_ID, {
+        type: 'object',
+        properties: {
+          parameter: {
+            type: 'object',
+            enum: [
+              {
+                required: true,
+              },
+            ],
+          },
+        },
+      });
+      const content = 'parameter:\n  required: true';
+      const result = await parseSetup(content);
+      assert.deepStrictEqual(result, []);
+    });
+
+    it('Test array enum value containing a nested boolean', async () => {
+      schemaProvider.addSchema(SCHEMA_ID, {
+        type: 'object',
+        properties: {
+          flags: {
+            type: 'array',
+            enum: [[true]],
+          },
+        },
+      });
+      const content = 'flags:\n  - true';
+      const result = await parseSetup(content);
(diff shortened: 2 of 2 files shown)


datho7561 (approved):
Looks good and fixes the specific case you mentioned. Thanks, Morgan!

PR #1195, opened 2026-02-25
Fix infinite $ref resolution loops

shin19991207:
### What does this PR do?
yamlSchemaService could keep expanding references forever and crash with heap/stack OOM when resolving schema with cyclic references.

Original implementation introduces a bug when $ref + siblings are rewritten to allOf and creates new nodes. This can cause dynamic ref paths to revisit the same resource repeatedly.

Fixed by:
- Added sibling cycle tracking in traversal state in `WalkItem.siblingRefCycleKeys`
- Stop rewriting when the same sibling-ref target is seen again
- Added a helper to short-circuit when a target resource is already in resolutionStack
- Changed local merge cloning to only clone for local $dynamicRef with dynamic scope

### What issues does this PR fix or reference?
Fixes https://github.com/redhat-developer/yaml-language-server/issues/1190

### Is it tested? How?

Automated test and manually with this example:
```
# yaml-language-server: $schema=https://raw.githubusercontent.com/OAI/spec.openapis.org/refs/heads/main/oas/3.1/schema/2025-11-23
openapi: "3.1.0"
info:
  version: 1.0.0
  title: Minimal ping API server
paths:
  /ping:
    get:
      responses:
        "200":
          description: pet response
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Pong"
components:
  schemas:
    # base types
    Pong:
      type: object
      required:
        - ping
      properties:
        ping:
          type: string
          example: pong
```
When triggering auto-completion, expected to see 'no suggestion' instead of no response (keep loading).

PR #1326, opened 2026-08-18
Fix: keep constraints on cyclic cross-file $refs in unindexed resources

### LINKED ISSUE #1325: Recursive `$ref` to a schema's own filename silently disables validation of nested values
## Describe the bug

When a schema recurses through a `$ref` that names a **file** (rather than a local `#/...` pointer), everything below the first level of recursion stops being validated. No error is reported — the nested subschema resolves to `{}`, which accepts anything.

The top level is still validated correctly, which makes this easy to miss: a schema looks like it is working.

This is a regression. It was introduced by #1195 and is present in 1.21.0 through 1.24.0 and on `main`.

## Expected Behavior
(See steps to Reproduce): Two errors, `Property bogus is not allowed.`, on line 3 and line 6.

## Current Behavior

Only **one** error, on line 3. The nested `bogus: 2` on line 6 is silently accepted.

Anything nested one or more levels deep is unvalidated — the depth does not matter, only that it is below the first recursion.

## Steps to Reproduce
Two files in the same folder.

**`node.json`** — note that `children.items` refers to this very file by name:

```json
{
  "type": "object",
  "properties": {
    "name": { "type": "string" },
    "children": {
      "type": "array",
      "items": { "$ref": "node.json" }
    }
  }
}
```

**`test.yaml`**:

```yaml
# yaml-language-server: $schema=node.json
name: root
bogus: 1
children:
  - name: child
    bogus: 2
```

`additionalProperties` is `false`, so **both** `bogus` keys are invalid.

## Environment
- yaml-language-server version: reproduces on 1.21.0 – 1.24.0 and `main`
- Last known good version: 1.20.0

- [x] Windows
- [ ] Mac
- [ ] Linux
- [ ] other (please specify)

## Cause

`_mergeIfResourceAlreadyInResolutionStack()` in `src/languageservice/services/yamlSchemaService.ts` returns `true` — "I have handled this `$ref`" — whenever the target resource is on the resolution stack, *even when `resourceIndexByUri` holds no entry for that resource*:

```ts
const _mergeIfResourceAlreadyInResolutionStack = (ref, resolvedResource, frag) => {
  if (!resolutionStack.has(resolvedResource)) return false;
  if (!seenRefs.has(ref)) {
    const source = resourceIndexByUri.get(resolvedResource)?.root;
    if (source && typeof source === 'object') {   // <-- skipped when the resource is not indexed
      _merge(next, source, resolvedResource, frag, !!recursiveAnchorBase);
    }
    seenRefs.add(ref);
  }
  return true;                                     // <-- but the $ref is claimed anyway
};
```

When the resource is not indexed, nothing is merged. By that point the caller has already deleted `next.$ref`, and returning `true` makes it `continue`, skipping `resolveExternalLink()`. The subschema is therefore left empty.

`resourceIndexByUri` is only populated for resources identifiable via `$id`/anchors. Schemas that just reference each other by relative file path — normal for draft-07 — never appear in it, so for them *every* recursive file `$ref` degrades into an empty, always-valid schema.

Logging at the point where the guard fires, for the reproducer above:

```
ref=node.json  inStack=true  indexHasKey=false  sourceType=undefined
index keys: []
```

## Suggested fix

Only claim the `$ref` when there is actually something to merge; otherwise fall through to the regular `resolveExternalLink()` path, as before #1195:

```diff
   if (!resolutionStack.has(resolvedResource)) return false;
+  const source = resourceIndexByUri.get(resolvedResource)?.root;
+  if (!source || typeof source !== 'object') return false;
   if (!seenRefs.has(ref)) {
-    const source = resourceIndexByUri.get(resolvedResource)?.root;
-    if (source && typeof source === 'object') {
-      _merge(next, source, resolvedResource, frag, !!recursiveAnchorBase);
-    }
+    _merge(next, source, resolvedResource, frag, !!recursiveAnchorBase);
     seenRefs.add(ref);
   }
   return true;
```

The cycle short-circuit still applies whenever the resource *is* indexed, so the infinite `$ref` resolution loops fixed in #1195 stay fixed — I verified that #1195's own regression test (`does not infinite loop on cyclic $ref with siblings`) still passes, and that 1.20.0 hangs on it, so reverting #1195 is not an alternative.

I have this change ready with a regression test and can open a PR.

AndreasReichel:
Fixes #1325.

## Problem

When a schema recurses through a `$ref` that names a **file** rather than a local `#/...` pointer, everything below the first level of recursion silently stops being validated. The nested subschema resolves to `{}`, which accepts anything, so no error is reported and the schema appears to work.

A single self-referencing file is enough to reproduce it (full reproducer in #1325):

```json
{
  "type": "object",
  "properties": {
    "name": { "type": "string" },
    "children": { "type": "array", "items": { "$ref": "node.json" } }
  }
}
```

The same recursion expressed as `#/$defs/node` behaves correctly, so only the file-name form is affected.

Regression from #1195, present in 1.21.0 through 1.24.0 and on `main`.

## Cause

`_mergeIfResourceAlreadyInResolutionStack()` returns `true` — meaning "I have handled this `$ref`" — whenever the target resource is on the resolution stack, *even when `resourceIndexByUri` holds no entry for that resource*:

```ts
if (!resolutionStack.has(resolvedResource)) return false;
if (!seenRefs.has(ref)) {
  const source = resourceIndexByUri.get(resolvedResource)?.root;
  if (source && typeof source === 'object') {   // <- skipped when the resource is not indexed
    _merge(next, source, resolvedResource, frag, !!recursiveAnchorBase);
  }
  seenRefs.add(ref);
}
return true;                                     // <- but the ref is claimed anyway
```

Doing the work is conditional; claiming it is not. When the resource is not indexed nothing gets merged, and because the caller has already deleted `next.$ref` and treats `true` as "skip `resolveExternalLink()`", the subschema is left empty.

`resourceIndexByUri` is only populated for resources identifiable via `$id`/anchors. Schemas that reference each other by relative file path — normal for draft-07 bundles — never appear in it, so for them every recursive file `$ref` degrades into an empty, always-valid schema. Logged at the point the guard fires:

```
ref=node.json  inStack=true  indexHasKey=false  sourceType=undefined
index keys: []
```

## Fix

Only claim the `$ref` when there is actually something to merge; otherwise fall through to the regular `resolveExternalLink()` path as before #1195. Reaching `_merge` is now the guarantee rather than a conditional:

```diff
   if (!resolutionStack.has(resolvedResource)) return false;
+  const source = resourceIndexByUri.get(resolvedResource)?.root;
+  if (!source || typeof source !== 'object') return false;
   if (!seenRefs.has(ref)) {
-    const source = resourceIndexByUri.get(resolvedResource)?.root;
-    if (source && typeof source === 'object') {
-      _merge(next, source, resolvedResource, frag, !!recursiveAnchorBase);
-    }
+    _merge(next, source, resolvedResource, frag, !!recursiveAnchorBase);
     seenRefs.add(ref);
   }
   return true;
```

The cycle short-circuit still applies whenever the resource *is* indexed, so the infinite `$ref` resolution loops fixed in #1195 stay fixed.

## Testing

Added `should keep constraints on cyclic cross-file $refs when the resources have no $id` to `test/yamlSchemaService.test.ts`. It asserts that the node closing the cycle keeps its `type`, `properties`, `required` and `additionalProperties`. It fails before this change (`cycle target lost its type: expected undefined to equal 'object'`) and passes after.

Verified across three shapes, comparing 1.20.0 (before #1195), 1.24.0 as released, and 1.24.0 with this fix:

| case | 1.20.0 | 1.24.0 | 1.24.0 + fix |
| --- | --- | --- | --- |
| cyclic `$ref` with siblings (the #1195 case) | hangs | pass | pass |
| multi-file cycle **with** `$id` | pass | pass | pass |
| multi-file cycle, relative refs, no `$id` (#1325) | pass | **fail** | pass |

Note the first row: reverting #1195 is not an alternative, since 1.20.0 hangs on the case it fixed. Its regression test (`does not infinite loop on cyclic $ref with siblings`) still passes with this change.

The full suite produces an identical set of failures before and after, plus the one new passing test (the pre-existing failures are Kubernetes/SchemaStore tests that need network access I do not have):

- without the fix: 1272 passing, 24 failing
- with the fix: 1273 passing, 23 failing

## Notes

Found while debugging a real draft-07 schema bundle split into one file per type, whose recursive `elements` property stopped reporting diagnostics for nested entries after a language-server update.

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -1161,11 +1161,10 @@ export class YAMLSchemaService implements IJSONSchemaService {
 
         const _mergeIfResourceAlreadyInResolutionStack = (ref: string, resolvedResource: string, frag: string): boolean => {
           if (!resolutionStack.has(resolvedResource)) return false;
+          const source = resourceIndexByUri.get(resolvedResource)?.root;
+          if (!source || typeof source !== 'object') return false;
           if (!seenRefs.has(ref)) {
-            const source = resourceIndexByUri.get(resolvedResource)?.root;
-            if (source && typeof source === 'object') {
-              _merge(next, source, resolvedResource, frag, !!recursiveAnchorBase);
-            }
+            _merge(next, source, resolvedResource, frag, !!recursiveAnchorBase);
             seenRefs.add(ref);
           }
           return true;
--- a/test/yamlSchemaService.test.ts
+++ b/test/yamlSchemaService.test.ts
@@ -305,6 +305,53 @@ describe('YAML Schema Service', () => {
       });
     });
 
+    it('should keep constraints on cyclic cross-file $refs when the resources have no $id', async () => {
+      const content = `# yaml-language-server: $schema=file:///schemas/node.json\nname: root\nchildren:\n  - name: child`;
+      const yamlDock = parse(content);
+
+      // draft-07, split across files, relative $refs, no $id anywhere:
+      // node.json -> wrapper.json -> node.json
+      const nodeSchema = {
+        $schema: 'http://json-schema.org/draft-07/schema#',
+        type: 'object',
+        required: ['name'],
+        additionalProperties: false,
+        properties: {
+          name: { type: 'string' },
+          children: { type: 'array', items: { $ref: 'wrapper.json' } },
+        },
+      };
+      const wrapperSchema = {
+        $schema: 'http://json-schema.org/draft-07/schema#',
+        oneOf: [{ $ref: 'node.json' }],
+      };
+
+      requestServiceMock = sandbox.fake((uri: string) => {
+        if (uri === 'file:///schemas/node.json') {
+          return Promise.resolve(JSON.stringify(nodeSchema));
+        }
+        if (uri === 'file:///schemas/wrapper.json') {
+          return Promise.resolve(JSON.stringify(wrapperSchema));
+        }
+        return Promise.reject<string>(`Resource ${uri} not found.`);
+      });
+
+      const service = new SchemaService.YAMLSchemaService(requestServiceMock, workspaceContext);
+      const schema = await service.getSchemaForResource('', yamlDock.documents[0]);
+
+      const children = schema.schema.properties.children as JSONSchema;
+      const wrapper = children.items as JSONSchema;
+      expect(wrapper.oneOf, 'wrapper lost its oneOf branches').to.have.length(1);
+
+      // This is the node that closes the cycle back to node.json. An empty schema
+      // here would silently accept anything nested under `children`.
+      const cycleTarget = wrapper.oneOf[0] as JSONSchema;
+      expect(cycleTarget.type, 'cycle target lost its type').to.equal('object');
+      expect(cycleTarget.properties, 'cycle target lost its properties').to.have.property('name');
+      expect(cycleTarget.required, 'cycle target lost its required list').to.eql(['name']);
+      expect(cycleTarget.additionalProperties, 'cycle target lost additionalProperties').to.equal(false);
+    });
+
     it('should resolve nested local sibling refs relative to the loaded sibling schema file', async () => {
       const content = `# yaml-language-server: $schema=file:///schemas/primary.json\nitem: ok`;
       const yamlDock = parse(content);


datho7561 (approved):
Seems good from my initial look at it, but I don't know this code that well.

@shin19991207 do you mind reviewing as well?

shin19991207 (approved):
Looks good and works well from my testing!

PR #1336, opened 2026-09-11
Fix schema references through non-standard containers

### LINKED ISSUE #1322: $ref '/components/schemas/pipelines_configuration' in 'https://bitbucket.org/product/features/pipelines' cannot be resolved.YAML(65536)
## Describe the bug
The schema used in the BitBucket Pipelines configuration is not parsed correctly: 
https://api.bitbucket.org/schemas/pipelines-configuration

The schema file is valid and JetBrains IDEs are able to load it properly.

## Expected Behavior
I would expect that the schema is successfully loaded and the YAML file is validated against it.

## Current Behavior
Error: `$ref '/components/schemas/pipelines_configuration' in 'https://bitbucket.org/product/features/pipelines' cannot be resolved.YAML(65536)`

## Steps to Reproduce

1. configure vscode to use the schema:
```json
{
    "yaml.schemas": {
        "https://api.bitbucket.org/schemas/pipelines-configuration": "bitbucket-pipelines.yml"
    }
}
````
2. Create a simple `bitbucket-pipelines.yml` file and open it, e.g.:
```yaml
pipelines:
  default:
    - step:
        name: Say Hello
        script:
          - echo "Hello, world!"
```
 
## Environment
- [x] Windows
- [x] Mac
- [x] Linux
- [ ] other (please specify)

shin19991207:
### What does this PR do?

Fixes `$ref` resolution when a root reference points into a non-standard schema container, such as `#/components/schemas/pipelines_configuration`, instead of a standard container such as `$defs` or `definitions`. The fix also covers nested references that target the same local schema file or a sibling file.

In Draft-07 and earlier, spec mentions that keywords alongside `$ref` should be ignored. During resolution, `YAMLSchemaService` removed constraint sibling keys, including non-standard schema containers. This made the referenced JSON Pointer through non-standard containers unavailable for subsequent lookups.

Schemas reached through these containers could also lose their local source URI. As a result, relative references such as `./schema.json` could be resolved against the schema’s remote `$id` instead of the local file location.

The fix preserves root keys for JSON Pointer lookup, indexes schemas by their retrieval URI as well as their `$id`, and propagates the local source URI through nested reference resolution.

Fixes https://github.com/redhat-developer/yaml-language-server/issues/1322

### Is it tested? How?

- Added new tests for non-standard container lookups, local self-references, and local sibling-file references
- Tested manually with the cases in new tests

### DIFF
--- a/src/languageservice/services/yamlSchemaService.ts
+++ b/src/languageservice/services/yamlSchemaService.ts
@@ -737,6 +737,14 @@ export class YAMLSchemaService implements IJSONSchemaService {
     type PlainNameFragmentMap = Map<string, FragmentEntry>;
     type ResourceIndex = { root?: JSONSchema; fragments: PlainNameFragmentMap };
     const resourceIndexByUri = new Map<string, ResourceIndex>();
+    const lookupRoots = new WeakMap<JSONSchema, JSONSchema>();
+
+    // Preserve root keys before $ref processing removes or rewrites siblings
+    const _preserveLookupRoot = (root: JSONSchema): void => {
+      if (root && typeof root === 'object' && !lookupRoots.has(root)) {
+        lookupRoots.set(root, { ...root });
+      }
+    };
 
     const _getResourceIndex = (resourceUri: string): ResourceIndex => {
       let entry = resourceIndexByUri.get(resourceUri);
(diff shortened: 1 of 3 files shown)


datho7561 (approved):
Works really well. Thanks, Morgan!

datho7561:
Ooops sorry I forgot to merge.

PR #1338, opened 2026-09-14
fix: report the underlying reason when a schema fails to load

### LINKED ISSUE #1337: bug: Every schema load failure is reported as "No content", hiding the real cause
## Describe the bug

When a remote schema cannot be fetched, the diagnostic always reads:

```
Unable to load schema from '<url>': No content.
```

This message is emitted regardless of why the fetch failed. An HTTP 429, an HTTP 403, a dropped connection, and a genuinely empty response body all produce identical text, so the message carries no information about what actually went wrong.

## Expected Behavior

The diagnostic should report why the schema could not be loaded, for example:

```
Unable to load schema from '<url>': Request failed with status code 429.
```

`No content.` should be reserved for a response that genuinely has no content.

## Current Behavior

Every failure mode collapses to the same text, injecting failures through `schemaRequestService`:

| Injected failure | Message |
| --- | --- |
| Throws `Request failed with status code 429` | `Unable to load schema from '<url>': No content.` |
| Throws `Request failed with status code 403` | `Unable to load schema from '<url>': No content.` |
| Throws `socket hang up` (`ECONNRESET`) | `Unable to load schema from '<url>': No content.` |
| Resolves `''` (genuinely empty) | `Unable to load schema from '<url>': No content.` |

Only the last row is accurate; the first three report something that did not happen.

## Steps to Reproduce

1. Create a language service whose `schemaRequestService` rejects with a realistic network error:

```ts
const service = getLanguageService({
  schemaRequestService: async () => {
    throw new Error('Request failed with status code 429');
  },
  workspaceContext: { resolveRelativePath: (r, b) => new URL(r, b).toString() },
});

service.configure({
  validate: true,
  schemas: [{ uri: 'https://example.com/ci.json', fileMatch: ['*.yml'], priority: 0 }],
  customTags: [],
  yamlVersion: '1.2',
});
```

2. Validate any YAML document matching `*.yml`.
3. Observe the diagnostic reads `... : No content.` rather than naming the 429.

## Environment
- [x] Windows
- [x] Mac
- [x] Linux
- [ ] other (please specify)

X-Guardian:
### What does this PR do?

Stops `loadSchema` from overwriting an accurate schema-load error with the generic `No content.` message, so failures report their actual cause:

```
Before:  Unable to load schema from '<url>': No content.
After:   Unable to load schema from '<url>': Request failed with status code 429.
```

`loadSchema` already composes a correct message for every load failure, including the HTTP status, and encodes that status into the error code. A later branch in the same method then replaced it:

```ts
if (errorMessage.toLowerCase().indexOf('load') !== -1) {
  errorMessage = l10n.t("Unable to load schema from '{0}': No content.", toDisplayString(schemaUri));
}
```

Because the correct message begins with `"Unable to **load** schema from"`, it always satisfied that substring test, so a well-formed message was guaranteed to be discarded by the branch intended to improve it.

The branch is removed rather than narrowed because it no longer has anything to contribute. `loadSchema` emits `No content.` itself when the response body is genuinely empty, so every message the branch could produce is already produced upstream of it — it could only ever lose information. It looks like a leftover from before `loadSchema` was reworked to build its own messages.

The adjacent `parse` branch is deliberately left untouched: it re-fetches the schema to turn a raw offset into a line and column, which is still a genuine improvement over the message it replaces. Only the `load` case is removed, and the `if` / `else if` collapses to a single `if`.

Behaviour is otherwise unchanged. `No content.` is still reported for a genuinely empty body, so the message now carries meaning instead of acting as a catch-all. The numeric diagnostic `code` was already correct and is unaffected.

This is user-visible text, so anything asserting on the exact string `No content.` for a *failed request* will now see the real reason instead. Assertions covering genuinely empty responses are unaffected.

Fixes #1337

### Is it tested? How?

Two regression tests added to `test/yamlSchema.test.ts`, alongside the existing load/parse error tests:

- **`Error while loading yaml should keep the underlying reason`** — request service rejects with `Request failed with status code 429`; asserts the message includes the status and does *not* include `No content`.
- **`Empty response while loading yaml should report no content`** — request service resolves `''`; asserts the message still includes `No content`, pinning the behaviour that should be preserved.

Verification performed:

1. **The new test fails without the fix.** Reverting only the `src` change and re-running gives:
   ```
   AssertionError: expected 'Unable to load schema from ...' to include 'Request failed with status code 429'
   ```
   Restoring the fix makes it pass, confirming it exercises the defect rather than passing incidentally.

2. **Full suite: 1321 passing, 0 failing.** This includes the pre-existing `test/schemaValidation.test.ts` case asserting `Problems loading reference ... No content`, which covers a genuinely unreachable `$ref` — so the honest `No content.` path is protected by an existing test, not only by the new one.

3. **End-to-end against a real HTTP server**, driving the actual `schemaRequestHandler` → `loadSchema` path (no stubbing of the request layer), with the message shown to the user:

   | Failure | Before | After |
   | --- | --- | --- |
   | HTTP 429, server sends a body | `No content.` | `Too Many Requests.` |
   | HTTP 403, server sends a body | `No content.` | `Forbidden.` |
   | HTTP 429, empty body | `No content.` | `HTTP status code 429.` |
   | HTTP 403, empty body | `No content.` | `Forbidden. The server is refusing to respond.` |
   | Genuinely empty body (HTTP 200) | `No content.` | `No content.` |
   | Connection refused | `No content.` | `Unable to connect to <url>.` |

   The exact text after the colon is whatever `schemaRequestHandler` produced: the server's own response body when there is one, otherwise `request-light`'s `getErrorStatusDescription(status)`. That is why a 403 with a body reads differently from a 403 without one, and why the status number appears only when the server sent no body.


datho7561:
Started looking at this late yesterday, but I'll keep looking today. When testing it out with vscode-yaml, I'm still getting a (slightly different) generic error, since when running with vscode, yaml-language-server proxies the request to the client using the custom `vscode/content` request. Unfortunately, the error response received in the `vscode/content` implementation doesn't propagate back to the language server properly. I'll see if I can throw together a vscode-yaml patch to address that.

datho7561 (changes requested):
Okay a few things that should probably be addressed before we merge this PR.

I have a nodejs webserver (expressjs) server set up to response 429 to a request, and this is the message I get:

"Unable to load schema from 'http://localhost:3000/schema.json': Too Many Requests."

Doesn't report the number 429 anywhere in the message. Is this okay behaviour, or should be adjust how the error message is massaged to be more clear?

Here's the message I get if I set up the nodejs webserver to respond with 404:

"Unable to load schema from 'http://localhost:3000/schemata.json': Not Found."

Similar idea, is that okay or should we include the number 404 somewhere in the text of the response.

For a server that doesn't respond (eg. I turn off the nodejs webserver), I get:

"Unable to load schema from 'http://localhost:3000/schema.json': ."

The space followed by a period is confusing, but it's a result of our existing mediocre massaging of the message that IMO should be improved before we expose this to the user.

X-Guardian:
I've updated the PR description to clarify the different messages that are returned with and without a body in the response. I've also added code to fix the Connection Refused message to be more useful.

datho7561 (approved):
Looks good; this is a really helpful improvement. Thanks, Simon!

We're going to have to update the translations for these messages as well, but it looks like those were already missing before this PR, so I figure we can do that in a follow up PR.

PR #1340, opened 2026-09-14
fix: allow a failed schema load to be retried later

### LINKED ISSUE #1339: bug: A failed schema load is cached for the lifetime of the process and never retried
## Describe the bug

Once a remote schema has failed to load, it is never requested again. Every subsequent validation reports the original failure, even if the URL became reachable immediately afterwards.

A momentary network problem therefore downgrades a schema for the rest of the session. Every file validated afterwards is checked without it, and every one of them reports the same stale failure. Editing and re-validating does not recover: the schema only loads again after a restart or a settings change.

This also makes a single failure look like many. One failed request produces a diagnostic on every document validated afterwards, which reads as a widespread problem rather than one transient event.

## Expected Behavior

A schema that has become reachable again should be picked up without restarting or changing settings, rather than staying unavailable for the rest of the session.

## Current Behavior

The first failure is retained permanently. With a request service that fails once and succeeds on every later call, three consecutive validations produce:

```
validation #1: requests=1  ->  Unable to load schema from '<url>': ... 429 ...
validation #2: requests=1  ->  Unable to load schema from '<url>': ... 429 ...
validation #3: requests=1  ->  Unable to load schema from '<url>': ... 429 ...
```

The request count never advances past 1: after the initial failure no further request is made, so the schema content that is now available is never fetched.

## Steps to Reproduce

1. Create a language service whose `schemaRequestService` rejects on its first call and resolves valid schema content on every call after that, then validate the same document three times:

```js
const { getLanguageService } = require('yaml-language-server');
const { TextDocument } = require('vscode-languageserver-textdocument');

let requests = 0;

const service = getLanguageService({
  // Fails once, then serves a valid schema on every later call
  schemaRequestService: async () => {
    requests++;
    if (requests === 1) {
      throw new Error('Request failed with status code 429');
    }
    return JSON.stringify({ type: 'object', properties: { foo: { type: 'string' } } });
  },
  workspaceContext: { resolveRelativePath: (rel, base) => new URL(rel, base).toString() },
});

service.configure({
  validate: true,
  schemas: [{ uri: 'https://example.com/schema.json', fileMatch: ['*.yml'], priority: 0 }],
  customTags: [],
  yamlVersion: '1.2',
});

(async () => {
  for (let i = 1; i <= 3; i++) {
    const doc = TextDocument.create('file:///example.yml', 'yaml', 0, 'foo: bar\n');
    const diagnostics = await service.doValidation(doc, false);
    console.log(`validation #${i}: requests=${requests}`, diagnostics.map((d) => d.message));
  }
})();
```

2. Observe the request service is called exactly once in total, and all three validations report the original failure:

```
validation #1: requests=1 [ "Unable to load schema from 'https://example.com/schema.json': ... 429 ..." ]
validation #2: requests=1 [ "Unable to load schema from 'https://example.com/schema.json': ... 429 ..." ]
validation #3: requests=1 [ "Unable to load schema from 'https://example.com/schema.json': ... 429 ..." ]
```

The second call to `schemaRequestService`, which would have returned a valid schema, is never made.

## Environment
- [x] Windows
- [x] Mac
- [x] Linux
- [ ] other (please specify)

X-Guardian:
### What does this PR do?

Stops a failed schema load from being cached for the lifetime of the process, so a schema that becomes reachable again is picked up without restarting.

`SchemaHandle.getUnresolvedSchema` remembers the promise from the first load attempt and returns it for every later call. A failed load is retained exactly like a successful one, so one momentary failure downgrades that schema for the whole session and reports the same stale error on every document validated afterwards.

This PR distinguishes the two cases. A successful load is cached as before. A failed load is held for `FAILED_SCHEMA_CACHE_MS` (5s), after which the next request for that schema discards the cached failure and tries again.

A negative cache is used rather than simply not caching failures, so that a persistently unreachable URL is not re-requested on every keystroke. Without that damping, the common case of a genuinely wrong URL would generate a request per validation — trading a sticky failure for a request storm. The window is short enough to recover quickly and long enough to coalesce the requests of an active editing session.

Two details are worth flagging for review:

- **A failure is detected through `errors`, not a rejection.** `loadSchema` resolves with an `UnresolvedSchema` carrying an `errors` array rather than rejecting, so the outcome is inspected after the promise settles.
- **The derived resolved schema is discarded too.** `getResolvedSchema` caches a value built from the unresolved one, so expiring only the unresolved schema would leave the stale failure being served from the resolved cache. `getResolvedSchema` now consults the unresolved schema first, and a test covers this specific path.

Only the failure path changes. Successful loads, resolution, dependency tracking and `clearSchema()` behave as before, with `clearSchema()` additionally resetting the new failure timestamp.

Fixes #1339

### Is it tested? How?

Four unit tests added to `test/yamlSchemaService.test.ts`, against a stubbed request service, so no network access is required:

- **A failed load is requested again once the window has passed** — a stub failing once then resolving valid content; asserts a second request is made and the schema validates. This is the regression test for the reported behaviour, which currently stops at a single request.
- **Failures are not re-requested within the window** — asserts only one request is made just before the window elapses.
- **Successful loads are still cached** — asserts exactly one request even well past the window, confirming no extra traffic on the normal path.
- **No stale resolved schema after expiry** — covers the `getResolvedSchema` path described above, which would otherwise keep serving the failure.

Verification performed:

1. **The tests catch the defect.** Neutralising the expiry check (restoring the old always-cache behaviour) fails 2 of the 4; the 2 that still pass are the ones asserting caching is *retained*, which should pass either way.
2. **End-to-end against a built server**, with a request service that fails once then serves a valid schema:
   ```
   #1 requests=1  ->  Unable to load schema from '...': No content.
   #2 (within window) requests=1  ->  Unable to load schema from '...': No content.
   #3 (after window) requests=2  ->  Incorrect type. Expected "number".
   ```
   The third validation re-requests and validates against the schema, where previously the count never advanced past 1.


datho7561:
Seems good from my testing and the code makes sense. 

One limitation is that the language server only attempts to reload the schema again if validation is re-triggered (eg. by modifying the file that references the schema that fails to load). I think this is fine though, because having a background job to manage reattempting to load the schema seems like a waste of resources, especially if the file that references the schema isn't being actively looked at or worked on.

Before I merge this PR, how will this interact with https://github.com/redhat-developer/yaml-language-server/pull/1342 ? The linked issues seem very similar. I think it makes sense to handle both in one PR, does this make sense to you?

X-Guardian:
Agreed on the reload limitation. It is also why this is a caching window rather than an invalidation timer: nothing fires when it expires, the next validation just finds the cached failure stale and requests again. No timers or background work.

Regarding merging the PRs, they are complementary, but at different layers. #1342 is in `schemaRequestHandler` around a single `xhr` call, and retries transient failures. This PR is in `SchemaHandle`, and time-limits how long a failure is cached, so a schema that becomes reachable is picked up again.

Happy to combine if you prefer. I've checked that the branches merge cleanly.

My mild preference is to keep them separate. #1342 is additive, in that nothing behaves differently unless a request already failed, whereas this changes caching logic that every validation uses. If someone reports an issue, being able to revert one without losing the other seems worth keeping.

If you want, I can create a separate branch/PR with both fixes combined which you can then test and decide which way to go.

datho7561:
Okay, I'm okay keeping them separately, especially if they merge cleanly. I'll try to get to reviewing them today

datho7561 (approved):
Thanks, Simon!

PR #1342, opened 2026-09-14
fix: retry transient schema request failures

### LINKED ISSUE #1341: bug: A single transient HTTP failure permanently disables a remote schema
## Describe the bug

A single transient failure when fetching a remote schema — an HTTP 429 from a rate-limited host, a 503, or a dropped connection — is treated the same as a permanent one, and the schema is reported as unloadable.

Statuses such as `429`, `502`, `503` and `504` indicate a temporary condition, and are often accompanied by a `Retry-After` header telling the client when to try again. An immediate second attempt would usually succeed, so validation is lost against a schema that is in fact reachable.

Popular public schema hosts rate limit by IP, so this is most likely to affect users behind a shared address, where unrelated requests can exhaust the allowance.

## Expected Behavior

A schema whose request fails only for a temporary reason should still load, rather than being reported as unloadable on the first failure.

Failures that are not temporary — `404`, `403`, `400` and similar — should continue to fail immediately, as trying again cannot change the outcome. Schema loading blocks validation, so recovering from a temporary failure should not noticeably delay interactive editing.

## Current Behavior

The first failure is final. A request service that fails once and would succeed on a second attempt still yields a failed schema load, and validation proceeds without the schema.

## Steps to Reproduce

1. Request a schema from a host that responds once with a temporary failure and would serve the schema on a second attempt:

```js
const { schemaRequestHandler } = require('yaml-language-server/out/server/src/languageservice/services/schemaRequestHandler');
const requestLight = require('request-light');
const { URI } = require('vscode-uri');

let attempts = 0;

// Fails once with a rate limit, then would serve a valid schema
requestLight.xhr = async () => {
  attempts++;
  if (attempts === 1) {
    const error = new Error('Too Many Requests');
    error.status = 429;
    error.responseText = 'Too Many Requests';
    throw error;
  }
  return { responseText: '{"type":"object"}', status: 200 };
};

const fs = { readFile: async () => '' };

schemaRequestHandler({}, 'https://example.com/schema.json', [], URI.parse(''), false, fs, false)
  .then((content) => console.log(`attempts=${attempts} content=${content}`))
  .catch((err) => console.log(`attempts=${attempts} failed=${err}`));
```

2. Observe that the request is abandoned after the first failure:

```
attempts=1 failed=Too Many Requests
```

The second call, which would have returned a valid schema, is never made.

## Environment
- [x] Windows
- [x] Mac
- [x] Linux
- [ ] other (please specify)

X-Guardian:
### What does this PR do?

Retries remote schema requests that fail for a transient reason, instead of treating the first failure as final.

Retries apply to statuses that indicate a temporary condition — `429`, `502`, `503`, `504` — and to connection-level errors such as `ECONNRESET`, `ETIMEDOUT` and socket hangups. Everything else, including `404`, `403` and `400`, continues to fail immediately: those are settled answers, and retrying only delays a guaranteed failure.

Where the server supplies a `Retry-After` header it is honoured, accepting either the delta-seconds or HTTP-date form. Otherwise the delay falls back to exponential backoff with jitter, so that several schemas failing at once do not retry in lockstep. Either way the delay is capped at 1s, since servers occasionally return `Retry-After` values measured in minutes.

The retry budget is deliberately small — 2 retries after the initial attempt, so 3 requests at most and a worst case near 2s — because schema loading blocks validation and must not stall interactive editing.

`schemaRequestHandler` takes a new optional trailing `retryOptions` argument, used by tests to inject the delay function and shrink the budget. It is additive, so existing callers are unaffected.

Fixes #1341

### Is it tested? How?

Eight unit tests added to `test/schemaRequestHandler.test.ts`, extending the existing `HTTP(S) schema requests` suite's approach of stubbing `request-light`'s `xhr`. No network access is required:

- **Retries a transient status** — rejects `429` once, then resolves; asserts two calls and that the content is returned.
- **Retries connection level failures** — rejects `ECONNRESET`, then resolves.
- **Does not retry a permanent status** — rejects `404`; asserts exactly one call and that the failure surfaces immediately.
- **Gives up after the budget** — rejects `503` every time; asserts exactly three calls.
- **Honours `Retry-After`** — asserts the server's `0.5` becomes a 500ms wait rather than the backoff default.
- **Caps an excessive `Retry-After`** — asserts `600` is capped to 1000ms.
- **Backs off between attempts** — asserts two waits within the expected exponential-with-jitter bounds, the second no shorter than the first.
- **Respects a zero budget** — asserts `maxRetries: 0` makes a single attempt.

The delay function is injected, so the retry suite runs in 7ms and is not timing-dependent.

Verification performed:

1. **The tests catch the defect.** Reverting the handler to a single unretried `xhr` call fails 6 of the 8; the 2 that still pass are the negative cases, which should pass either way.
3. **End-to-end against the built server**, with `xhr` replaced by a stub failing twice with `429` then succeeding:
   ```
   attempts=3 elapsed=936ms result={"type":"object"}
   ```
   The schema loads where a single attempt would have failed, and the elapsed time confirms real backoff within the intended bound.


datho7561 (approved):
This works well! Thanks, Simon!

Something that I noticed from collecting logs from the test server I was using is that the requests were coming in bursts of 5. I'm pretty sure this isn't caused by the changes this PR, but it's probably worth investigating.

This doesn't work in vscode-yaml, since we use the custom endpoint to request the schema and don't go through this code path. We'll need to provide a similar functionality there in a separate PR. The VS Code endpoint seems to have some sort of rerequesting logic already, so we'll need to work around that.

PR #1344, opened 2026-09-14
ci: remove unmaintained setup-xvfb action

### LINKED ISSUE #1343: CI warns on every run that Node 20 is deprecated, via an unmaintained action
## Describe the bug

Every run of the `CI` workflow emits a deprecation warning on all three matrix legs:

> Node 20 is being deprecated. This workflow is running with Node 24 by default. If you need to temporarily use Node 20, you can set the `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true` environment variable. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

The warning comes from the `Run Test` step, which uses `coactions/setup-xvfb@v1.0.1`. That action declares `runs: using: node20` in its `action.yml`. It is the only Node 20 action left in the repository.

It cannot be fixed by bumping the version: `coactions/setup-xvfb` is unmaintained, with its last release in December 2022 and a default branch that still declares `node20`.

The action also appears to be unnecessary here. `npm run coveralls` is `nyc` plus `mocha` over the TypeScript tests, with no VS Code harness, Electron or browser, so nothing opens a display for xvfb to provide. Its own source confirms it only applies xvfb on Linux — on `macos-latest` and `windows-latest` it passes the command straight through to `exec`, so it contributes nothing there beyond the deprecation warning.

## Expected Behavior

CI runs without deprecation warnings, and does not depend on an unmaintained third-party action pinned to a deprecated runtime.

## Current Behavior

The warning is emitted on every run. Once GitHub removes the Node 20 runtime from its runners, the `Run Test` step will fail rather than warn, taking the whole workflow with it — including the Coveralls upload that depends on its output.

Setting `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true` suppresses the warning but only defers the failure, behind an explicitly unsecure opt-in.

## Steps to Reproduce

1. Open any recent `CI` workflow run in the Actions tab.
2. Expand the `Run Test` step on any of the `macos-latest`, `windows-latest` or `ubuntu-latest` legs.
3. Observe the Node 20 deprecation warning in the annotations.

## Environment
- [x] Windows
- [x] Mac
- [x] Linux
- [ ] other (please specify)

X-Guardian:
### What does this PR do?

Removes the `coactions/setup-xvfb` wrapper from the `Run Test` step in `.github/workflows/CI.yaml`, running `npm run coveralls` directly instead.

That action declares `runs: using: node20`, and is the source of the Node 20 deprecation warning on every CI run. It is the only Node 20 action left in the repository.

Bumping it is not an option — it is unmaintained, with the last release in December 2022 and a default branch that still declares `node20`.

Removing it is safe because this project never needed it: `npm run coveralls` is `nyc` plus `mocha` over TypeScript tests, with no VS Code harness, Electron or browser, so nothing opens a display. The action only applies xvfb on Linux — on `macos-latest` and `windows-latest` it already passed the command straight through — so Linux is the only leg where behaviour changes at all.

Fixes #1343

### Is it tested? How?

- The CI Test job passes with no display.


datho7561 (approved):
Thanks, Simon!

PR #1345, opened 2026-09-17
Adds support for negated schema file-match patterns

shin19991207:
### What does this PR do?

Adds support for negated schema file-match patterns in `yaml.schemas` setting using a leading `!`, this includes exclusions of absolute paths such as `!file:///path/to/test.yaml`, as well as glob such as `!*test.yaml`.

This allows clients to exclude individual files from broad schema associations without removing the original glob.

Related to https://github.com/redhat-developer/vscode-yaml/issues/1212

### Is it tested? How?

- [x] New automated tests
- [x] Manually tested with https://github.com/redhat-developer/vscode-yaml/pull/1283


datho7561 (approved):
Looks good and works well. Thanks, Morgan!

PR #1346, opened 2026-09-17
fix: honor completion snippet support capability

### LINKED ISSUE #161: Server returns snippets even though client doesn't support them
The completion request always returns snippet-type completions, even if the client initialisation request doesn't claim to support them.

Logs:

* Initialize does not include `snippetSupport: true` which should be interpreted as not supporting snippets (only plaintext)

```
2019-07-05 22:43:11,807 - DEBUG - TX: Sending message: b'Content-Length: 654\r\n\r\n{"id": "1", "jsonrpc": "2.0", "method": "initialize", "params": {"capabilities": {"textDocument": {"completion": {"completionItemKind": {"valueSet": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25]}}, "hover": {"contentFormat": ["plaintext", "markdown"]}, "signatureHelp": {"signatureInformation": {"documentationFormat": ["plaintext", "markdown"], "parameterInformation": {"labelOffsetSupport": false}}}}}, "initializationOptions": {}, "processId": 59070, "rootPath": "/Users/ben/.vim/bundle/YouCompleteMe/third_party/ycmd/ycmd", "rootUri": "file:///Users/ben/.vim/bundle/YouCompleteMe/third_party/ycmd/ycmd"}}'
```

* completion request 

```
2019-07-05 22:46:30,803 - DEBUG - TX: Sending message: b'Content-Length: 215\r\n\r\n{"id": "4", "jsonrpc": "2.0", "method": "textDocument/completion", "params": {"position": {"character": 2, "line": 0}, "textDocument": {"uri": "file:///Users/ben/Development/lsp/yaml/tests/ansible/tasks/main.yml"}}}'
```

* response (snipped)

```
2019-07-05 22:46:30,828 - DEBUG - RX: Received message: b'{"jsonrpc":"2.0","id":"4","result":{"items":[{"kind":10,"label":"shell","insertText":"shell: $1","insertTextFormat":2,"documentation":"","textEdit":{"range":{"start":{"line":0,"character":2},"end":{"line":0,"character":4}},"newText":"shell: $1"}},
```

Note: `"insertTextFormat":2` is 'snippet' :

```
	/**
	 * The primary text to be inserted is treated as a snippet.
	 *
	 * A snippet can define tab stops and placeholders with `$1`, `$2`
	 * and `${3:foo}`. `$0` defines the final tab stop, it defaults to
	 * the end of the snippet. Placeholders with equal identifiers are linked,
	 * that is typing in one will update others too.
	 */
	export const Snippet = 2;
```

Shubham-Padkonde:
### What does this PR do?

Honor `textDocument.completion.completionItem.snippetSupport`. When it is missing or false, generated completions use schema defaults directly and omit tab stops and snippet escaping. Snippet-capable clients retain the existing placeholders and choices. Literal dollar signs in plain-text defaults are preserved.

The capability is exposed in the public type and `LATEST` constant. Schema-provided `defaultSnippets` are offered only to snippet-capable clients, rather than attempting to interpret arbitrary snippet syntax for other clients. This behavior is documented in the README.

The review follow-up removes `vscode-snippet-parser` entirely. Plain text is generated at the insertion points instead of converting completed snippets afterward; no new dependency remains.

Fixes #161.

### Is it tested? How?

Windows, Node 24.15.0:
- All 14 capability regressions pass, including omitted/false/true support, defaults, literal dollar signs, nested required properties, and capability-gated schema snippets.
- Full test run: 1,356 passing, one pending, two external-schema validation timeouts. A targeted run with the original pre-PR completion source reproduced the drone-schema timeout; the other timeout passed on that baseline retry. The full suite is not claimed green.
- `npm run build` passes ESLint, TypeScript, CommonJS, UMD and ESM builds. Prettier and `git diff --check` pass.

The hosted Node/platform matrix has not been run locally. The original unsupported-capability regression evidence remains in the earlier revision history; the current review adds direct plain-text generation coverage.

AI assistance: Codex implemented this review follow-up, ran the stated checks, and drafted this description. No independent human review is claimed.


datho7561 (changes requested):
It looks like vscode-snippet-parser is unmaintained (no commits in 9 years). Also, it's not published by the VS Code team, but a seemingly by unmaintained project called onivim. As a result, I don't like adding it as a dependency.

I think this approach in general isn't great, since it's removing the tab stops after the fact rather than avoiding putting them in in the first place. I think the implementation could be much simpler if we wrapped each time we add a tab stop with an if/else. It seems like all the tab stops are added in `getInsertTextForGuessedValue` or `getInsertTextForProperty`. Also, we have the helper method `this.evaluateTab1Symbol` available that can be used to substitute the tab stop with the default value assigned for the tab stop if we want to go the route of pre-evaluating each tab stop.

